PR #173, opened 2026-03-05
feat: add postgres support to switch between databases and schemas

json-nan:
# Summary

Adds support to change between databases and schemas using postgres engine

image

image


json-nan:
@datlechin check again please

PR #215, opened 2026-03-09
fix: harden ssh local tunnel binding against TOCTOU races

### LINKED ISSUE #216: Bug: SSH Tunnel TOCTOU race attack vulnerability
### What happened?

Right now, TablePro scans for a free local port, releases it, and only then launches ssh -L .... 
Another local process can bind that port in the gap between the availability check and the actual ssh bind, which is a potential vulnerability and attack surface.

### Environment

0.15.0 / macOS 26.3.1 / Apple Silicon

### Screenshots / Logs

_No response_

eliottwantz:
Closes #216 

## Summary

Harden SSH local port forwarding against a TOCTOU race during tunnel setup.

Previously, TablePro scanned for a free local port, released it, and only then launched `ssh -L ...`. Another local process could bind that port in the gap between the availability check and the actual `ssh` bind.

This change removes that race by letting `ssh` perform the bind directly and retrying only when `ssh` reports a local forwarding bind failure.

## What changed

- Replaced the pre-bind `findAvailablePort()` flow with launch-and-retry logic over candidate ports
- Added `-o ExitOnForwardFailure=yes` so local forwarding bind failures fail fast
- Bound the local forward explicitly to `127.0.0.1`
- Tightened readiness checks so the forwarded port must:
  - accept connections, and
  - be owned by the launched process tree
- Added focused unit tests for:
  - retryable local bind failures
  - process-tree ownership detection

## Why

The old flow had a classic TOCTOU window:

1. TablePro checked whether a local port was free
2. The check released the port
3. `ssh` later attempted to bind the same port

A competing local process could win that race and either:
- cause `ssh` to fail binding, or
- be mistaken for the expected tunnel if readiness only checked whether *something* was listening

This PR makes tunnel setup fail closed and retries only on the specific local bind race case.

## Testing

- Added unit coverage for bind-failure classification
- Added unit coverage for process-tree ownership checks

### DIFF
--- a/TablePro/Core/SSH/SSHTunnelManager.swift
+++ b/TablePro/Core/SSH/SSHTunnelManager.swift
@@ -45,6 +45,12 @@ struct SSHTunnel {
     let createdAt: Date
 }
 
+private struct SSHTunnelLaunch {
+    let process: Process
+    let errorPipe: Pipe
+    let askpassScriptPath: String?
+}
+
 /// Manages SSH tunnels for database connections using system ssh command
 actor SSHTunnelManager {
     static let shared = SSHTunnelManager()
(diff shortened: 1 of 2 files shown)


eliottwantz:

PR #647, opened 2026-04-09
feat: support all monospaced fonts installed on the local system

nexxai:
## Problem

I wanted to use a monospaced font that *wasn't* one of the 4 default included ones (specifically JetBrains Mono) but was obviously unable to as the list was hardcoded.

## Solution

I replaced the hardcoded font option flow with a shared runtime resolver that discovers installed monospaced font families and uses them in both settings UI and runtime theme font resolution. The resolver keeps `System Mono` as a safe default and fallback when a saved font is unavailable. Theme storage remains string-based and backward-compatible, and docs/tests were updated to cover dynamic listing and fallback behavior.

This is my first contribution to this repo so if I'm missing anything, please let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix idle ping spin loop caused by exhausted AsyncStream iterator (#618)
 - Skip exact row count for large tables — use database statistics estimate (#519)
 
+### Changed
+
+- Theme font pickers now list installed monospaced fonts dynamically instead of a fixed built-in list
+
 ## [0.28.0] - 2026-04-07
 
 ### Added
--- a/TablePro/Models/Settings/EditorSettings.swift
+++ b/TablePro/Models/Settings/EditorSettings.swift
@@ -9,22 +9,14 @@ import Foundation
 internal struct FontFamilyOption: Equatable, Identifiable, Sendable {
     let id: String
     let displayName: String
-    let isRecommended: Bool
 }
 
 internal enum EditorFontResolver {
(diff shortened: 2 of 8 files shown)


nexxai:

datlechin (commented):
Thanks for the contribution! Clean approach overall. A few things to address:

1. **Remove trailing comma** in `recommendedFamilies` — project convention is no trailing commas (see CLAUDE.md / `.swiftformat`)

2. **`isRecommended` is unused** — it's set but never read anywhere (not in the picker, not in sorting). Remove it, or use it in the UI to distinguish recommended fonts. Same for `displayName(for:)` — not called anywhere.

3. **Cache `availableMonospacedFamilies()`** — this iterates all system fonts and creates `NSFont` instances to check monospace traits. It runs on every SwiftUI re-evaluation since it's called inside `ForEach` in the view body. A simple `static let` cache would fix this — fonts don't change at runtime.

After these, good to merge.

datlechin (changes requested):
Nice PR, but have few things to fix:

- isRecommended and displayName seem like definded but no read anywhere
- availableMonospacedFamilies() make it static let instead

nexxai:
Sorry about the extra properties.  I had been trying out a couple different features I've been thinking about and forgot to `git add -p` here.  I've removed them and switched the `availableMonospacedFamilies` to a `static let` property.

datlechin:
Hi @nexxai, thanks for the quick update. Also, please update the tests as well. Currently, the test still has parentheses around `EditorFontResolver.availableMonospacedFamilies()`.

nexxai:
Yeah, I guess that would be helpful to include in the commit too.

datlechin (approved):

PR #943, opened 2026-04-29
feat: allow manual AI model names for custom providers

FaiChou:
## Summary

Adds a manual model name input in AI provider settings.

Some OpenAI-compatible providers do not expose a `/models` endpoint. This is common for a few third-party AI APIs and coding-plan providers. Before this change, TablePro only allowed selecting models fetched from the provider, so users could not configure those endpoints even when chat completions worked with a known model name.

## Changes

- Add a model name text field in the AI provider detail sheet
- Keep the fetched model list as an optional picker when `/models` is available
- Trim whitespace before saving the selected or manually entered model
- Update AI Assistant docs and changelog

## Verification

- Ran `xcodebuild -project TablePro.xcodeproj -scheme TablePro -configuration Debug build -skipPackagePluginValidation`
- Retried with `CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO`
- Build is blocked locally because `Secrets.xcconfig` is missing. The standard run is also blocked by missing Mac Development signing assets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Shift+Tab navigates to the previous cell in the data grid
 - Copy rows writes TSV, HTML table, and plain text to the clipboard for richer paste in spreadsheet apps
 - Row drag adds TSV and HTML representations alongside the internal drag type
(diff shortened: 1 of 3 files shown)


FaiChou:

datlechin (approved):
Thank you

PR #1028, opened 2026-05-06
fix: drop hardcoded team prefix to support external-contributor builds

### LINKED ISSUE #1020: Hardcoded team prefix and bundle ID block external contributors from building locally
## Summary

External contributors who don't belong to the TableProApp Apple Developer team (`D7HJ5TFYCU`) cannot get a working Debug build out of the box. Several places hardcode either the bundle ID `com.TablePro` or the team prefix `D7HJ5TFYCU`, both of which only work for members of the official team. Each lands as a separate failure mode and forces the contributor to maintain a stack of local-only patches that must be carefully kept out of every commit.

## Where it shows up

A contributor on a free personal Apple ID (or any non-official team) hits these in order:

### 1. Provisioning fails: iCloud capability not supported on personal teams

`TablePro/TablePro.entitlements` enables iCloud / CloudKit and pins the team:

```xml
<key>com.apple.developer.icloud-container-identifiers</key>
<array><string>iCloud.com.TablePro</string></array>
<key>com.apple.developer.icloud-services</key>
<array><string>CloudKit</string></array>
<key>com.apple.developer.team-identifier</key>
<string>D7HJ5TFYCU</string>
```

Personal teams don't support the iCloud capability at all, so Xcode refuses to create a profile.

### 2. Bundle ID conflict

`PRODUCT_BUNDLE_IDENTIFIER = com.TablePro` is already registered to `D7HJ5TFYCU` in Apple's developer portal. A second team can't claim the same ID, so the contributor has to rename it (e.g. `com.<theirhandle>.TablePro`) to even register a new App ID under their personal team.

### 3. Keychain access fails at runtime

`TablePro/Core/Storage/KeychainHelper.swift:27` hardcodes the access group:

```swift
private let accessGroup = "D7HJ5TFYCU.com.TablePro.shared"
```

Even if the contributor strips iCloud from a local entitlements file, this still fails because:

- `baseQuery` always sets `kSecAttrAccessGroup = "D7HJ5TFYCU.com.TablePro.shared"`, an access group the contributor's signature cannot claim.
- `kSecUseDataProtectionKeychain = true` additionally requires `keychain-access-groups` to be present in the entitlements (with at least one valid group), independent of what the query passes. An entitlements file with no group at all returns `errSecMissingEntitlement` for every write.

The visible symptom is "Access denied for user 'X'@'Y' (using password: NO)" on Test Connection, because the password is silently dropped by the keychain write.

### 4. Same pattern in CloudKitSyncEngine and SyncCoordinator

`CloudKitSyncEngine.init` calls `CKContainer(identifier: "iCloud.com.TablePro")`, which traps with `EXC_BREAKPOINT` on launch when the iCloud entitlement is missing. `SyncCoordinator.currentAccountId` does the same.

## Net effect

To run Debug locally, an external contributor currently has to:

1. Create a separate `TablePro.local.entitlements` (no iCloud, with a personal-team-friendly `keychain-access-groups` declaring `$(AppIdentifierPrefix)com.TablePro` so the data protection keychain works at all).
2. Switch the Debug `Code Signing Entitlements` build setting to that file in Xcode.
3. Change the Debug `PRODUCT_BUNDLE_IDENTIFIER` to something unique under their team.
4. Patch `KeychainHelper`, `CloudKitSyncEngine`, and `SyncCoordinator` to fall back gracefully when the official entitlements aren't present.
5. Remember to never commit any of those changes (they would break the official Release signing path).

This is fragile and unfriendly to first-time contributors, especially since failures don't surface as configuration errors but as misleading runtime symptoms ("password not saved", "unsupported URL", crashes on launch).

### Proposed solution

The mainline source can be made portable across teams without changing Release behavior:

1. **Use `$(AppIdentifierPrefix)` in entitlements** instead of literal `D7HJ5TFYCU.`:
   ```xml
   <key>keychain-access-groups</key>
   <array><string>$(AppIdentifierPrefix)com.TablePro.shared</string></array>
   ```
   `codesign` substitutes the active team's prefix at sign time. The official build still produces `D7HJ5TFYCU.com.TablePro.shared`. A personal-team build produces `<their-team>.com.TablePro.shared`. Both work.

2. **Read the access group at runtime in `KeychainHelper`** instead of hardcoding `D7HJ5TFYCU.com.TablePro.shared`. The team prefix can be fetched from the app's signing info (`SecTaskCopyValueForEntitlement` for `keychain-access-groups`) and combined with the suffix `com.TablePro.shared`. That way the constant works for any team.

3. **Make iCloud a soft dependency**. Detect the iCloud entitlement at runtime in `CloudKitSyncEngine.init` (using `SecTaskCopyValueForEntitlement` for `com.apple.developer.icloud-services`); if absent, leave `container` / `database` as `nil` and have the sync methods throw `SyncError.accountUnavailable`. `SyncCoordinator` already has the `do/catch` paths to surface this gracefully as "iCloud unavailable" in the UI. Same treatment for `SyncCoordinator.currentAccountId`.

4. **Keep `PRODUCT_BUNDLE_IDENTIFIER = com.TablePro`** for Release. For Debug, document a one-line setup step (rename in Build Settings) or provide a separate Debug `xcconfig` that contributors can copy.

(1)–(3) are pure-source changes that the official build wouldn't notice. (4) is a docs/setup nudge, not a code change.

## Why this matters

Without these, every external PR author repeats the same multi-hour debugging session before they can run their first Debug build. The current error messages don't point at the team-prefix issue, so the diagnosis time is high. The fixes are localized to a few files and don't change anything for in-team developers.

Happy to send a PR if this direction sounds reasonable.

### Related database type

None

tonghs:
## Summary

Implements [#1020](https://github.com/TableProApp/TablePro/issues/1020). Makes the codebase portable across Apple Developer teams so external contributors can build and run a Debug build on a personal team without hand-patching files. Release behavior under the official `D7HJ5TFYCU` team is unchanged: codesign reproduces the same effective entitlements.

Three failure modes were blocking external contributors:

1. **Provisioning** — `TablePro.entitlements` declared the iCloud capability, which free personal teams don't support, so Xcode refused to create a profile.
2. **Keychain** — `KeychainHelper.accessGroup` hardcoded `D7HJ5TFYCU.com.TablePro.shared`, an access group only the official team can sign for. Every keychain write returned `errSecMissingEntitlement`. Visible symptom: "Access denied (using password: NO)" on Test Connection because the password write silently failed.
3. **Runtime** — `CloudKitSyncEngine.init` called `CKContainer(identifier:)` unconditionally, which traps with `EXC_BREAKPOINT` on launch when the iCloud entitlement is absent.

## Approach

Three pure-source changes to support both teams from the same code, plus one new shipped file and a docs update:

- `TablePro.entitlements` keychain group is now `$(AppIdentifierPrefix)com.TablePro.shared`. `codesign` substitutes the active team's prefix at sign time, so the official build still signs as `D7HJ5TFYCU.com.TablePro.shared` and a personal-team build signs as `<their-prefix>.com.TablePro.shared`. Removed the redundant `application-identifier` and `team-identifier` keys (codesign auto-injects these).
- `KeychainHelper` reads `keychain-access-groups` from the running process via `SecTaskCopyValueForEntitlement` at init, prefers the entry ending in `.com.TablePro.shared`, and validates it matches `^[A-Z0-9]{10}\..+` (rejects unsubstituted `$(...)` templates from ad-hoc builds). When no valid group is declared, `baseQuery` simply omits `kSecAttrAccessGroup`, letting the keychain fall back to the default per-app group.
- `CloudKitSyncEngine` adds a `static func hasICloudEntitlement()` and makes `container` / `database` optional. `init` only constructs them when the entitlement is present; otherwise it logs a warning and stays disabled. Each method `guard let container/database else { throw SyncError.accountUnavailable }`. `SyncCoordinator.currentAccountId` was reaching across into a duplicated `CKContainer(identifier:)` literal — it now delegates to a new `engine.currentAccountId()` method, removing that duplication.
- `TablePro/TablePro.Debug.entitlements` is new: identical to the default minus the iCloud keys. Contributors on personal teams point Debug builds at it instead of the iCloud-enabled file. iCloud sync degrades gracefully at runtime when the entitlement is absent (existing `do/catch` paths surface "iCloud unavailable" through the existing UI).

## Tests

`TableProTests/Core/Sync/CloudKitSyncEngineTests.swift` — 6 cases verifying the soft-dependency path. Each gates with `try #require(!CloudKitSyncEngine.hasICloudEntitlement())` so it auto-skips when the test host happens to be signed with iCloud (otherwise the tests would hit real CloudKit). Covers `checkAccountStatus`, `ensureZoneExists`, `push` (non-empty + empty short-circuit), `pull`, and `currentAccountId`.

## Contributor setup

`CONTRIBUTING.md` adds a section on the one-time Xcode UI setup for non-official teams: Team selection (with note about repeating for other targets if signing fails), Bundle Identifier change, and switching `Code Signing Entitlements` to the new Debug variant. Also documents that the resulting `pbxproj` changes shouldn't be committed (revert or use `git update-index --skip-worktree`).

## CHANGELOG

Entry under `[Unreleased]` → `### Fixed`. Fixes #1020.

## Test plan

- [x] `swiftlint lint --strict` — 0 violations
- [x] Debug build succeeds on a personal Apple ID team after the documented Xcode UI setup
- [x] Saving a database connection password and reconnecting works (verifies keychain access group fallback)
- [x] iCloud sync UI surfaces "disabled" state instead of crashing on a build without iCloud entitlement
- [x] `xcodebuild ... test -only-testing:TableProTests/CloudKitSyncEngineTests` passes locally
- [ ] Maintainer to verify: Release build on the official team produces a binary with the same effective entitlements (`codesign -d --entitlements -` should still show `D7HJ5TFYCU.com.TablePro.shared` as the keychain access group)


datlechin:
Thank you.

PR #1029, opened 2026-05-06
fix: use configured model for AI provider connection tests

FaiChou:
Fixes AI provider connection tests for OpenAI-compatible/custom providers to use the configured model name when available, while preserving the existing "test" fallback for empty model fields.

Also treats JSON error bodies as valid API responses even when the Content-Type header is missing or incorrect, so parameter errors from reachable providers do not incorrectly appear as connection failures.


datlechin (approved):
Thank you

PR #1040, opened 2026-05-06
fix: align data grid header padding

overtrue:
## Summary

- Add the same 4pt horizontal inset to data grid column headers that result cells use.
- Render the row-number `#` header with the same header cell and right alignment as row-number cells.
- Add coverage for `SortableHeaderCell.titleRect(forBounds:)`.

### before

Clipboard_Screenshot_1778067344

### after

Clipboard_Screenshot_1778067352

## Root Cause

Data cells applied 4pt leading and trailing constraints, but unsorted headers used the default `NSTableHeaderCell` interior drawing. That left header titles flush against grid edges, and the row-number header used a different cell path from the row-number content cells.

## Validation

- `xcodebuild -project TablePro.xcodeproj -scheme TablePro test -skipPackagePluginValidation -only-testing:TableProTests/SortableHeaderCellTests CODE_SIGNING_ALLOWED=NO`
- `swiftlint lint --strict --config .swiftlint.yml` using the SwiftLint plugin binary
- `xcodebuild -project TablePro.xcodeproj -scheme TablePro -configuration Debug build -skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO`


datlechin (approved):
LGTM, thank you

PR #1123, opened 2026-05-08
fix(postgres): escape underscore in schema filter so pgboss/pgcrypto/etc. appear

michel:
## Summary

The schema-listing query in the PostgreSQL and Redshift drivers filters built-in system schemas with `LIKE 'pg_%'`. In SQL `LIKE`, `_` is a single-character wildcard, so the pattern matches not just `pg_catalog` / `pg_toast` / etc. but also any user schema starting with `pg` followed by at least one character — `pgboss`, `pgcrypto`, `pgvector`, `pgaudit`, `pgrouting`, and so on. These were silently dropped from the sidebar and the Cmd+K schema picker even when the connected role had `USAGE` on them.

The fix escapes the underscore: `LIKE 'pg\_%' ESCAPE '\'`. The `_` is now matched literally and only built-in `pg_*` schemas are excluded.

## Reproduction

1. Create a schema named `pgboss` (or install [pg-boss](https://github.com/timgit/pg-boss), which creates one on first run).
2. Grant `USAGE` to the connected role: `GRANT USAGE ON SCHEMA pgboss TO myrole;`.
3. Connect with TablePro. The sidebar shows only `public`. `Cmd+K` does not surface tables under `pgboss`.
4. With the fix, the sidebar lists `public` **and** `pgboss`.

A user-side check confirms the role can see the schema:

```sql
SELECT n.nspname, has_schema_privilege(current_user, n.nspname, 'USAGE') AS usage
FROM pg_namespace n
WHERE n.nspname NOT LIKE 'pg\_%' ESCAPE '\'
  AND n.nspname <> 'information_schema'
ORDER BY 1;
```

## Changes

- `Plugins/PostgreSQLDriverPlugin/PostgreSQLSchemaQueries.swift` — new file; extracts the two schema-listing queries into a shared enum so they can be unit-tested without a live DB connection.
- `Plugins/PostgreSQLDriverPlugin/PostgreSQLPluginDriver.swift`, `RedshiftPluginDriver.swift` — `fetchSchemas()` now reads the SQL from the shared enum.
- `TableProTests/PluginTestSources/PostgreSQLSchemaQueries.swift` — symlink mirroring the existing pattern (`RedisQueryBuilder.swift`, `MongoDBQueryBuilder.swift`, …) so the constants compile into the test target.
- `TableProTests/Plugins/PostgreSQLSchemaFilterTests.swift` — new Swift Testing suites covering the filter behaviour with a small SQL-LIKE evaluator. Parametrised cases:
  - `pgboss`, `pgcrypto`, `pgvector`, `pgaudit`, `pgrouting` are **kept**
  - `pg_catalog`, `pg_toast`, `pg_temp_1`, `pg_toast_temp_1` (and Redshift's `pg_internal`) are **filtered**
  - `information_schema` is filtered
  - plain user schemas (`public`, `auth`, `myapp`, `analytics`) are kept
- `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.

## Verification

```
xcodebuild -project TablePro.xcodeproj -scheme TablePro -derivedDataPath build/dd \
  test-without-building -skipPackagePluginValidation \
  -only-testing:TableProTests/PostgreSQLListSchemasTests \
  -only-testing:TableProTests/RedshiftListSchemasTests
# → 40/40 passed
```

I sanity-checked the regression coverage by temporarily reverting the `ESCAPE '\'` fragment: the `retainsPgPrefixedUserSchemas` cases fail as expected.

`swiftlint lint --strict` is clean on all touched files.

## Test plan

- [x] Unit tests pass with the fix
- [x] Unit tests fail without the fix (regression coverage verified)
- [x] SwiftLint clean on touched files
- [x] Local Debug build of `TablePro.app` confirms patched binary connects to AWS RDS over an SSH tunnel and lists `pgboss` alongside `public` for a role with `USAGE` on both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- PostgreSQL/Redshift: schema picker no longer hides user schemas whose names start with `pg` (`pgboss`, `pgcrypto`, `pgvector`, `pgaudit`, etc.). The system-schema filter now escapes the underscore in `LIKE 'pg\_%'` so it is matched literally instead of as SQL LIKE's single-character wildcard.
 - AI Chat: `@` mention detection no longer breaks when the cursor sits right after an emoji or other non-BMP character
 - AI Chat: Fix Error prompt now reads "MongoDB query" and "Redis command" using the database display name, instead of the raw query language label
 - Internal: tab session registry binds automatically when a coordinator falls back to creating its own registry, so unit tests no longer trip the filter-state debug assertion
--- a/Plugins/PostgreSQLDriverPlugin/PostgreSQLPluginDriver.swift
+++ b/Plugins/PostgreSQLDriverPlugin/PostgreSQLPluginDriver.swift
@@ -641,12 +641,7 @@ final class PostgreSQLPluginDriver: PluginDatabaseDriver, @unchecked Sendable {
     }
 
     func fetchSchemas() async throws -> [String] {
-        let result = try await execute(query: """
-            SELECT schema_name FROM information_schema.schemata
-            WHERE schema_name NOT LIKE 'pg_%'
-              AND schema_name <> 'information_schema'
-            ORDER BY schema_name
-            """)
+        let result = try await execute(query: PostgreSQLSchemaQueries.listSchemas)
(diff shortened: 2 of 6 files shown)


michel:

datlechin (approved):
LGTM, thank you.

PR #1210, opened 2026-05-11
feat(backup): add File > Backup… for PostgreSQL via pg_dump

samirmhsnv:
## Summary
- New `File > Backup…` menu item for PostgreSQL and Redshift connections. Pick a database, choose a destination, runs `pg_dump -Fc` with live byte progress and cancel.
- Reuses the connection's active SSH tunnel via `session.effectiveConnection`, so tunneled connections work without opening a second port forward.
- `pg_dump` is auto-detected from `PATH` and common Homebrew locations. Override under Settings > Terminal > CLI Paths > pg_dump.

## Notes for reviewers
- `DatabaseSwitcherSheet` gains a `Mode` enum (`.switch` default, `.backup`). Existing call sites stay on the default and are unchanged. In backup mode the schemas tab and create/drop affordances are hidden, the primary button reads "Back Up…", and the callback does not auto-dismiss so the host sheet can transition to the save panel.
- `PostgresBackupService` polls file size every 250ms for progress (pg_dump has no progress hook). Cancel sends SIGTERM and removes the partial file. Stderr is captured (capped at 64KB) and shown on failure.
- SSL mode raw values are mapped to libpq's `PGSSLMODE` values (`require`, `verify-ca`, `verify-full`, etc.) instead of being passed through verbatim.
- `--no-password` is passed to `pg_dump` so it fails fast with an auth error instead of hanging on a TTY prompt when `PGPASSWORD` is missing.
- Headless `xcodebuild` from CLI fails locally on a SwiftLint plugin sandbox issue in `CodeEditTextView`/`CodeEditSourceEditor` (unrelated to this change). SwiftLint `--strict` is clean on every changed file. Verify the build from Xcode.

## Test plan
- [ ] Postgres connection without SSH: open Backup…, pick a DB, save, verify the `.dump` file is a valid custom-format archive (`pg_restore -l file.dump`).
- [ ] Postgres connection with SSH tunnel: same flow, verify it uses the existing tunnel (no second port forward in `lsof`).
- [ ] Cancel mid-backup: confirm the partial file is removed and the sheet shows "Backup Cancelled".
- [ ] Non-Postgres connection: Backup… menu item is disabled.
- [ ] Custom pg_dump path: set under Settings > Terminal > CLI Paths > pg_dump, verify it is used.
- [ ] No pg_dump on system: backup fails with the install hint.
- [ ] Redshift connection: backup works (Redshift shares the PostgreSQL plugin).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - File > Backup... for PostgreSQL and Redshift connections: pick a database from the current connection, choose a destination, and run `pg_dump` in custom archive format (`-Fc`) with progress and cancel. Reuses the existing SSH tunnel when one is active, and honors a custom pg_dump path under Settings > Terminal > CLI Paths.
+- File > Restore... for PostgreSQL and Redshift connections: pick a `pg_dump` backup file, pick a target database on the current connection, and run `pg_restore --no-owner --no-acl` with progress and cancel. Reuses the existing SSH tunnel, and honors a custom pg_restore path under Settings > Terminal > CLI Paths.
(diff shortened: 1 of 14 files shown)


samirmhsnv:

samirmhsnv:
Superseded by #1211, which covers both Backup and Restore with a combined description and test plan.

Issue #1296, opened 2026-05-16
Display a Schema Grouping system on the "Tables" sidebar.


luizvergennes:
### Problem

I often work with multiple schemas in my databases and normally have to work with several at once. The current interface requires me to use a popover setting in the header to switch schemas every time I need to check a table from another schema.

### Proposed solution

It would be amazing to have a setting that, when enabled, displays the schemas in an accordion/dropdown format on the "Tables" sidebar, where I can easily expand them and interact with the tables of each schema.

### Related database type

None

PR #1300, opened 2026-05-16
feat(sidebar): show schemas as collapsible sections in the Tables sidebar

luizvergennes:
solves #1296

Adds a Sidebar group in the settings, with a "Display schemas on the sidebar" option:
image

Enabling this option, shows all schemas on the sidebar, and they can be expanded:
image
image
image

✨ Used AI to help me create this PR.


luizvergennes:

luizvergennes:
I do feel it could still be expanded a bit, to allow the tabs to also work across, schemas. I did not test, on my build. But on the production app, the tabs close as soon as I switch to another schema.

datlechin:
Thank you. The PR looks good, but it needs some improvement. I’ll continue to improve it.

datlechin:
Found one more bug while testing the branch: clicking a table in a non-current schema (e.g. \`auth.audit_log_entries\` while connected to \`public\`) errored with \`relation "audit_log_entries" does not exist\` because the SELECT was unqualified.

Root cause: \`onDoubleClick\` → \`openTableTab(tableName)\` only passed the table name, never the schema. \`tableContext.schemaName\` was being set to the connection's current schema, and the SELECT was built without the schema prefix. The PR's earlier attempt to fix this by calling \`switchSchema\` before opening was brittle (mutates connection-wide state, races with other tabs, fails silently for drivers without \`SchemaSwitchable\`).

Deeply correct fix in 12b20b889 on \`pr-1300-review-fixes\`:

- \`openTableTab\` gains a \`schema: String?\` parameter
- Threads schema through \`addTableTab\` / \`addPreviewTableTab\` / \`buildBaseTableQuery\` (the SQL builder already supported it)
- Sidebar callback passes \`table.schema\` directly
- Fast-path tab match now compares schema too (so opening \`auth.users\` and \`public.users\` correctly creates separate tabs instead of reusing)
- \`handleTableOpen\` schema-switch hack removed entirely - it's now just \`onDoubleClick?(table)\`

Result: SELECT is built as \`SELECT * FROM "auth"."audit_log_entries" LIMIT 1000\` regardless of which schema the connection is currently pointed at. No connection-state mutation, no schema thrashing across windows.

luizvergennes:
Merged your fix into mine! Thank you so much for your help/work! 🙌

PR #1401, opened 2026-05-23
feat(connections): AWS IAM authentication for PostgreSQL and MySQL (#1291)

### LINKED ISSUE #1291: Support RDS IAM Auth
### Problem

I need to connect to RDS using IAM authentication (token-based, no static password). 

### Proposed solution

Add an "AWS IAM" auth mode for PostgreSQL and MySQL connections that generates a short-lived auth token using the AWS credential chain, similar to how the DynamoDB driver works.

### Alternatives considered

Manually running aws rds generate-db-auth-token and pasting it as the password, but tokens expire every 15 minutes, making this impractical.

### Related database type

N/A / General

datlechin:
## What

Adds an **AWS IAM** authentication mode for PostgreSQL and MySQL connections to RDS and Aurora. Instead of a static password, TablePro resolves AWS credentials and generates a short-lived RDS IAM auth token (a SigV4-presigned `rds-db` connect URL, valid 15 minutes) to use as the password. Closes #1291.

In the connection form, set **Authentication** to one of:
- **AWS IAM (Access Key)**: access key ID, secret access key, optional session token
- **AWS IAM (Profile)**: a named profile from `~/.aws/credentials`
- **AWS IAM (SSO)**: a profile backed by IAM Identity Center (`aws sso login` first)

The token is regenerated on every connect and reconnect, so a 15-minute expiry never reaches the user. SSL is forced on for IAM. The AWS Region is auto-derived from the RDS hostname and can be overridden.

## Design decisions

- **No AWS SDK dependency.** Token generation is hand-rolled SigV4 over CommonCrypto, in a self-contained host-app module (`TablePro/Core/Database/AWS/`). The project has no AWS SDK, and DynamoDB already signs this way; `aws-sdk-swift` would pull in a C runtime (aws-crt) + Smithy, which is disproportionate for token generation and complicates the universal binary. The SSO flow (INI parse → token cache → `GetRoleCredentials`) is ported from the DynamoDB driver, whose copy is internal to a separate registry binary the host can't link.
- **Injection at the factory seam.** The token is generated in `DatabaseDriverFactory.resolvePassword` and passed as the driver password. The PostgreSQL/MySQL plugins are AWS-agnostic and unchanged; no PluginKit ABI bump.
- **`resolvePassword` / `createDriverFromPlugin` are now `async throws`** so SSO can do its credential-exchange network call and so credential errors surface with actionable messages. Their only caller, `createDriver`, was already async.
- **Token is never cached.** `session.cachedPassword` is left nil for IAM connections, so every connect and every health-monitor reconnect regenerates a fresh token rather than replaying a dead one. This is the correctness-critical part.
- **Single-level connection fields.** One `awsAuth` dropdown drives all credential-source fields directly, because `isFieldVisible` doesn't resolve nested visibility. `AuthPaneViewModel.hidesPassword` now hides the password when a dropdown auth field is non-default (DynamoDB's `.secure`-based hiding is unaffected).

## Files

New module: `AWSCredentials`, `AWSAuthError`, `AWSSSO` (+ `AWSSSOError`), `AWSSigV4`, `AWSCredentialResolver`, `RDSAuthTokenGenerator`, `RDSEndpoint`.

Changed: `DatabaseDriver` (async IAM branch + SSL floor), `DatabaseManager+Sessions`/`+Health` (never cache the token for IAM), `DatabaseConnection.usesAWSIAM`, `AuthPaneViewModel`, `MySQLPlugin`/`PostgreSQLPlugin` (auth fields), CHANGELOG, both DB docs.

## Tests

`AWSIAMAuthTests`: SigV4 primitives against NIST/RFC 4231 vectors, token shape + determinism + session-token inclusion, region derivation, the access-key resolver, and AWS config INI parsing. Profile/SSO file and network paths aren't unit-tested (they need the filesystem/AWS).

## Verify on real RDS (not unit-testable)

- **MySQL cleartext plugin:** this relies on MariaDB Connector/C sending the token directly over TLS (AWS states MariaDB clients don't need `--enable-cleartext-plugin`). If MySQL IAM fails with a plugin error, the fix is a small `additionalFields`-driven connector option (no ABI bump). Please test against an RDS MySQL instance with an IAM-enabled user.
- **SSO** requires a prior `aws sso login --profile <name>`.

## Not included

- pg_dump/restore under IAM (it resolves its own password). In-app browse/query/edit works; external dump under IAM is out of scope for this PR.


imdkbj:
@datlechin This doesn't support the cli login using _aws login_.
There is no entry created in credentials file which seems required! Entry are only in config file like as below.

```
[profile c9]
login_session = arn:aws:iam::x:user/y
region = ap-south-1

// adhoc to support for older sdk
[profile c9dev]
credential_process = aws configure export-credentials --profile c9 --format process
region = ap-south-1
login_session = arn:aws:iam::x:user/y
```

PR #1415, opened 2026-05-25
fix(settings): persist font size changes in Appearance theme editor

### LINKED ISSUE #1381: Font size changes in Appearance settings do not persist or take effect
### What happened

In **Settings > Appearance > Fonts**, changing **Editor Font Size** or **Data Grid Font Size** has two visible failures:

1. The change is not applied to running editor or data grid views.
2. After quitting and relaunching TablePro, the font size resets to the previous value, even though a `(Custom)` theme appears in the theme list.

Each size change also leaves a new orphan theme JSON in `~/Library/Application Support/TablePro/Themes/` that is never selected as the preferred theme.

### Steps to reproduce

1. Open **Settings > Appearance**, pick a built-in theme (e.g. *Default Light*).
2. In the right pane, switch to the **Fonts** tab.
3. Change **Editor Font > Size** from 13 to 18.
4. Open a SQL editor or browse a table. Font size is unchanged.
5. Quit TablePro and relaunch.
6. Settings still show the original built-in theme; the 18pt change is gone.
7. The theme list now contains a stray `Default Light (Custom)` entry that is not the preferred theme. Repeating step 3 produces another orphan each time.

### Expected behavior

Changing the font size should:

1. Take effect immediately in editor and grid views.
2. Persist across relaunch.
3. Not produce duplicate `(Custom)` themes on every adjustment.

### Environment

- TablePro: `0.43.3` (commit `a6e115da`, current `main`)
- macOS: 14+ / Apple Silicon
- Reproduces on a freshly synced fork of `TableProApp/TablePro@main`.

I can post a PR with the fix if useful.

shouwang0527:
Fixes #1381.

## Summary

Two-file fix for the font size bug:

1. `ThemeEngine.saveUserTheme` now upserts the saved theme into `availableThemes` synchronously before kicking off `reloadAvailableThemes()`. The previous code relied on the async reload completing before the SwiftUI binding setter ran, so the new `(Custom)` theme was missing from the list on the same tick and the picker silently reverted to the original built-in.
2. `ThemeEditorFontsSection.updateFont` no longer calls `engine.activateTheme(copy)` directly. Activation goes through the `onThemeDuplicated` callback, which also writes `preferredLightThemeId` / `preferredDarkThemeId`. Without that, relaunch reconciliation reverted to the built-in theme and the font size change was lost.

Together these fix all three symptoms in the issue: change applies immediately, persists across relaunch, and no orphan `(Custom)` themes accumulate per adjustment.

## Test plan

- [ ] Open Settings > Appearance > Fonts, change Editor Font Size from 13 to 18, confirm SQL editor updates immediately.
- [ ] Quit and relaunch, confirm size is still 18 and the active theme is the saved `(Custom)` theme.
- [ ] Repeat the size change and confirm only one `(Custom)` theme exists in `~/Library/Application Support/TablePro/Themes/`.

Local build skipped (no Xcode locally), relying on CI.


datlechin (approved):
LGTM, thank you.

shouwang0527:
Thanks @datlechin for the quick review and merge!

Since this fixes a regression that makes the editor font feel too small and any adjustment in Appearance > Fonts gets silently reverted, it's a noticeable papercut for daily use. Would it be possible to include this in the next release? Happy to help verify a pre-release build if that's useful.

Thanks again for maintaining TablePro.

PR #1432, opened 2026-05-27
fix(connections): support ~/.aws/config profiles and credential_process for AWS IAM auth

datlechin:
Addresses @imdkbj's report on #1401: AWS IAM "profile" auth only read `~/.aws/credentials` and only static keys, so profiles defined in `~/.aws/config` (SSO, login_session, assume-role, or `credential_process`) were never picked up.

## Changes

- **Read both files.** `resolveProfile` now reads `~/.aws/config` (`[profile NAME]`, and `[default]`) and `~/.aws/credentials` (`[NAME]`), merging per the AWS precedence rule (credentials-file keys win).
- **`credential_process` support.** If the resolved profile has `credential_process`, TablePro runs it and parses the standard JSON (`Version` must be 1, `AccessKeyId`, `SecretAccessKey`, optional `SessionToken`/`Expiration`), matching the AWS CLI/SDK spec. This is the general escape hatch: SSO, IAM Identity Center, and assume-role all work through `aws configure export-credentials --format process`, which is exactly the workaround in the report.

## Implementation notes (to spec, per AWS docs)

- The command is **argv-split respecting double quotes and run via `/usr/bin/env`, not a shell** (no injection; matches "runs the command as specified"). PATH is augmented with the common bin dirs so a GUI-launched app can still find `aws`.
- Runs off the cooperative thread pool (the process can do network I/O for SSO refresh).
- Non-zero exit surfaces the tool's stderr as the error; unsupported `Version` and malformed output are rejected with clear messages.

## Tests

Unit tests for the command tokenizer (plain, double-quoted-with-spaces, blank) and the credential JSON parsing (temporary vs long-term, `Version != 1` rejected, malformed/incomplete rejected). Validated the logic against the AWS spec.

Docs: PostgreSQL and MySQL pages note that profiles can come from `~/.aws/config` and use `credential_process`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - MongoDB: opening a collection no longer crashes when a document contains a NaN or infinite number. (#1418)
 - Opening a saved connection that fails now shows the detailed troubleshooting dialog with suggested fixes, the same one Test Connection shows, instead of a generic error alert. (#1425, #483)
 - Oracle connection errors no longer surface the driver's raw internal message; failures now explain the cause in plain language. (#483)
+- AWS IAM authentication with a named profile now reads `~/.aws/config` (not just `~/.aws/credentials`) and supports `credential_process`, so profiles backed by SSO, IAM Identity Center, or assume-role work through `aws configure export-credentials`. (#1291)
 
 ## [0.45.0] - 2026-05-26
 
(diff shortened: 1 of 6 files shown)


imdkbj:
hey @datlechin 
I think this should be supported in dynamo/redis etc...
image

datlechin:
Hello @imdkbj could you please create an issue for tracking, i’ll take a look later

imdkbj:
Thanks 

https://github.com/TableProApp/TablePro/issues/1567

imdkbj:
Dear @datlechin 
I am still not able to connect via IAM auth while manually entering the temporary token is working fine

```connection to server at "127.0.0.1", port 54321 failed: FATAL:  PAM authentication failed for user "deepak_test"```

PR #1447, opened 2026-05-28
feat(datagrid): auto-detect JSON and PHP serialized values in text cells

datlechin:
## Summary

Cells holding JSON or PHP `serialize()` values in text columns (`TEXT`/`VARCHAR`/`LONGTEXT`) now open in a structured viewer automatically, without requiring the column type to be JSON.

- **JSON-in-text auto-detect**: closes the asymmetric gap where `CellInteractionResolver`'s editable branch sniffed `looksLikeJson` but the read-only branch didn't.
- **PHP serialized read-only viewer**: net-new. Tree view via `OutlineGroup` with `Picker(.segmented)` Tree/Raw toggle, mirroring the existing JSON viewer surface (popover, right-sidebar inspector, pop-out window).
- **Per-column override**: `ValueDisplayFormat` gains `.json` and `.phpSerialized` cases for text columns. `.raw` opts out of detection entirely.

Competitive context: TablePlus #1144 (PHP unserialize) has been open 7+ years with no implementation; Sequel Ace's PHP support is a deprecated Bundle, broken on macOS 12+. TablePlus, Postico, and DataGrip all gate JSON viewing on declared column type (TablePlus #3215, #3477). This PR ships the auto-detect path none of them have.

## Architecture

Detection runs only on cell open, never during grid render (zero render cost). One shared `CellValueContentDetector` consumed by both `CellInteractionResolver` (grid) and `FieldEditorResolver` (right sidebar) so future formats can't reintroduce the asymmetry.

- Detector: O(1) first-scalar pre-filter → `looksLikeJson` (existing) or `PhpSerializeParser.looksLikePhpSerialized` (new). 5 MB cap via `(value as NSString).length`.
- PHP parser: recursive descent over UTF-8 bytes. Handles `N b i d s S a O C r R`; degrades `o` and unknowns to `.unsupported`; PHP `O:` decodes protected (`\0*\0name`) and private (`\0ClassName\0name`) mangling; `C:` is a single leaf (no recursion into Serializable payload); `r:`/`R:` rendered as `→ #N` markers (no cycle traversal). 256 depth cap, 5000 node cap.
- Parse runs off-main via `.task { ... }`; raw mode is always available as the safety net.
- JSON viewer (`JSONViewerContentView`/`JSONViewerView`/`JSONViewerWindowController`) is column-type-agnostic and reused as-is. Existing `JSONTreeParser` preserves source key order.

## Tests

- `PhpSerializeParserTests` — every token, INF/NAN, multi-byte strings, declared-length mismatches, depth cap, reference markers, protected/private mangling decode.
- `CellValueContentDetectorTests` — positive/negative for both formats, 5 MB guard, ambiguous-prefix false positives.
- `ValueDisplayFormatTests` — rawValue, Codable round-trip, `applicableFormats(for:)` membership.
- `FieldEditorResolverTests` — JSON/PHP detection plus override precedence (`.raw` skips both, `.json`/`.phpSerialized` force).
- `CellInteractionResolverTests` — extended with PHP detection cases; the previous `readOnlyJsonLikeTextWithoutTypeReturnsViewInline` test (which documented the bug) is inverted to `...ReturnsViewJson` since that's now the correct behavior.

All new and modified tests pass locally (`xcodebuild test`). Full project build is clean.

## Manual test

A test fixture and walkthrough are in the development notes (10 sample rows covering JSON object/array, deeply nested, PHP scalars, nested arrays, objects with protected/private properties, references, and false-positive guards).

## Notes for review

- **`Localizable.xcstrings` contains some pre-existing WIP entries** (pagination strings, BigQuery dataset strings, AWS credential_process strings, SSH/cloudflared messages) that propagated from the working tree when the branch was created. The PHP-specific keys I added are: `Invalid PHP Serialized Value`, `Maximum depth reached`, `PHP — %@`, `PHP Serialized`, `PHP Viewer`, `private (%@)`, `protected`, `The value could not be parsed. Use raw mode to inspect it as text.`, `Unsupported token: %@`, `Value Too Large`, `Value too large to parse`, `This value is too large to parse. Use raw mode to inspect it as text.` (12 keys). Feel free to split the Localizable changes into a separate commit before merging.
- **Em dashes in window titles** (`PHP — %@`) match the existing `JSONViewerWindowController.swift` convention (`JSON — %@`) and macOS HIG ("Document — App"). CLAUDE.md's anti-em-dash rule is generally for prose; if you want to refactor both window titles to a different separator, that's a small follow-up.
- **`CellInteractionMode.editPhpSerialized`** is defined but unused (the editable branch always returns `.viewPhpSerialized` for PHP, since PHP write-back is out of scope). Easy to remove if you'd rather not carry an unused case.
- **PHP fields in the sidebar force `isReadOnly`**, which hides the field-action menu (NULL/Default/Empty/Function) on those fields. Reasonable since they can't be edited via the viewer, but if you want those actions reachable, easy to gate differently.

Closes the request from https://x.com (public reply asking for auto JSON parse and PHP unserialize).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Oracle Database 11g (11.1 and 11.2) now connects. Previously only 12c and later worked, so 11g servers failed with a "Server Version Not Supported" error. (#1425)
 - Oracle connections can now use a SID instead of a service name. Set Connection Type to SID in the connection form and enter the SID. (#1425)
 - Cmd-click a foreign key arrow to open the referenced table in a new tab instead of the current one. The right-click menu has the same Open in New Tab option. (#1421)
+- Cells holding JSON or PHP serialized values in text columns now open in the structured viewer automatically, without requiring the column type to be JSON.
 
 ### Changed
 
--- a/TablePro/Core/Services/Formatting/CellValueContentDetector.swift
+++ b/TablePro/Core/Services/Formatting/CellValueContentDetector.swift
@@ -1,0 +1,33 @@
+//
+//  CellValueContentDetector.swift
+//  TablePro
+//
+
+import Foundation
+
+internal enum CellValueContent: Equatable {
+    case json
+    case phpSerialized
+    case plain
+}
+
+internal enum CellValueContentDetector {
+    private static let sizeCapBytes = 5_000_000
+
+    static func detect(_ value: String) -> CellValueContent {
+        guard !value.isEmpty else { return .plain }
+        guard (value as NSString).length <= sizeCapBytes else { return .plain }
+
+        let first = value.unicodeScalars.first
+        if first == "{" || first == "[" {
+            if value.looksLikeJson { return .json }
+        }
+
+        let phpFirstScalars: Set<Unicode.Scalar> = ["N", "b", "i", "d", "s", "S", "a", "O", "C", "o", "r", "R"]
+        if let first, phpFirstScalars.contains(first) {
+            if PhpSerializeParser.looksLikePhpSerialized(value) { return .phpSerialized }
+        }
+
+        return .plain
+    }
+}
--- a/TablePro/Core/Services/Formatting/PhpSerializeParser.swift
+++ b/TablePro/Core/Services/Formatting/PhpSerializeParser.swift
@@ -1,0 +1,286 @@
+//
+//  PhpSerializeParser.swift
+//  TablePro
+//
+
+import Foundation
+
+internal indirect enum PhpValue: Equatable {
+    case null
+    case bool(Bool)
+    case int(Int)
+    case float(Double)
+    case string(String)
+    case array([PhpKeyValue])
+    case object(className: String, properties: [PhpProperty])
+    case serializable(className: String, rawPayload: String)
+    case reference(id: Int)
+    case unsupported(token: String)
+    case depthExceeded
+    case tooLarge
+
+    static func == (lhs: PhpValue, rhs: PhpValue) -> Bool {
+        switch (lhs, rhs) {
+        case (.null, .null): return true
+        case (.bool(let a), .bool(let b)): return a == b
+        case (.int(let a), .int(let b)): return a == b
+        case (.float(let a), .float(let b)):
+            if a.isNaN && b.isNaN { return true }
+            return a == b
+        case (.string(let a), .string(let b)): return a == b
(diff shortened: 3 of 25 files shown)


datlechin:
Addressed review feedback in two follow-up commits:

**`7f8663bf` fix(datagrid): respect display-format override before column type and content detection**
- Restructured `CellInteractionResolver.resolve(...)` so user override is consulted first, then declared column type, then content detection. `.raw` now short-circuits to plain text rendering (read-only → `viewInline`, editable → `editInline`/`editOverlay`).
- This fixes the editable-branch bug where `.raw` `break`'d out of the switch but content detection still ran below.
- It also gives the per-column "Display as → Raw Value" picker a consistent meaning: `.raw` always wins, even on declared JSON columns (matches the user's explicit intent).
- New tests: `editableOverrideRawSkipsJson`, `editableOverrideRawSkipsPhp`, `editableOverrideRawMultilineReturnsOverlay`, `readOnlyOverrideRawBypassesJsonColumn`.

**`7dc38a8b` refactor(datagrid): drop unused PHP edit mode, unify parser depth guard, avoid byte copies**
- Removed `CellInteractionMode.editPhpSerialized` and collapsed its dispatch arm in `DataGridView+Click.swift`. PHP is always read-only, so a separate edit case was dead code.
- Removed the redundant `depth >= depthCap` guards inside `parseArray` and `parseObject` — `parseValue`'s `depth <= depthCap` guard is the single source of truth; recursion through `parseValue(depth: depth + 1)` catches the cap at the children entry.
- Switched `parseString`, `parseObject`, `parseSerializable`, and `readUntil` to pass `ArraySlice` directly to `String(bytes:encoding:)` instead of materializing an intermediate `Array`. For a 1 MB PHP string value this avoids a 1 MB byte copy per occurrence.
- New tests: `englishStartingWithPhpTokenChars` covers 10 false-positive English-text starters for every PHP token char (b, i, d, S, O, C, o, r, R, N); `malformedPhpStillDetected` documents that the detector intentionally promotes malformed PHP to the viewer so the user sees the "couldn't parse, use Raw" affordance rather than getting a silent plain-text fallback.

All 52 tests in the 7 affected suites pass. SwiftLint strict is clean. Build is clean.

Not addressed (your call):
- Em dashes in `"PHP — %@"` window title — matches existing `JSONViewerWindowController.swift` convention. Refactoring both is a one-line follow-up if desired.
- `Localizable.xcstrings` pre-existing WIP entries — splitting requires either a separate commit removing them (and restoring later) or rebase + cherry-pick; left to you per the original PR body note.

PR #1459, opened 2026-05-28
fix(datagrid): cell/selection copy, persist hidden columns & filters, JSON popover

J2TeamNNL:
## Summary

- Context menu **Copy** copies the clicked cell value, or the rectangular grid selection when one exists, matching Cmd+C. The row-number column copies the full row. **Copy As** keeps Rows, With Headers, JSON, CSV, Markdown, and the SQL variants.
- Hidden columns persist when you change them and are restored before the first load, so reopening a table keeps them hidden and the query omits them instead of running `SELECT *`.
- Applied filters persist on Apply, so reopening a table restores the last filter.
- Filter bar visibility follows the **Panel State** setting: Always Hide stays closed, Always Show stays open, Restore Last shows the bar only when a saved filter exists.
- JSON detail popover shows long string values up to 300 characters in the tree view (raw value unchanged).

## Test plan

- [ ] Right-click a cell → Copy → pastes the cell value (blob → `0x…`, NULL → `NULL`)
- [ ] Select a range of cells → right-click → Copy → pastes the selection (matches Cmd+C)
- [ ] Right-click the row-number column → Copy → pastes the full row
- [ ] Copy As → Rows / With Headers still work
- [ ] Hide a column, close the tab, reopen the table → column stays hidden and the query omits it
- [ ] Apply a filter, close the tab, reopen → filter restored (Restore Last)
- [ ] Panel State Always Hide → table opens with the filter bar closed
- [ ] Panel State Restore Last with no saved filter → filter bar stays closed
- [ ] Panel State Always Show → filter bar open even with no filter
- [ ] Open a row with a long JSON string → popover shows up to 300 chars, copy gives the full value


datlechin:
Pushed review fixes (commits `15d0198a`, `ac29ba8b`):

- **Persistence:** hidden columns and applied filters now persist on change, so reopening a table restores the hidden-column set (no more `SELECT *`) and the last applied filter. Previously these only saved on tab switch, so closing without switching lost them. Regression tests added.
- **Filter bar default:** the three Panel States are now distinct, resolved in one place (`FilterCoordinator.resolvedRestoredState`). Restore Last shows the bar only when a saved filter exists, instead of popping an empty bar on every table open. Reverted the docs/CHANGELOG line that described the old behavior.
- **Copy menu:** removed the redundant `Copy As → Cell Value` (top-level Copy already copies the cell). Context-menu Copy now honors a rectangular grid selection, matching Cmd+C.
- Deduped the column-scoping query rebuild and removed dead code.

Build and `swiftlint --strict` clean; affected suites pass.

Note: this branch is behind `main` (CHANGELOG entries sit under `[0.46.0]`). Rebase onto `main` before merge so they move to `[Unreleased]`.

PR #1525, opened 2026-05-31
fix(ios): adopt iOS 26 close, cancel, and confirm button roles in sheets

### LINKED ISSUE #1524: iOS 26 Close/Cancel Button UI
### What happened?

The current Cancel / Close button does not seem to follow the native iOS 26 design.

Compared with Apple system apps such as Mail, the button should appear as a compact Liquid Glass-style control with the system-provided shape, material, spacing, and typography. The current implementation looks like a custom flat button, so it feels inconsistent with iOS 26.

Also, since iOS 26, SwiftUI supports semantic button roles for these cases:

```swift
Button(role: .cancel) {
    action()
}

Button(role: .close) {
    action()
}

Button(role: .confirm) {
    action()
}
```

| TablePro.app |  Mail.app |
| - | - |
| Image | Image |

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Database type

None

### TablePro version

_No response_

### macOS version & chip

_No response_

### Screenshots / Logs

_No response_

zunda-pixel:
Closes #1524

iOS sheet dismiss and confirm buttons were custom flat buttons, inconsistent with the native iOS 26 controls used by system apps like Mail. iOS 26 adds semantic button roles that let the system render the Liquid Glass control (shape, material, spacing, typography) automatically.

## Changes

- Add `SemanticButtons.swift` with `CloseButton`, `CancelButton`, and `ConfirmButton`. Each uses the iOS 26 role (`.close` / `.cancel` / `.confirm`) when available and falls back to a titled button on iOS 18.
- Adopt them across the sheets: connection form, insert row, filters, tag/group forms, tag/group management, FK preview, row detail, and the Settings sheet.
- Use `CancelButton` for the Cancel actions in alerts and confirmation dialogs, and for the inline Cancel on the connecting screen.

## Notes

- Deployment target stays at iOS 18. The new roles are gated with `if #available(iOS 26.0, *)`.
- On iOS 26 the role buttons use the system-provided presentation, so the explicit titles ("Done", "Cancel") only render on the iOS 18 fallback path.

| Before | After |
| - | - |
|  image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Destructive queries (DROP, TRUNCATE, DELETE without WHERE) now always ask for confirmation, even with Safe Mode off. (#1481)
 - Table structure changes, table creation, maintenance, column reorder, and saved data-grid edits now follow the connection's Safe Mode and read-only setting. (#1481)
 - AI assistant and MCP queries now follow the same Safe Mode confirmation, read-only, and authentication rules as the editor. (#1481)
+- iOS: sheet close, cancel, and confirm buttons use the native iOS 26 button roles, matching system apps like Mail. iOS 18 keeps titled buttons. (#1524)
 
 ### Removed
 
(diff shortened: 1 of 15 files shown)


datlechin (approved):
LGTM, thank you.

PR #1554, opened 2026-06-02
feat(ai-providers): add Apple Intelligence on-device provider for macOS 26+

### LINKED ISSUE #1048: AI Chat: Apple Intelligence transport via Foundation Models
Add Apple Intelligence as a chat provider on macOS 26+. On-device, no API key, no network. Schema and queries do not leave the Mac.

## What it does

- New `AppleIntelligenceTransport: ChatTransport` backed by the Foundation Models framework (`LanguageModelSession`, `@Generable`, `Tool`).
- Gated by `if #available(macOS 26, *)` and `SystemLanguageModel.default.availability`. Users on older macOS or Intel Macs see it greyed out with the reason: `.deviceNotEligible`, `.appleIntelligenceNotEnabled`, `.modelNotReady`.
- Bridges `ChatTool` (from the chat rewrite) to Foundation Models' `Tool` protocol so the on-device model can call MCP tools.
- Default provider for new users on macOS 26+ when Apple Intelligence is available.

## Depends on

#1047. The `ChatTransport` and `ChatTool` abstractions land there first.

## Out of scope

Bumping the project's macOS minimum. Stays at 14.0. The transport is feature-gated.

datlechin:
Closes #1048.

Adds Apple Intelligence as an on-device AI chat provider on macOS 26 and later, backed by the Foundation Models framework. No API key, no network: schema and queries stay on the Mac. The app's macOS 14.0 minimum does not change; everything is gated behind `#available(macOS 26, *)` and `FoundationModels` is weak-linked.

## What it does

- New `AppleIntelligenceTransport: ChatTransport` over `LanguageModelSession`. It flows through the existing registry, factory, and transport pipeline as a real `AIProviderConfig` (`type: .appleIntelligence`, new `.device` auth style), so the model picker, per-turn attribution, and persistence work with no special-casing.
- Default for new users on macOS 26+ when the model is available (seeded only when no providers exist, so existing setups are untouched).
- A pinned Apple Intelligence row in Settings, AI. When the model is unavailable it is greyed with the reason: needs Apple silicon, Apple Intelligence off in System Settings (with a button to open them), model still downloading, or OS older than macOS 26. The row re-checks while the model downloads.
- One on-device model, so no model list, no reasoning, no image options.

## How the hard parts work

- Streaming: `streamResponse` yields cumulative snapshots, so the transport diffs each against the previous (with a prefix guard) to emit text deltas.
- Tools: the chat tools have runtime JSON schemas. `Tool.Arguments` is `GeneratedContent` and the `GenerationSchema` is built at runtime from our `JsonValue` via `DynamicGenerationSchema`. The framework runs the tool loop itself; each `Tool.call` bridges to the existing approval and execution path through `toolInvocationRequest` and a reply token.
- Availability: a non-gated `AppleIntelligenceStatus` facade wraps the `#available` and `SystemLanguageModel.default.availability` check so the un-gated UI can read availability without importing the framework.
- Multi-turn: the transport is stateless and rebuilds a `Transcript` from the conversation each call, so it needs no reset wiring.

The generic Copilot tool-dispatch functions were renamed to provider-neutral names since this transport reuses them.

## Tests

New tests cover the availability facade mapping, the schema bridge, seeding idempotency and no-override, and the factory guard (an Apple Intelligence config never falls back to an OpenAI-compatible transport). Foundation Models tests are gated to macOS 26.

## Before merge

- Needs a build on Xcode 26. The `Transcript` initializers in `AppleIntelligenceTransport.buildTranscript` were written against Apple's docs but not compiled here.
- The System Settings deep link (`com.apple.Siri-Settings.extension`) is inferred; it falls back to the Settings root if the identifier is wrong. Worth confirming on a real macOS 26 machine.


datlechin:
Closing this. After getting the on-device transport working end to end, Foundation Models isn't a good fit for TablePro's agentic AI chat, and the blocker is structural rather than a bug we can patch.

## What works
Plain chat and single-turn tool calls run fine. Sign-in and availability detection, the transcript building, schema generation, and the tool bridge all work on macOS 26.

## Why it can't ship
The on-device model has a fixed ~4k-token context window. TablePro's AI chat sends a large system prompt plus every tool's JSON schema, and each tool result (for example `list_tables` with row counts on a 30+ table database) gets added back into the transcript. That fills the window after one or two agentic turns, and the next generation fails with `GenerationError.exceededContextWindowSize`.

Making it usable would mean trimming conversation history and truncating tool outputs for this provider only. That degrades answers: the model would see partial results and forget earlier turns. It isn't a tradeoff worth shipping when every other provider runs the same agent flow without it.

## Worth keeping for a future attempt
Two findings from this work (the code stays on the `feat/1048-apple-intelligence-transport` branch for reference):

- Tool definitions must be declared in the `Transcript.Instructions` entry, not just passed via the session's `tools:` parameter. Without that, multi-pass tool calling dies mid-stream with an opaque `GenerationError -1`.
- Foundation Models wraps tool-call failures in `ToolCallError`, so `error as? GenerationError` misses the real cause. You have to unwrap `ToolCallError.underlyingError` to get the actual reason, which is how the context-window error above was finally identified.

Worth revisiting if a future macOS release grows the on-device context window, or if we add a context-condensation strategy for small-window models. Reopening then.

PR #1558, opened 2026-06-02
fix(plugin-mssql): send database in login packet so DB-scoped logins connect

brunorodmoreira:
## Problem

Connecting to an Azure SQL Database fails with `Login failed for user '...'` (SQL Server error 18456) when the login can only reach a single database, for example an Azure SQL contained user. The same credentials work in DBeaver and other clients.

The driver set every login property except the database, connected, then ran `USE [database]`:

```swift
guard let proc = dbopen(login, serverName) else { ... }   // authenticates against master
if !options.database.isEmpty {
    if dbuse(proc, options.database) == FAIL { ... }       // USE runs only after login
}
```

A login scoped to one database has no rights in `master`, so `dbopen` is rejected before `USE` ever runs. Microsoft's own driver (and DBeaver, via JDBC) avoid this by putting the database in the login packet, the same as an ODBC `Database=` / `Initial Catalog`.

## Fix

Set the database on the login record with `DBSETDBNAME` before `dbopen`, so FreeTDS sends it in the LOGIN7 packet and the server authenticates directly against it. The post-connect `dbuse` is removed: it is the exact step that cannot work for a scoped login, and it is redundant once the database is in the login packet. `switchDatabase()` keeps using `dbuse`, which is correct for an already authenticated session.

This is more correct for every SQL Server, not just Azure, and matches what the official drivers do.

## How it was diagnosed

Reproduced through the same FreeTDS library (pymssql) against a live Azure SQL server: login with the database in the packet succeeds, login without it fails with the identical 18456. The macOS unified log showed `FreeTDS msg 18456 sev 14` from the plugin at the `dbopen` stage.

## Tests

Extracted the login parameter assembly into `MSSQLLoginParameters`, a pure builder with no FreeTDS dependency, and added `MSSQLLoginParametersTests`: the database is included when set, omitted when blank, and the credentials and encryption flag carry through. `FreeTDSConnection` maps each field to its `DBSET*` constant.

## Checklist

- [x] Tests added
- [x] `CHANGELOG.md` updated under `[Unreleased]`
- [x] Docs updated (`docs/databases/mssql.mdx`)
- [x] No new user-facing strings
- [x] swiftlint --strict clean on changed files


brunorodmoreira:

datlechin (approved):
LGTM, thank you

PR #1566, opened 2026-06-03
fix(connections): skip password prompt for AWS IAM connections

antoncazalet:
## Summary

- AWS IAM connections prompted for a password on connect. IAM supplies the credentials, so the prompt was never needed and only blocked the connection flow.

## Changes

- Gate the password prompt on `!usesAWSIAM` in the health-check reconnect path (`DatabaseManager+Health`) and in session creation (`DatabaseManager+Sessions`).
- Stop treating IAM connections as prompt-for-password in the connection form (`ConnectionFormCoordinator`).

## Test plan

- [ ] Connect an AWS IAM (RDS) connection and confirm no password prompt appears.
- [ ] Confirm a normal prompt-for-password connection still prompts.
- [ ] Confirm an IAM connection auto-reconnects after a dropped session without prompting.


antoncazalet:

datlechin (approved):
Thank you.

PR #1572, opened 2026-06-03
feat(plugins): config profiles, credential_process, SSO, AssumeRole, ElastiCache, Keyspaces (#1567)

### LINKED ISSUE #1567: Feat request: support ~/.aws/config profiles and credential_process for AWS IAM auth in dynamo/redis etc
### Problem

Image

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Related database type

None

datlechin:
Fixes #1567. AWS-backed connections could not use profiles defined in `~/.aws/config` (the screenshot in the issue: a DynamoDB "AWS Profile" connection failing with "Profile not found or incomplete in ~/.aws/credentials"). The root cause was that each AWS plugin reimplemented credential resolution, and the copies only read `~/.aws/credentials`. This unifies AWS credential resolution into one shared layer in `TableProPluginKit` and uses it everywhere, then extends it to full AWS-CLI parity across all AWS-backed connections.

## What changed, by stage

1. **Foundation.** Moved `AWSCredentials`, `AWSCredentialResolver`, `AWSSSO`, `AWSAuthError`, plus new `AWSConfigFile`/`AWSSigV4`/`AWSSTS` into `TableProPluginKit/AWS/` as public API so the app and every plugin share one implementation. Additive and ABI-safe (no `currentPluginKitVersion` bump). `credential_process` (which uses `Process`) is gated to macOS so the iOS PluginKit target still builds. Credentials now carry an `expiration` for refresh.
2. **DynamoDB (the reported bug).** Deleted the in-plugin resolver and uses the shared one: reads `~/.aws/config` + `~/.aws/credentials`, supports `credential_process` and SSO, re-resolves on expiry, and shows accurate errors.
3. **Profile picker.** The connection form shows an editable combo box of profiles discovered from `~/.aws/config` + `~/.aws/credentials` (normalizing `[profile X]`/`[X]`), with a free-text fallback. Implemented as an additive `ConnectionField.dynamicOptions` (no `@frozen` change).
4. **AssumeRole + in-app SSO.** `role_arn` + `source_profile` chaining via STS AssumeRole (with `external_id`, `duration_seconds`, `credential_source = Environment`). When an SSO session has expired, a prompt runs the OIDC device-login flow in the browser and refreshes the cached token.
5. **ElastiCache Redis IAM.** New AWS IAM auth mode on Redis; TablePro generates the ElastiCache IAM token and uses it as the password.
6. **Keyspaces / Cassandra SigV4.** SigV4 challenge-response authenticator wired into the cpp-driver, matching AWS's documented Keyspaces algorithm.

## Tests
Unit tests for every deterministic/protocol piece: dual-file profile resolution (incl. the config-only case that reproduces the bug), SSO parsing/token-cache/role-fetch, STS AssumeRole response parsing, SSO device-login parsing/polling/cache, and the RDS/ElastiCache/Keyspaces token generators. All AWS suites pass; SwiftLint `--strict` clean on changed lines.

## Needs live-AWS verification (please test before release)
The deterministic logic is unit-tested against AWS's documented algorithms, but these live paths can't be verified in CI and need real AWS:
- **ElastiCache** — a live IAM-enabled cache over TLS; the replication-group-id input may need tuning per cluster type.
- **Keyspaces** — the SigV4 handshake against a live Keyspaces endpoint (signer matches the reference algorithm; cpp-driver callbacks compile).
- **In-app SSO device login + MFA** — a real SSO portal and browser hand-off. MFA-protected assume-role profiles intentionally return a clear "not supported yet" message.

## Release
DynamoDB and Cassandra are registry-only plugins and need a registry re-release after merge; Redis is bundled. The PluginKit changes are additive (no version bump).


datlechin:
Addressed the review and the failing **PluginKit ABI Gate**:

- **ABI (was the real failure):** the gate flagged that `ConnectionField.init` had gained a `dynamicOptions` parameter — a breaking change for already-compiled registry plugins that call that initializer. Reworked it: the original init is untouched, `dynamicOptions` is now an additive optional `var` set via a `withDynamicOptions(_:)` chaining method. Re-ran `scripts/check-pluginkit-abi.sh` against the PR base: the remaining interface diff is **purely additive** (no removed/changed symbols), so **no `currentPluginKitVersion` bump is needed**. The gate stays red because it flags any interface change for human review; this one is additive.
- **ElastiCache (High):** removed the silent fallback to the endpoint host. The cache name (replication group ID) is now a required field, and connecting without it (or without TLS) fails with a clear, actionable error instead of signing a token with the wrong host.
- **TLS enforcement:** ElastiCache and Keyspaces IAM now error early if SSL is disabled (both require in-transit encryption).
- **Field unification:** the relational plugins' inline AWS field list is gone; MySQL/PostgreSQL/Redis/Cassandra all use the shared `AWSAuthFields.standard()`, removing the duplication.
- **AssumeRole:** `role_session_name` is now sanitized to the STS-allowed character set.

Build green, all AWS/ConnectionField suites pass, SwiftLint `--strict` clean on changed lines.

PR #1578, opened 2026-06-04
feat(plugins): import data from CSV and TSV files into a table (#1568)

### LINKED ISSUE #1568: Feature Request: Support import from csv
### Problem

Hey @datlechin 

It would be awesome to support the csv import.

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Related database type

PostgreSQL

datlechin:
Closes #1568.

Adds CSV and TSV import into a database table for any SQL target.

## What it does

Pick File > Import > From CSV and choose a `.csv` or `.tsv` file. The row import sheet opens with CSV parsing options. Map columns to an existing table, or create a new table with inferred, editable types.

## Options

- Delimiter: auto-detect, comma, semicolon, tab, pipe
- Quote character: double or single
- Encoding: auto-detect, UTF-8, ISO Latin 1, Windows-1252
- First row is a header
- Trim leading and trailing spaces
- Treat empty values as NULL, plus an optional NULL token
- On error, wrap in transaction, delete existing rows (shared with SQL and JSON import)

Changing any dialect option re-reads the file so the column mapping reflects it.

## How it works

- Reuses the existing CSV parser. `CSVStreamingParser`, `CSVDialect`, and `CSVTypeInferrer` move from `CSVInspectorPlugin` into `TableProPluginKit` as public types, so the importer and the inspector share one RFC 4180 tokenizer (quoted commas and newlines, doubled quotes, BOM, delimiter and encoding detection). Additive PluginKit ABI, no version bump.
- New `CSVImportPlugin` bundle memory-maps the file, indexes rows, and inserts in 500-row parameterized batches. Memory is bounded by the row-range index, not the row data. Cancellable per batch, wrapped in a transaction.
- `JSONImportSheet` becomes `RowImportSheet`, shared by JSON and CSV. The format plugin supplies the icon, name, and options view. `JSONImportTypeMapper` becomes `ImportTypeMapper`. A new `fieldDetectionSignature` hook (additive, defaults to empty) drives live re-detection when an option changes.

## Tests

- New `CSVImportPluginTests`: dialect resolution, header and header-less naming with dedup, NULL/empty/trim handling, ragged rows, type mapping, quoted/embedded/doubled-quote parsing, semicolon auto-detect.
- The moved CSV parser suites and the JSON import suites still pass after the move. 116 cases green.

## Notes

- Inserts are parameterized, so CSV values are never concatenated into SQL.
- Docs updated in `docs/features/import-export.mdx`; CHANGELOG entry under Unreleased.


datlechin:
Pushed 8578ce16 addressing the review findings:

**Shared import loop extracted to PluginKit.** New `RowImportRunner` owns the transaction lifecycle, batching, error modes, and progress for both JSON and CSV import (additive ABI, no version bump). The duplicated `performImport`/`flush`/`insert` bodies in both plugins are gone.

**Behavior fixes (apply to JSON import too, CHANGELOG entries added):**
- "Delete existing rows" now runs inside the transaction, so a failed import restores the deleted rows.
- Skip-and-continue inserts row by row from the start instead of retrying a partially written batch, which could duplicate rows (`insertRows` is multi-statement when chunked by bind-parameter limits).
- "Stop and Commit" commits the rows inserted before the error instead of rolling them back. Cancellation still rolls back in every mode.

**Review finding 3:** field detection now applies the trim and NULL-token options to samples, so inferred types and sample values match what the import inserts. The detection signature keys are all meaningful now.

**Review finding 2:** `decode` keeps the lossless Latin-1 fallback but drops the unreachable `?? ""` for an explicit total byte-to-scalar mapping.

Also: blank CSV rows now advance the progress bar, and the unused plugin loggers are removed.

Tests: 7 new `RowImportRunnerTests` with a mock sink cover transaction ordering, rollback/commit per mode, no-batch-retry in skip mode, the error cap, and cancellation. 2 new detection tests cover trim and NULL-token sampling. `CSVImportPluginTests`, `JSONImportPluginTests`, and `CSVInspectorTests` all pass (102 cases across the affected suites). `swiftlint lint --strict` clean.

PR #1581, opened 2026-06-04
feat(editor): complete database, schema, and table segments in qualified names

xantiagoma:
## Summary

Improves SQL autocomplete for connections that organize objects as database → schema → table (BigQuery today, Snowflake with #1580):

- For hierarchical-schema plugins the completion feed previously received an empty table list (`SchemaService.runHierarchicalLoad` publishes `.loaded([])` and per-schema tables never reached `SQLSchemaProvider`); it now receives the union of all loaded per-schema tables plus schema and database names
- Qualified names complete at every segment: `ANALYTICS_PR` → database, `ANALYTICS_PROD.` → schemas, `ANALYTICS_PROD.DBT_MARTS.` → tables — including schemas never expanded in the sidebar, fetched on demand through the metadata pool
- Alias column completion respects schema-qualified tables: `FROM db.schema.orders o WHERE o.` previously fetched columns from the *active* schema and came back empty; `TableReference` now carries the schema segment end to end
- The completion popup includes the active connection's dialect function list — previously dialect functions only fed syntax highlighting while the popup used a small generic set. This benefits every database plugin (e.g. `SAFE_CAST` on BigQuery, `JSONB_AGG` on PostgreSQL)

No behavior change for flat-schema connections beyond the dialect functions appearing in completions.

## Testing

- Live-verified against Snowflake and BigQuery-style hierarchical browsing: segment completion, on-demand schema tables, schema-qualified alias columns, dialect functions in the popup
- All five autocomplete test suites pass in isolation (605 cases); the order-dependent failures of `testProviderAcceptsDatabaseType` / `testMySQLProviderTypes` when suites run together reproduce identically on `main` and predate this PR
- `swiftlint lint --strict` clean on all changed files

## Checklist

- [x] Tests updated
- [x] `CHANGELOG.md` updated under `[Unreleased]`
- [x] Docs updated (`docs/features/autocomplete.mdx`)
- [x] User-facing strings localized (n/a — no new UI strings)
- [x] No SwiftLint violations

Related: #1420, #1580

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1586, opened 2026-06-04
fix(editor): capture quoted identifiers in qualified table references

xantiagoma:
## Summary

Fixes the `SQL Context Analyzer` failure currently red on `main` ("Strips quoting from the captured schema segment", added in the #1581 squash).

The table-reference regexes captured names with `([\w.]+)`, which cannot cross a quoted segment — `FROM "sales".orders` captured only `sales`, so the schema was lost and the test failed. The capture class now includes quote characters, and both the parsed table name and the schema segment strip `` ` ``, `"`, and `'`:

```sql
SELECT * FROM "sales".orders      -- schema=sales, table=orders ✓
SELECT * FROM db."schema"."tbl" t -- schema=schema, table=tbl, alias=t ✓
```

## Testing

- `SQLContextAnalyzerTests` (including the four schema-segment tests) and `SQLContextAnalyzerCaseInsensitiveTests` pass locally
- `swiftlint lint --strict` clean

Related: #1581, unblocks CI for #1580

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/TablePro/Core/Autocomplete/SQLContextAnalyzer.swift
+++ b/TablePro/Core/Autocomplete/SQLContextAnalyzer.swift
@@ -247,14 +247,14 @@ final class SQLContextAnalyzer {
 
     private static let tableRefRegexes: [NSRegularExpression] = {
         let patterns = [
-            "(?i)\\bFROM\\s+[`\"']?([\\w.]+)[`\"']?" +
+            "(?i)\\bFROM\\s+[`\"']?([\\w.`\"']+)[`\"']?" +
             "(?:\\s+(?:AS\\s+)?[`\"']?([\\w]+)[`\"']?)?",
             "(?i)(?:LEFT|RIGHT|INNER|OUTER|CROSS|FULL)?\\s*(?:OUTER)?\\s*JOIN\\s+" +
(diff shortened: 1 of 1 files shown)


xantiagoma:
The failing check is unrelated to this PR: the SQL Context Analyzer suite (which this PR fixes) now passes here.

The remaining failure is the ExecuteQueryTool suite, which has been failing identically on every main run since ad82403 (fix(editor): make statement splitting dollar-quote aware for PostgreSQL). Same two issues both there and here: `error.code` is `-32602` instead of `requestCancelled` (`-32002`), and progress notifications never fire (`methods.count == 0`).

First failing main run with the same errors: https://github.com/TableProApp/TablePro/actions/runs/26887465313

Happy to look into that one separately if useful.

datlechin:
Hello @xantiagoma, pr look good to me, could you please resolve conflict

datlechin:
Thanks for the fix. Closing as superseded: #1587 (merged 2026-06-05) rewrote the same table-ref regexes with a per-segment grammar and per-segment quote trimming, and the failing test this PR targeted ("Strips quoting from the captured schema segment") now passes on main. Resolving the conflict would leave this PR with an empty diff.

PR #1592, opened 2026-06-05
fix(editor): sync external text binding changes into the source editor

### LINKED ISSUE #1576: JSON View needs a re-click
### What happened?

Selecting to view row data as JSON needs a re-click from Tree to Text

### Steps to reproduce

1. Select a row
2. Click on JSON
3. Observe the screen is blank
4. Click on Tree and Again Text
5. Now data is there as JSON

### Expected behavior

JSON viewer should render the data after clicking on JSON navigation

### Database type

None

### TablePro version

0.48.0(94)

### macOS version & chip

_No response_

### Screenshots / Logs

_No response_

datlechin:
Fixes #1576

## Problem

Selecting rows and switching the results view to JSON showed a blank editor. The data only appeared after clicking Tree and then Text, which forces SwiftUI to recreate the editor.

## Root cause

`SourceEditor` set the text binding's value once in `makeNSViewController` and never again. `updateNSViewController` only handled cursor, scroll, find panel, and configuration, so a binding that changed after creation never reached the text view. `ResultsJsonView` computes its JSON asynchronously, so the editor was always created with an empty string and stayed empty.

The same gap made `DDLTextView`'s `.onChange(of: ddl)` writeback a no-op, and stale JSON stayed on screen when the row selection changed while in text mode.

## Fix

- The coordinator tracks `lastSyncedText`, the last value synced in either direction. `updateNSViewController` now pushes the binding value into the text view when it differs. The comparison is O(1) when nothing changed because the unchanged binding returns the same string instance.
- User typing is never clobbered: `textViewDidChangeText` records `lastSyncedText` on both the immediate and the debounced (>500k chars) writeback paths, and a pending debounce is cancelled before an external value is applied. Programmatic `setText` runs under `isUpdatingFromRepresentable`, which the change notification handler now respects.
- `ResultsJsonView` shows a small spinner until the first JSON formatting pass finishes, instead of an empty editor in text mode or an "Invalid JSON" flash in tree mode. Recomputes triggered by selection changes keep the previous content visible until the new result is ready.

## Tests

- New `SourceEditorBindingSyncTests` (5 tests, all passing): external change syncs down, in-flight user edits are not clobbered, typing writes back and records the synced value, notifications during programmatic sync do not write back, and same-value syncs leave the text storage untouched.
- Repaired the package test target, which did not compile: `TreeSitterClientTests`, `HighlighterTests`, and `HighlightProviderStateTest` referenced `CodeLanguage.swift` and `.c`, which the vendored `CodeEditLanguages` no longer ships. They now use `.sql` and `.json`.
- No UI automation: the flow needs a live database connection with result rows, which does not run deterministically in `TableProUITests`.

## Verification

`swiftlint lint --strict` passes on the changed app file; the package's SwiftLint build plugin passes. Manual check: select rows, switch to JSON, data renders immediately; changing the selection updates the text view in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The JSON results view shows row data right away instead of staying blank until you switch between Tree and Text, and it updates when the row selection changes. A spinner shows while large results are being formatted. (#1576)
 - Double-clicking or pressing Enter on a JSON cell now edits its value inline, like other cells; on a blob cell it opens the hex editor. The chevron still opens the tree or hex editor. Neither cell responded to double-click before. (#1588)
 - Query results appear as soon as the database returns rows. Column defaults, foreign keys, and row counts now load in the background instead of holding up the grid, which removes a multi-second wait on remote databases whose system tables are slow to query. (#1574)
 - MySQL and MariaDB queries are ready to edit right away. Primary key and nullability come back with the rows, so an editable query no longer waits on a separate metadata query. (#1574)
--- a/LocalPackages/CodeEditSourceEditor/Sources/CodeEditSourceEditor/SourceEditor/SourceEditor+Coordinator.swift
+++ b/LocalPackages/CodeEditSourceEditor/Sources/CodeEditSourceEditor/SourceEditor/SourceEditor+Coordinator.swift
@@ -17,6 +17,7 @@ extension SourceEditor {
         var isUpdatingFromRepresentable: Bool = false
         var isUpdateFromTextView: Bool = false
         var text: TextAPI
+        var lastSyncedText: String?
         @Binding var editorState: SourceEditorState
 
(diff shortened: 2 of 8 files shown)


datlechin:
Follow-up to the self-review notes, in 337b3798:

- `Coordinator.init` now captures `lastSyncedText` from the binding, so there is no nil window between coordinator creation and `makeNSViewController`.
- `updateNSViewController` refreshes `coordinator.text`, and the debounced writeback reads the binding at fire time instead of the one captured when typing started.
- Copy JSON is disabled until the first formatting pass finishes, so it can no longer copy an empty string.
- Two new tests: the >500k debounce path (no writeback before the delay, no clobber from an interim sync, writeback lands after) and out-of-bounds selection dropping when a shorter external text arrives. 7 tests total, all passing.
- Kept `setText` over `replaceCharacters` on purpose: `replaceCharacters` is the user-edit path (gated on `isEditable`, runs mutation filters, fires suggestion triggers). The doc comment on `syncBindingText` records this.

PR #1615, opened 2026-06-07
fix(editor): re-highlight the document after a programmatic text replacement (#1612)

### LINKED ISSUE #1612: Code highlighting is lost after the code is formatted
### What happened?

Code highlighting is lost after the code is formatted

### Steps to reproduce

1. Create a new query label
2. Enter the query statement
3. Click the Format button in the upper right corner to format the code

### Expected behavior

Keep the code highlighted.
And why is there only code beautification but no code minimization function?

### Database type

MySQL / MariaDB

### TablePro version

0.49.1

### macOS version & chip

26.5.1

### Screenshots / Logs

_No response_

datlechin:
Fixes #1612.

## Problem

Clicking Format replaced the query text and all syntax highlighting disappeared, never recovering. With #1611 merged, Format flows exclusively through the binding push-down (`syncBindingText` -> `setText`), which made the loss deterministic.

## Root cause, two layers

1. `syncBindingText` called the TextView-level `textView.setText(...)`, bypassing `TextViewController.setText(_:)`, which is the documented controller-level path and rebuilds the highlighter for the replacement storage. The old highlighter kept tree-sitter state for the dead document.
2. Rebuilding alone is not enough: a fresh `Highlighter` only queries when triggered (an edit, a frame/bounds notification, or `invalidate()`). Initial load works because the first layout pass fires `frameDidChangeNotification`; after a mid-session storage swap no trigger is guaranteed, so the rebuilt highlighter sat idle and the document stayed plain.

## Fix

- `syncBindingText` routes through `controller.setText(newValue)` so the highlighter is rebuilt for the new storage.
- `setUpHighlighter()` now ends with `highlighter.invalidate()`, making "after setup, the editor is highlighted" an invariant for every caller instead of a timing accident.

## Tests

- `test_syncRebuildsHighlighterForReplacementStorage`: the highlighter instance is rebuilt on a binding push-down.
- `test_syncQueriesHighlightsForReplacementStorage`: a mock provider proves the rebuilt highlighter is re-set-up AND re-queried; this test fails with the route-through change alone.
- All existing binding-sync and highlighter tests pass.

## Note

The issue also asks for a minify counterpart to Format; that is a feature request and should be tracked separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Cancelling a SQLite query no longer races a disconnect happening at the same moment. (#1610)
 - Typing in the query editor no longer erases characters or drops focus on each keystroke, a timing-dependent bug most visible on macOS 15. (#1608)
 - The autocomplete popup now filters in place as you type instead of closing and reopening on every keystroke. (#1608)
+- Syntax highlighting no longer disappears after formatting a query. (#1612)
 
 ## [0.49.1] - 2026-06-06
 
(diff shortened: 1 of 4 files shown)


datlechin:
Folded into #1616, which now targets main and carries both the fix and the refactor.

PR #1647, opened 2026-06-10
fix(datagrid): Delete multi-select, filter panel redesign, Cmd+R refresh

J2TeamNNL:
## Summary

- **D12A** — Delete key now reads `gridSelection.affectedRows` first before falling back to `selectedRowIndexes`, matching the existing `copy()` pattern. `validateUserInterfaceItem` and `deleteSelectedRowsIfPossible` updated accordingly.
- **D12B** — `rightMouseDown` override in `KeyHandlingTableView`: when the clicked row is already inside the selection, the override skips `super` (which would collapse the selection) and calls `NSMenu.popUpContextMenu` directly.
- **D13** — Filter panel redesign per owner spec: "Unset" renamed to "Clear" (calls `clearAppliedFilters()` — keeps filter rows, returns table to unfiltered); per-row Apply/Applied buttons removed; "Apply Only This Filter" moved to each row's context menu; tri-state checkbox added to panel header to toggle all rows enabled/disabled; "Remove All Filters" added to the `...` options menu (calls `clearFilterState()`).
- **D14** — After `currentQueryTask?.cancel()`, immediately set `currentQueryTask = nil` and reset `tab.execution.isExecuting = false` so `runQuery()` is not blocked by the guard check.

## Test plan

- [ ] Select multiple rows with Shift/Cmd click, press Delete — all selected rows should be marked for deletion
- [ ] Select cells across rows via drag (grid selection), press Delete — affected rows should be marked for deletion
- [ ] Select multiple rows, right-click a row inside the selection — context menu appears without collapsing the selection
- [ ] Filter panel: click "Clear" — filter rows remain, table reloads without filters applied
- [ ] Filter panel `...` menu: click "Remove All Filters" — all filter rows removed, table reloads without filters
- [ ] Filter panel: right-click a row, choose "Apply Only This Filter" — only that row's filter is applied
- [ ] Filter panel header checkbox: with mixed enabled state shows dash icon; clicking enables all; clicking again disables all
- [ ] While a query is executing, press Cmd+R — query cancels and a fresh query starts immediately

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Switcher, menus, and alerts now use each database's own container name: Dataset for BigQuery, Keyspace for Cassandra and ScyllaDB. (#509)
 
+### Changed
+
+- Filter panel header "Unset" renamed to "Clear": it now keeps filter rows in place and only removes the applied state, returning the table to unfiltered results. Use "Remove All Filters" in the filter options menu to discard all filter rows at once.
+- Per-row Apply and Applied buttons removed from the filter panel; "Apply Only This Filter" is now in each row's right-click context menu.
+- A tri-state checkbox in the filter panel header toggles all filter rows enabled or disabled at once.
+
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


J2TeamNNL:
Splitting into separate PRs per fix (D12, D13, D14).

PR #1665, opened 2026-06-12
fix(plugin-postgresql): resolve effective schema when public does not exist

### LINKED ISSUE #1662: No database shown if schema "public" does not exist
### What happened?

On a single Postgres instance I'm running different databases
- database A with schema foo
- database B with schema bar
- database C with schema baz

When the "public" schema does not exist (for a database) I can NOT see/select any tables.
So I have to add an empty "public" schema for the schema select to appear.
Then I'm able to select e.g. the schema "foo" and see the tables for database A

### Steps to reproduce

See above

### Expected behavior

If there's only a single schema available use this automatically instead of relying on "public"

### Database type

PostgreSQL

### TablePro version

_No response_

### macOS version & chip

_No response_

### Screenshots / Logs

_No response_

datlechin:
Fixes #1662

A PostgreSQL database with no `public` schema showed no tables and no schema selector. The user had to create an empty `public` schema to get the picker to appear.

Two defects combined:

- `LibPQDriverCore.currentSchema` was hardcoded to `public` and only updated when `SELECT current_schema()` returned non-NULL. PostgreSQL returns NULL there when neither `public` nor a `$user` schema exists (default `search_path` is `"$user", public`), so every metadata query ran against a schema that does not exist.
- The schema picker only rendered with more than one schema, so a single-schema database left no way to switch.

Changes:

- `connect()` now falls back when `current_schema()` is NULL: first `current_schemas(false)[1]` (the first existing search_path entry, which honors `ALTER ROLE/DATABASE SET search_path`), then the first user schema alphabetically. On fallback it issues `SET search_path` so editor queries resolve to the same schema the sidebar shows. Runs before `onPostConnect`, so the metadata connection pool inherits the right schema.
- The schema picker shows whenever at least one schema is known, still gated to schema-switching drivers.
- `setSearchPath` no longer appends `, public` when switching schemas. Note: unqualified calls to extension functions installed in `public` (uuid-ossp, PostGIS) now require qualification after switching to another schema.
- The database list table count now counts all user schemas instead of only `public`.

CockroachDB and Redshift share `LibPQDriverCore` but always have a `public` schema, so the fallback never triggers for them. No PluginKit types touched, no ABI change.

Tests: `PostgreSQLSearchPathTests` (updated for the new SET SQL), new `PostgreSQLDefaultSchemaFallbackTests` and `SchemaPickerVisibilityTests`. All pass locally; `swiftlint lint --strict` clean.

DBeaver had this same bug and fixed it the same way in 7.1.4 (reading the effective search_path instead of assuming `public`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- PostgreSQL databases without a "public" schema now load tables from the first available schema, the schema selector also appears when only one schema exists, and the database list counts tables in every user schema instead of only "public". Switching schemas no longer keeps "public" on the search path. (#1662)
 - Format Query can now be undone with Cmd+Z; the formatting is applied as a single editor edit instead of clearing the undo history. (#1645)
 - Format Query now formats only the selected text when a selection is active, and the full query when nothing is selected. (#1656)
 - Foreign key jump arrows no longer disappear after sorting, filtering, or paginating a table, and a failed foreign key lookup is retried on the next load instead of hiding the arrows for the whole session.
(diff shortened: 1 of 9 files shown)


datlechin:
Follow-up from self-review (4f922c8a):

- Schema probe now distinguishes a failed query from a NULL result. A transient error on `SELECT current_schema()` keeps the prior schema instead of triggering the fallback chain and overwriting the search path.
- Fallback queries are injected per driver: Redshift uses its `USAGE`-filtered `pg_namespace` list, so the chosen default is always readable by the connected role. PostgreSQL and CockroachDB keep the `information_schema.schemata` list.
- CHANGELOG: the search-path behavior change moved to its own `Changed` entry with the extension-function caveat spelled out, separate from the `Fixed` entry.
- New tests: `PostgreSQLSchemaProbeTests` (5 cases for the failure/NULL/text mapping) and Redshift fallback ordering. All targeted suites pass, swiftlint strict clean.

PR #1700, opened 2026-06-17
perf(sidebar): load tree tables before routines to cut database expand latency

desperadoxhy:
## What

Expanding a database in the tree sidebar now loads tables first and fills in procedures and functions in the background.

## Why

`loadObjects` fired three queries through `async let`: `fetchTables`, `fetchProcedures`, `fetchFunctions`. All three ran on the same driver (one libpq / mariadb connection), and `MetadataConnectionPool.runSerially` chains same-connection work into strict order, so the `async let` parallelism was lost. The tree list only appeared after all three round-trips finished in sequence, and most schemas have no routines at all.

## Change

- Split `DatabaseTreeMetadataService.objects` into `tablesState` and `routinesState`, each with its own `OnceTask` dedup.
- `loadTables` runs only `fetchTables`. The tree renders as soon as it returns, one round-trip.
- `loadRoutines` runs procedures and functions on a separate `MetadataConnectionPool` connection with `workload: .bulk`, physically isolated from the tables query so they no longer block it.
- `objectsContent` keys off `tablesLoadState`; routines load via a `.task` after tables render and append when ready. `MetadataLoadState.isSettled` (added) keeps the empty-state from spinning forever if routines fail.
- `refreshObjects`, `handleDisconnect`, and `resetPending` updated for the two split dictionaries.

## Verification

- `xcodebuild build` (Debug, Xcode 27 beta): BUILD SUCCEEDED.
- `SchemaServiceTests`: 3/3 pass, no regression.
- `MetadataLoadStateTests` (new): 3/3 pass, covers `isSettled`.

No unit test for `DatabaseTreeMetadataService.loadTables/loadRoutines` itself: the service reads `DatabaseManager.shared` and `MetadataConnectionPool.shared` singletons and cannot take an injected driver (unlike `SchemaService`, which the existing tests cover). Decoupling that is a separate refactor.

## Notes

Built with Xcode 27 beta (`DEVELOPER_DIR=/Applications/Xcode-beta.app`). Worth a second pass on a stable Xcode before merge.


desperadoxhy:

datlechin (approved):

PR #1707, opened 2026-06-17
fix: keep sidebar filter when opening a table from the filtered list

### LINKED ISSUE #1690: Filter table issue
### What happened?

When I filter a table and then double-click on a table name, the search filter is removed

### Steps to reproduce

1. Using the sidebar filter, filter the table
2. Double-click on a table 
3. The selected table is displayed, but the search filter is removed

### Expected behavior

_No response_

### Database type

PostgreSQL

### TablePro version

0.51.0 (98)

### macOS version & chip

Mac M1

### Screenshots / Logs

_No response_

mvanhorn:
## Summary

In the tables sidebar, typing a filter and then double-clicking a table to open it cleared the active filter text. The table opened but the filter was wiped.

Root cause is in `SidebarContainerViewController`. The `NSSearchFieldDelegate.searchFieldDidEndSearching(_:)` callback unconditionally wrote `""` back to the persisted search text. When a table is opened, focus leaves the search field, AppKit fires `searchFieldDidEndSearching`, and the filter was cleared even though the field still held text.

The fix guards that callback so it only clears the persisted text when the field's `stringValue` is actually empty (the cancel/clear path leaves an empty value). The live-typing path (`controlTextDidChange`) is untouched, so typing and clearing via the cancel button still work, and tab scoping between the tables and favorites filters is preserved.

## Why this matters

The maintainer confirmed in #1690 that the filter pane is scoped per table tab and is expected to persist. Prefix filtering (for example `dp_`) should survive opening a table; clearing it on every open made the feature unusable for its main purpose.

## Testing

- `swiftlint lint --strict` passes on the touched files.
- Added `TableProTests/Views/SidebarSearchPersistenceTests.swift` covering the decision: keep the filter when the field still has text, clear it when the field is empty.
- The clear decision is extracted into a pure `nonisolated` helper (`shouldClearOnEndSearching`) so it is unit-testable without standing up AppKit.

Fixes #1690

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The sidebar filter no longer clears when you open a table from the filtered list. It only clears when you empty the field. (#1690)
 - DuckDB VARIANT columns now show their value as text instead of an empty cell.
 
 ## [0.51.1] - 2026-06-16
--- a/TablePro/Core/Services/Infrastructure/SidebarContainerViewController.swift
+++ b/TablePro/Core/Services/Infrastructure/SidebarContainerViewController.swift
@@ -131,7 +131,12 @@ extension SidebarContainerViewController: NSSearchFieldDelegate {
     }
 
     func searchFieldDidEndSearching(_ sender: NSSearchField) {
+        guard Self.shouldClearOnEndSearching(fieldValue: sender.stringValue) else { return }
         writeSearchText("")
+    }
+
+    nonisolated static func shouldClearOnEndSearching(fieldValue: String) -> Bool {
+        fieldValue.isEmpty
     }
 
     func control(_ control: NSControl, textView: NSTextView, doCommandBy commandSelector: Selector) -> Bool {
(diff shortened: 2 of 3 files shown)


mvanhorn:

datlechin:
Hi @mvanhorn, could you please describe more detail on the behavior both prior to and following this adjustment?

mvanhorn:
The failing `macOS App Tests` check isn't from this PR. The test target fails to **build** on `EtcdQueryBuilderTests.swift` — `#expect(parsed?.filterType == .none)` is ambiguous between `Optional.none` and `EtcdFilterType.none` under the current Xcode/Swift toolchain, so the whole test target won't compile and none of the tests actually run. This PR only touches `SidebarContainerViewController.swift` and `SidebarSearchPersistenceTests.swift` — nothing in the Etcd plugin.

Happy to open a separate PR disambiguating those comparisons (e.g. `== EtcdFilterType.none`) to get the test target building again, if that's useful. (Side note: the `cla` check shows red, but the CLA Assistant bot already posted that all contributors have signed.)

datlechin:
Hi @mvanhorn, thanks for your contribution! I've already merged a fix for keeping sidebar filters between tabs in PR #1715, so I'll close this PR.
Feel free to open another PR if you find any other issues or improvements. Thanks again!

PR #1714, opened 2026-06-18
fix(sidebar): refresh table tree after dropping a table

desperadoxhy:
## Problem

Dropping a table from the sidebar (right-click Delete, then Save Changes) did not remove it from the tree on multi-database servers like MySQL and PostgreSQL. The dropped table stayed visible until you refreshed the schema by hand.

## Root cause

`reloadSchema` only reloaded `SchemaService`. The tree for `.bySchema` and `.byDatabase` servers (`DatabaseTreeView`) reads from `DatabaseTreeMetadataService`, and `loadTables` short-circuits once a list is loaded. Nothing refreshed that store after a drop, so the stale table list stayed put.

## Fix

- Add `DatabaseTreeMetadataService.refreshLoadedTables(connectionId:)`: cancel dedup, clear, and refetch every expanded database/schema's tables for the connection.
- `MainContentCoordinator.reloadSchema` calls it right after `schemaService.reload`, so both data sources update.

Scope is tables only (no routines).

## Tests

Two new tests in `DatabaseTreeMetadataServiceTests`:
- refetch replaces a previously loaded list after it changes
- no-op when no tables are loaded

Both pass. App and test targets compile.

## Notes

`connectionObjectKeys` is now `nonisolated`. It is a pure function with no actor state, and the old MainActor isolation broke the existing tests under Xcode 27 beta strict concurrency. Stable Xcode accepts the annotation too.

Verified by hand on a MySQL connection: drop a table from the sidebar, Save Changes, and it disappears from the tree without a manual refresh.


desperadoxhy:

datlechin:
Pushed a maintainer update to this branch:
- Merged latest `main` so this runs against the new `NSOutlineView` sidebar tree (#1717) and the connection-restore fix (#1719). Verified the tree tests pass on the merge.
- Refactored `refreshLoadedTables` to update each loaded node in place with an equality guard, instead of clearing and reloading. Same result, but it avoids a transient loading flicker on every refresh and keeps the outline diff stable. Added a test for the multi-schema refresh and the `(#1714)` CHANGELOG reference.

The red `macOS App Tests` were unrelated SQL-parser suites (Derived Table Parser / SQL Context Analyzer), not this change; the re-sync should clear them. Still needs the CLA signed on your side before it can merge. Thanks for the fix!

datlechin (approved):
Thank you.

PR #1718, opened 2026-06-18
fix(tests): disambiguate EtcdFilterType.none in EtcdQueryBuilderTests

mvanhorn:
The Etcd query-builder test target fails to build under the current Swift toolchain: `parsed?.filterType == .none` resolves `.none` to `Optional<EtcdFilterType>.none` rather than `EtcdFilterType.none`. Under warnings-as-errors that breaks the whole `macOS App Tests` target, and semantically the comparison checked `filterType` against nil instead of the `.none` enum case (each call site already asserts `parsed != nil`).

This spells out `EtcdFilterType.none` at the four call sites so the assertions check the intended enum case and the target builds. No production code changed.

### DIFF
--- a/TablePro/Core/Autocomplete/DerivedTableParser.swift
+++ b/TablePro/Core/Autocomplete/DerivedTableParser.swift
@@ -263,8 +263,8 @@ internal struct DerivedTableParser {
         let c = ns.character(at: index)
         if c == Self.backtick || c == Self.doubleQuote {
             let end = matchingQuote(in: ns, from: index, limit: limit, quote: c)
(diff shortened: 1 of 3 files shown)


datlechin (approved):

mvanhorn:
Thanks for reviewing and merging, @datlechin. Disambiguating EtcdFilterType.none gets the query-builder test building under the current Swift toolchain.

mvanhorn:
Thank you @datlechin. Disambiguating EtcdFilterType.none so the query-builder test builds under the current Swift toolchain is a clean fix.

PR #1745, opened 2026-06-21
fix(editor): keep autocomplete delegate attached for the editor's lifetime (#1731)

### LINKED ISSUE #1731: Autocomplete completely non-functional in SQL editor
**Description**
Autocomplete does not work at all in the SQL editor. No suggestion dropdown ever appears — not for keywords, table names, or column names. The only editor feature working is auto-capitalisation of keywords (e.g. "select" becomes "SELECT").

**Steps to reproduce**
1. Connect to a PostgreSQL database with tables
2. Open a new query tab
3. Type anything — keywords, table names, column names

**Observed behaviour**
- No autocomplete dropdown appears at any point, for anything
- Auto-capitalisation works (select → SELECT) — so the editor is otherwise receiving input
- Schema is fully loaded (all tables visible in the sidebar)

**Expected**
An autocomplete dropdown suggesting keywords, table names, and column names as I type.

**Notes**
- No autocomplete toggle found in Settings > Editor (assumed always-on)
- Also: backspace deletes two characters instead of one (possibly a related editor bug). On reloading, this fixed the issue. 

**Environment**
- TablePro version: Version 0.52.0 (100)
- macOS: 26.5.1
- Database: PostgreSQL 18 (local, via DBngin)
- Mac: Apple M4

datlechin:
Fixes #1731.

## Problem
SQL autocomplete intermittently stopped appearing for any category (keywords, tables, columns), even with the schema fully loaded. Auto-capitalization still worked. It was tied to app/editor lifecycle and only a relaunch fixed it.

## Root cause
`TextViewController.completionDelegate` is `weak`, and the only strong owner was an optional `@State completionAdapter` that `SQLEditorView.onDisappear` set to `nil`. macOS does not guarantee `onAppear` fires after every `onDisappear` for views in native window-tab groups. When it didn't, the adapter was never recreated, `controller.completionDelegate` stayed `nil`, and every keystroke hit the trigger guard `guard let completionDelegate ... else { return }`, killing all completions (including schema-independent keywords) until relaunch.

## Fix
The completion adapter now lives for the editor's whole lifetime.

- `SQLEditorView`: `completionAdapter` is non-optional, created eagerly with nil schema; removed the `editorReady` two-phase gate and flattened the body; `onDisappear` no longer nils the adapter. `SourceEditor.updateNSViewController` continuously re-asserts a live, non-nil delegate. Keywords work from frame one; tables/columns enrich once schema loads.
- `SQLCompletionAdapter`: replaced `updateSchemaProvider` with `configure(schemaProvider:databaseType:)` accepting optionals, so the eager adapter can attach before schema loads and reconfigure when it arrives.
- `SQLEditorCoordinator.revive()`: reinstalls the editor-settings and window-key observers that `destroy()` tears down, so editor-settings and Vim live updates survive the destroy/revive cycle that tab switching now routinely exercises.

The "schema race" an earlier analysis pass suspected was not real: `services.schemaProviderRegistry` is `.shared` and the coordinator's `init` already calls `getOrCreate`, so the reconcile lookup always succeeds. No change made there.

## Tests
New `SQLCompletionAdapterLifecycleTests` (3 tests, all passing): the engine returns keyword completions with a nil schema (with and without a database type), and the adapter survives nil <-> provider `configure` transitions. The SwiftUI `onDisappear`/`onAppear` delegate detachment itself is an AppKit lifecycle behavior that is not deterministically unit-testable; these tests guard the invariants the fix relies on.

## Verification
- `BUILD SUCCEEDED`
- `swiftlint lint --strict`: 0 violations on changed files
- New tests pass (3/3)

Needs manual confirmation in-app: type to get suggestions, switch between query tabs and connection windows and back, and confirm autocomplete keeps appearing without a relaunch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- SQL autocomplete no longer stops appearing until the app is relaunched; the suggestion list stays available after switching query tabs and windows. (#1731)
 - Oracle connections no longer crash the app when the server sends a backend message the driver cannot decode; the query fails with a clear error and the connection reconnects. (#483)
 - MongoDB TLS handshake failures now report the actual cause instead of always blaming a cipher or protocol mismatch. (#1418)
 - The External Clients access level no longer reverts to Read Only after saving and reopening a connection, so MCP clients keep the write access you granted. (#1730)
--- a/TablePro/Views/Editor/SQLCompletionAdapter.swift
+++ b/TablePro/Views/Editor/SQLCompletionAdapter.swift
@@ -44,9 +44,9 @@ final class SQLCompletionAdapter: CodeSuggestionDelegate {
         self.completionEngine = Self.makeEngine(schemaProvider: schemaProvider, databaseType: databaseType)
     }
 
-    /// Update the schema provider (e.g. when connection changes)
-    func updateSchemaProvider(_ provider: SQLSchemaProvider, databaseType: DatabaseType? = nil) {
-        completionEngine = Self.makeEngine(schemaProvider: provider, databaseType: databaseType)
+    /// Rebuild the completion engine for the current connection (nil schema still yields keyword completion)
+    func configure(schemaProvider: SQLSchemaProvider?, databaseType: DatabaseType?) {
+        completionEngine = Self.makeEngine(schemaProvider: schemaProvider, databaseType: databaseType)
         completionEngine.updateFavoriteKeywords(favoriteKeywords)
     }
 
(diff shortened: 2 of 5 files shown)


karapy:
I am still experiencing this issue on **Version 0.55.0 (104)** running on **macOS**. 

Despite the fix mentioned in #1745, the autocomplete/IntelliSense feature for table columns (e.g., typing `orders.` after the table name) remains completely non-functional for me. This issue heavily impacts the workflow, to the point that I've had to stop using the app for now. 

Could you please clarify if this fix has been included in version 0.55.0 (104), or if it will be rolled out in a future release? Let me know if you need any logs or further environment details to debug this.

PR #1763, opened 2026-06-23
feat(plugin-oracle): show login handshake phase in connection-dropped diagnostic (#1746)

datlechin:
## Why

The #1746 crash is fixed, but the reporter (Oracle 11.2.0.4, no encryption) still can't connect: TablePro shows "the server closed the connection during the login handshake" (`.authConnectionDropped`, from oracle-nio's `uncleanShutdown`). There's no information about *where* in the handshake the server dropped us, so the report is opaque.

The fork already supports 11g at the version/auth layer (`TNS_VERSION_MIN_ACCEPTED = 313`, 10G/11G/12C verifiers), so this is not the upstream "12.1 floor" case. The drop is somewhere in negotiation/auth, and oracle-nio already records the phase in `OracleSQLError.handshakePhase` (`connect` / `advancedNegotiation` / `protocolNegotiation` / `dataTypeNegotiation` / `preAuthentication` / `authentication`). The plugin was discarding it.

Note: the pcap attached to the issue is an SQLcl session, not TablePro, so it doesn't show the failure. This change captures the missing signal directly from the app.

## What

Instrument-only, no fork change:

- Log the handshake phase and oracle-nio code on connect failure.
- Surface the phase in the "Connection Dropped During Handshake" dialog as a `DiagnosticEntry` (same pattern as the existing "Verifier flag" entry), plus a suggested action to file an issue with the phase.
- Extract the connect-error string classification into a pure `OracleConnectErrorClassifier` in TableProPluginKit (mirrors `OracleChannelFatalCode`) so it's unit-testable instead of dead-gated.

Once the reporter comes back with the phase: `dataTypeNegotiation` points at the uncleared Oracle 23ai capability bits in `adjustForServerCompileCapabilities`; `authentication` points at a verifier issue; `connect` at TLS/listener.

## Notes

- Oracle is a registry-only plugin, so this ships via a `plugin-oracle-v*` release.
- `OracleConnectErrorClassifier` is a new public PluginKit type (additive ABI, no version bump).

## Tests

- `OracleConnectErrorClassifierTests`: `uncleanShutdown` → dropped, `serverVersionNotSupported` → version, `unsupportedVerifierType(...)` → verifier (flag preserved), other → generic.
- `swiftlint lint --strict` clean on changed files.

https://claude.ai/code/session_01NVomHPErfCr8FWRVywmyVP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The ER diagram now arranges tables in a compact layout that fills the canvas in both directions, keeps tables linked by foreign keys together, and tints each group of connected tables with its own header color. (#1755)
+- When an Oracle server closes the connection during login, the error dialog now shows which handshake phase it stopped at, so dropped-handshake failures (for example on Oracle 11g) are easier to pin down. (#1746)
 
 ### Fixed
 
--- a/Plugins/OracleDriverPlugin/OracleConnection.swift
+++ b/Plugins/OracleDriverPlugin/OracleConnection.swift
@@ -27,7 +27,7 @@ struct OracleError: Error {
         case protocolError
         case authVerifierUnsupported(flag: String)
         case authVersionNotSupported
-        case authConnectionDropped
+        case authConnectionDropped(phase: String?)
     }
 
     let message: String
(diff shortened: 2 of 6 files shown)


Bin43:
我的锅，现在附上TablePro 会话的pcap 文件
[oracle.pcap.zip](https://github.com/user-attachments/files/29320269/oracle.pcap.zip)

PR #1776, opened 2026-06-27
feat(editor): per-tab database picker and tab interaction improvements

desperadoxhy:
## Summary

Adds a per-tab database picker to the query editor and reworks tab interactions so switching databases no longer clears tabs. Also applies the sidebar tree filter to every database picker and smooths out tab switching.

## Changes

- **Per-tab database picker** in the query editor toolbar. Each SQL tab binds to its own database; the active database switches when the picker changes and again before a query runs. Read-only for reconnect-to-switch databases (PostgreSQL, Redshift, CockroachDB).
- **Single-click a table** in the sidebar tree opens it in the current tab; **double-click** opens a new tab.
- **Switching databases keeps existing tabs open** instead of closing them, so tabs for different databases coexist. Dropped the dead `clearTabs` branch from `switchDatabase` in the process.
- **Database pickers** (toolbar switcher, quick switcher, query editor) **follow the sidebar tree filter** and only list enabled databases.
- **Tab switching perf**: filter settings now persist off the main thread, and switching to a table tab in a different database refreshes only that database's table list instead of every expanded one.

## Notes

- `scripts/run-local.sh` added to patch and launch unsigned CLI Debug builds locally (copy OpenSSL dylibs into Frameworks, ad-hoc sign, open).
- CHANGELOG updated under [Unreleased].

## Test plan

- [ ] Open a multi-database MySQL connection. Use the per-tab picker to bind tabs to different databases, run SQL, confirm each runs in its bound database.
- [ ] Single-click a table opens it in the current tab; double-click opens a new tab.
- [ ] Switch databases from the sidebar tree, toolbar chip, and quick switcher; confirm tabs are preserved.
- [ ] Configure the sidebar database filter; confirm toolbar switcher, quick switcher, and editor picker only show enabled databases.
- [ ] Open table tabs in different databases and switch between them rapidly; confirm it stays smooth.


datlechin (approved):
Thank you.

PR #1778, opened 2026-06-28
fix(plugin-mssql): use active schema for object queries

### LINKED ISSUE #1774: Invalid Object name on views and table in other then DBO schema
https://github.com/TableProApp/TablePro/issues/1754#issue-4718786370

 Installed new release, problem still exists. 
 Completely uninstalled Table pro and removed the files in ~application support, reinstalled TablePro and MSSQL driver.
 
 Problem still exists.  

 _Originally posted by @martijnmcb in [#1754](https://github.com/TableProApp/TablePro/issues/1754#issuecomment-4808128793)_

andiahmads:
## Summary
- Resolve MSSQL object queries against the active schema when callers do not pass an explicit schema.
- Keep explicit schema arguments taking precedence over the active schema.
- Add TableProMSSQLCore tests for schema resolution and non-dbo browse query generation.

## Root cause
MSSQL browse, filter, and generated edit statements treated a nil schema as unqualified object access. For connections whose active schema was not dbo, this produced queries like SELECT * FROM [Orders], which SQL Server resolved through the login default schema instead of the selected schema.

Fixes #1774

## Validation
- swift test --package-path Packages/TableProCore --filter MSSQLSchemaQueriesTests
- xcodebuild -project TablePro.xcodeproj -scheme TablePro -configuration Debug -skipPackagePluginValidation CODE_SIGNING_ALLOWED=NO build
- git diff --check
- swiftlint lint --strict


datlechin:
Thanks for digging into this and for the detailed writeup. Closing it for two reasons.

First, #1774 was fixed on main in bab52742 and shipped in 0.55.0, at the app layer: the coordinator resolves the schema from the active session when a tab is created and passes it explicitly into browse, filter, and edit statement generation. A driver-side fallback to `_currentSchema` cannot cover the reported path, because the app builds browse SQL on a cached driver instance that never connects. On that instance `freeTDSConn` is nil and `_currentSchema` stays at its `dbo` init default, so the `allowCurrentSchemaFallback: freeTDSConn != nil` gate switches the fallback off exactly where the bug happens, and removing the gate would qualify with `dbo` rather than the active schema.

Second, where the fallback would fire (a connected driver receiving a nil schema), qualifying with `_currentSchema` can change which table an UPDATE or DELETE targets after the user switches schemas, which is riskier than leaving the statement unqualified.

The iOS CI change bundled here is unrelated to the fix but useful. I split it out into #1812, rebuilt on `simctl` with the device name derived from `TEST_DESTINATION`, since the `-showdestinations` grep also matched "iPhone 17 Pro Max" and ineligible destinations.

If you can still reproduce "Invalid object name" on 0.55.0 or later, please add the steps to #1774.

andiahmads:
Thanks for the detailed explanation. That makes sense, especially the risk around qualifying UPDATE/DELETE from a connected driver after schema switches.

I will treat this PR as closed, verify the flow again on 0.55.0/main, and add concrete repro steps to #1774 if I can still reproduce the issue.

PR #1790, opened 2026-06-30
feat(datagrid): show column type and comment as persistent header subtitle

bulang-J:
…title


datlechin:
Hello @bulang-J, could you take screenshots the header columns after these changes?

bulang-J:
> Hello @bulang-J, could you take screenshots the header columns after these changes?

Sorry, I can't provide it.

PR #1792, opened 2026-06-30
feat(datagrid): show column comment as a persistent header subtitle (#1789)

### LINKED ISSUE #1789: 数据网格表头注释目前为悬浮显示，直观性差，建议优化为常驻展示形式。
### Problem

数据网格表头注释目前为悬浮显示，直观性差，建议优化为常驻展示形式。

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Related database type

None

datlechin:
Closes #1789. Supersedes #1790.

## Problem

When **View > Show Object Comments** is on, a column's comment was only added to the header's hover tooltip. It was never visible without hovering. #1789 asks for the comment to show persistently in the column header.

## Change

The comment now renders as a second line under the column name when the setting is on. The tooltip stays as the full-text fallback for truncated comments. With the setting off, or for columns without a comment, the header is the single line it is today.

- `SortableHeaderCell` splits the cell into a name band (top, sized to the single-line height) and a comment band (bottom). Title, sort chevron, priority badge, and filter funnel all align to the name band, so indicators never overlap the subtitle. Comment uses `secondaryLabelColor` and a smaller font, truncates with an ellipsis, and is appended to the VoiceOver label.
- `SortableHeaderView` captures AppKit's natural single-line header height and enforces the taller height in `setFrameSize`, so it survives AppKit re-tiling on column and window resize.
- `DataGridColumnPool` decides the two-line state from the visible columns that have a comment and drives both the cells and the header view.
- `DataGridView` keeps the row-number header aligned and adds a comments signature to the update snapshot so comments that load after the columns first appear still refresh.

Scoped to comments only, which is what #1789 asks. #1790 also showed the column type under the same toggle; that is left out here.

## Tests

- `SortableHeaderCellTests`: name-band/comment-band split, indicator alignment to the name band, single-line fallback.
- `DataGridColumnPoolTests`: cells get the comment and two-line flag, single line when no comment, hidden columns excluded.

## Notes

- No new user-facing strings (reuses the existing toggle).
- CHANGELOG and the MySQL/PostgreSQL docs updated.

https://claude.ai/code/session_0198faM6VCrViRU4XwRoS1DC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The data grid column header now shows a column's comment on a second line under the name, not just in the hover tooltip. Turn it on with View > Show Object Comments. (#1789)
+
 ## [0.54.0] - 2026-06-30
 
 ### Added
--- a/TablePro/Views/Results/DataGridColumnPool.swift
+++ b/TablePro/Views/Results/DataGridColumnPool.swift
@@ -11,6 +11,7 @@ final class DataGridColumnPool {
     private weak var attachedTableView: NSTableView?
 
     var totalSlots: Int { pooledColumns.count }
+    private(set) var requiresCommentLine = false
 
     func attach(to tableView: NSTableView) {
         attachedTableView = tableView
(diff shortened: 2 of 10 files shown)


datlechin:
Superseded: the same implementation is now pushed directly to #1790.

Issue #956, opened 2026-04-30
sqlite3 does not apply limit?


fuzzy76:
### What happened?

```
SELECT uuid FROM TMTask WHERE project='BwcQ91hS4C9fQaUZGMWDmT' AND status IN (2,3) ORDER BY stopDate DESC LIMIT 10;
```

From the query editor returns all rows it can find (806 in my case).

### Expected behavior

10 rows, just like I get when I run the query from my terminal:
```
fuzzy76@Havard-sin-Mac-mini:~/repos/things3export % sqlite3 
SQLite version 3.51.0 2025-06-12 13:14:41
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> .open main.sqlite
sqlite> SELECT uuid FROM TMTask WHERE project='BwcQ91hS4C9fQaUZGMWDmT' AND status IN (2,3) ORDER BY stopDate DESC LIMIT 10;
SEF7iW5HMFZmoGoPWLWzVX
KUQxLX5wbTpPrHK368hbJA
VaBWgVbpibp2UNFgXXqkUx
HENVc28D9TLP7nuRFEWsgV
T92seQtp7pZrNprUfFTQUS
3GZ2ooceCBACWrPJpSDoFj
RkMfNi7cfXRhr8Mk1iQRpa
RBqjmfZBcbSYFUHYxm2CP5
VP6LeW7E8RwiBp4iwFQf7J
VVi2pjniPy4fKvFpk5HJro
sqlite>
```

### Database type

SQLite

### TablePro version

0.36.0

### macOS version & chip

Tahoe 26.4.1 on M2

datlechin:
Thank you, checking right now

PR #1804, opened 2026-07-02
feat(editor): apply the row cap as a LIMIT on executed queries

### LINKED ISSUE #1794: 自动添加 limit
### Problem

操作 mysql 等数据库时, 有时会遗漏添加 LIMIT 导致拉取过多数据

### Proposed solution

建议查询时, 真正执行查询时, 自动添加 LIMIT, 但是查询窗口中的语句不变.

### Alternatives considered

_No response_

### Related database type

None

datlechin:
Closes #1794.

## Problem

The row cap trimmed results only after the full result set crossed the wire. PostgreSQL materializes the whole result in libpq before the app sees a row, and MySQL streams but drains everything up to the 5,000,000-row emergency ceiling. A forgotten LIMIT still pulled millions of rows. On top of that, Execute All Statements bypassed the cap entirely, and a query starting with a comment slipped past both the cap and Safe Mode classification.

## What changed

- SELECT and WITH statements with no row constraint of their own now run with `LIMIT (cap+1)` appended to the SQL sent to the server. The editor text never changes. The `cap+1` over-fetch keeps the existing post-fetch cap, the "truncated / Fetch All" status bar link, and Fetch All (which stores the pre-injection SQL) working unchanged.
- New `SQLLimitInjector` scans with paren depth, string, comment, and dollar-quote awareness. A user-written top-level `LIMIT`, `OFFSET`, or `FETCH` always wins. An inner LIMIT in a CTE or subquery does not suppress injection. Insertion lands before trailing comments and semicolons. Statements ending in `FOR UPDATE`, `INTO`, `LOCK`, ClickHouse `FORMAT`/`SETTINGS` bail out to the post-fetch cap.
- Dialect routing wires up the previously unused `AutoLimitStyle` metadata: MySQL, PostgreSQL, SQLite, ClickHouse, DuckDB, Cassandra, Cloudflare D1, and libSQL get injection; Redis, MongoDB, etcd resolve `.none`; SQL Server and Oracle keep post-fetch capping only. A new additive PluginKit hook `injectRowLimit(_:limit:)` (default nil) lets those plugins own `TOP`/`FETCH FIRST` insertion later without an ABI change.
- Execute All Statements applies the same per-SELECT cap and gains the Fetch All affordance.
- Comment-prefixed queries are now classified correctly for the row cap and for Safe Mode write/destructive checks.
- Query history records the statement that actually ran.
- MySQL threads the user cap into its streaming fetch loop break, mirroring the SQLite override.
- UI: Execute is a split button (SwiftUI `Menu` with `primaryAction`) with Execute Without Limit; the same command is in the Query menu with Cmd+Option+Return (customizable). Settings footer copy and docs updated.
- Removed the dead `MainContentCoordinator+MultiStatement.swift` proxy (zero callers) and deduplicated the two multi-statement loops into shared helpers.

## Design note

This reverses the v0.37.0 removal of SQL rewriting (#956). That removal fixed a naive detector that clobbered user-written LIMITs. The new detector only ever appends and never touches a statement that constrains itself; the `ExecuteUserQueryTests` boundary contract (SQL at the driver boundary runs verbatim) still holds because injection happens before that boundary.

## ABI

`Plugins/TableProPluginKit/PluginDatabaseDriver.swift` gains one protocol requirement with a default implementation. `scripts/check-pluginkit-abi.sh main` reports exactly the two added `injectRowLimit` lines, nothing removed or changed, so no version bump. Label `abi-additive` applied.

## Tests

- `SQLLimitInjectorTests` (21 cases): existing-LIMIT detection, CTE and subquery inner LIMITs, trailing comments, UNION forms, dollar quotes, escaped quotes, identifier false positives, unsupported trailing clauses, malformed input.
- `QueryExecutorTests`: row-cap qualification including comment-prefixed SELECTs; `QueryClassifierTests`: comment-prefixed write/destructive classification.
- `ExecuteUserQueryTests`: new case documenting the byte-for-byte boundary contract for post-injection SQL; existing cases unchanged.
- `DriverPluginMetadataTests`: registry defaults declare the expected auto-limit style per dialect.
- Pagination and status bar suites green. `swiftlint --strict` clean on changed app files.

No UI automation added: the without-limit flow needs a live database connection, which TableProUITests cannot drive deterministically.

Known flake (pre-existing, not touched by this change): `QueryExecutorTests/parseSchemaMetadataExtractsEnumValues` fails in parallel batch runs and passes alone.


datlechin:
Review pass on this branch surfaced 8 confirmed findings. All are fixed in 123e6fe4, plus the lower-priority items the review noted.

**Correctness fixes**

1. Cassandra/ScyllaDB `ALLOW FILTERING`: `ALLOW` and `FILTERING` are now blocking keywords, so those queries are sent as written with the post-fetch cap instead of getting an injected LIMIT that CQL rejects.
2. Fetch All after a multi-SELECT batch no longer replaces the visible result set with rows from a different query. Truncation state (`isTruncated`, `baseQuery`, `baseQueryParameterValues`) now lives on each `ResultSet`; switching result tabs syncs the tab-level pagination from the active result set, so Fetch All always re-runs the query behind the grid you are looking at.
3. A capped SELECT earlier in a batch is no longer silent: its result tab keeps its own truncated state and shows Fetch All when selected.
4. `bypassRowLimit` reaches the parameterized multi-statement branch. The two batch loops are also merged into one (the duplication caused the drop), so the flag cannot diverge again.
5. Sorting after a parameterized run falls back to the editor text when bound parameter values are present, so the `$1`/`?` form is never re-run unbound.
6. History records your query text again: the original `:name` form for parameterized statements (single and batch) and the pre-injection SQL for capped SELECTs. No more `LIMIT cap+1` or `$1` placeholders in history or favorites. Failed queries still record the statement that actually ran, for debugging.
7. Statement classification is token-based now: `SELECT\n`, `SELECT*`, tab after the keyword, and the same shapes for writes and destructive statements are all classified correctly, for the row cap and for Safe Mode.
8. "Your own LIMIT always wins" is now true: a statement with its own top-level `LIMIT`/`FETCH FIRST` (or `TOP` on SQL Server) runs with no cap at all instead of being post-fetch truncated. `SQLLimitInjector` returns `injected` / `alreadyLimited` / `notInjectable` and the plan drops the cap for `alreadyLimited`.

**Also picked up**

- MySQL `#` line comments are handled by the injector (dialect-gated via `SqlDialect.supportsHashLineComments`), so the LIMIT lands before a trailing `#` comment instead of inside it.
- The MySQL parameterized fetch loop now honors the user row cap instead of reading up to the 5M emergency ceiling and trimming after.
- The injector reuses `SqlDollarQuote`'s identifier classification and matches keywords without per-word string allocation; insertion is a single in-place edit instead of three full copies.
- Dead proxies removed (`resolveRowCap`, the parameterized forwarders on `MainContentCoordinator`).
- Fixed a pre-existing always-failing test on main (`parseSchemaMetadataExtractsEnumValues`): `parseSchemaMetadata` now falls back to `EnumValueParser.parseMySQLEnumOrSet` when a driver reports enum values only in the column type.

Docs and CHANGELOG updated to match the corrected semantics. ABI check against main reports only the three additive symbols (`injectRowLimit` + default, `supportsHashLineComments`); no version bump. 86 tests green across the affected suites.

PR #1838, opened 2026-07-09
feat(sidebar): add New Query to Favorites plus menu

J2TeamNNL:
## Summary
- Add **New Query** as the first item in the Favorites sidebar **+** menu; it opens an empty SQL query tab via the existing `newTab()` path (same as `Cmd+T`).
- Update docs (`favorites.mdx`, `tabs.mdx`) and CHANGELOG.
- Add localization strings for the new menu label and tooltip.

Autocomplete (table/column suggestions), execute shortcuts (`Cmd+Enter`), and optional auto-uppercase keywords are unchanged; this PR only adds a clearer entry point for creating a new query from the Favorites sidebar.

## Test plan
- [ ] Connect to a database, switch to the Favorites sidebar tab.
- [ ] Click **+** > **New Query** and confirm an empty query tab opens with editor focus.
- [ ] Confirm **New Favorite**, **New Folder**, and linked-folder actions still work.
- [ ] Confirm `Cmd+T` still opens a new query tab.
- [ ] Confirm `Cmd+Enter` still executes a query.
- [ ] Confirm Settings > Editor still shows **Auto-uppercase keywords** (default off).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The Favorites sidebar **+** menu now includes **New Query**, which opens an empty SQL query tab.
+
 ## [0.56.0] - 2026-07-09
 
 ### Added
--- a/TablePro/Resources/Localizable.xcstrings
+++ b/TablePro/Resources/Localizable.xcstrings
@@ -49282,6 +49282,62 @@
         }
       }
     },
+    "New Query" : {
+      "localizations" : {
+        "tr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Yeni Sorgu"
+          }
+        },
+        "vi" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Truy vấn mới"
+          }
+        },
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "新建查询"
+          }
+        },
+        "zh-Hant" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "新增查詢"
+          }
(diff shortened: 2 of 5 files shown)


datlechin:
Addressed the review findings and pushed to this branch (`maintainerCanModify`):

**Rebased onto current main / CHANGELOG conflict resolved.** The branch was based before 0.56.1/0.56.2. Merged `main` and put the New Query entry back under `[Unreleased]`, above the released `[0.56.2]` section. PR is mergeable again; the effective diff is unchanged (5 files: CHANGELOG, string catalog, `FavoritesTabView`, `favorites.mdx`, `tabs.mdx`).

**`tabs.mdx` SQL-button row removal is correct.** I checked: there is no toolbar button that opens a new query tab (the only `"SQL"` references are `queryLanguageName: "SQL"` plugin metadata). That doc row was stale, so dropping it is a fix, not a regression.

**Minor points from the review, resolved:** `FavoritesTabView` lints clean under `swiftlint --strict`; the old `"Add"` help string is not orphaned (still used in `CustomSlashCommandsSection` and `FavoriteEditDialog`).

**On test coverage:** no automated test added, and here's why. The action is a single button wired to `commandActions.newTab()`, which is the same path as `Cmd+T` and is already exercised. The only end-to-end surface is the Favorites sidebar `+` menu, which is only reachable with an active connection window, and `TableProUITests` currently has just launch smoke tests with no connection-seeding harness. A UI test would mean building that harness (seed a SQLite connection at launch), which is out of scope for this change and would be flaky. If a connection-seeding UI harness lands later, this flow is a good candidate to cover then.

One design note for the maintainer, not blocking: putting **New Query** in the `+` (add to favorites) menu is slightly off the "add an item to this list" idiom, since it opens a tab rather than adding to the sidebar. The divider and updated tooltip mitigate it, and it's a reasonable convenience, but worth a conscious call.

PR #1844, opened 2026-07-09
feat(connections): native Cloud SQL Auth Proxy integration

### LINKED ISSUE #1728: Native Cloud SQL Auth Proxy Integration
### Problem

Google Cloud SQL users commonly connect through the Cloud SQL Auth Proxy, which provides IAM-based authentication and secure TLS connections without requiring direct database exposure or certificate management.

Today, using TablePro with Cloud SQL requires users to:

1. Install the Cloud SQL Auth Proxy manually.
2. Start the proxy from a terminal.
3. Manage proxy lifecycle separately from TablePro.
4. Configure database connections to localhost ports exposed by the proxy.

This creates additional setup complexity compared to other managed database services.

### Proposed solution

Add built-in Cloud SQL Auth Proxy support to TablePro.

Possible implementation options:

Option 1: Managed Proxy Process (Preferred)

Allow users to configure:

* GCP Project
* Instance Connection Name (project:region:instance)
* Authentication method:
    * Application Default Credentials
    * Service Account JSON
    * gcloud authenticated user

TablePro would automatically:

* Launch the Cloud SQL Auth Proxy in the background
* Manage proxy lifecycle
* Connect through the local proxy endpoint
* Restart the proxy when needed

Option 2: Proxy-Aware Connection Type

Add a dedicated "Google Cloud SQL" connection type that simplifies setup while still using the Cloud SQL Auth Proxy under the hood.

Benefits

* Much easier onboarding for Cloud SQL users
* Supports Google’s recommended connection method
* Enables IAM-based authentication workflows
* Reduces manual terminal usage
* Better developer experience for teams using GCP

Supported Databases

This feature would benefit all Cloud SQL engines currently supported by TablePro:

* PostgreSQL
* MySQL
* SQL Server

Additional Considerations

* Auto-download Cloud SQL Auth Proxy binaries per platform
* Proxy status indicators in the connection UI
* Automatic port allocation
* Support for multiple Cloud SQL instances simultaneously
* Secure credential storage using the macOS Keychain

References

* Google Cloud SQL Auth Proxy
    * PostgreSQL https://docs.cloud.google.com/sql/docs/postgres/sql-proxy
    * MySQL https://docs.cloud.google.com/sql/docs/mysql/sql-proxy
    * SQL Server https://docs.cloud.google.com/sql/docs/sqlserver/sql-proxy
* Cloud SQL Auth Proxy https://github.com/GoogleCloudPlatform/cloud-sql-proxy

### Alternatives considered

Using `cloud-sql-proxy` was considered as a workaround because it can provide secure connectivity to Cloud SQL instances without requiring the proposed changes. However, this approach introduces additional manual setup and does not address the underlying issue, such as port conflicts when running multiple connections. As a result, it is considered a temporary workaround rather than the preferred long-term solution.

### Related database type

PostgreSQL

datlechin:
Closes #1728.

Adds native Google Cloud SQL Auth Proxy support so Cloud SQL users no longer run `cloud-sql-proxy` in a terminal and hand-point a connection at its port. TablePro starts the proxy when you connect and stops it when you disconnect, the same way it manages SSH and Cloudflare tunnels.

## Design

The proxy is a per-connection transport attribute (`cloudSQLProxyMode` on `DatabaseConnection`), not a `DatabaseType`: a Cloud SQL instance is still MySQL, PostgreSQL, or SQL Server, and only the network path differs. It mirrors the existing Cloudflare tunnel end to end. `CloudSQLProxyManager` (an actor) supervises the `cloud-sql-proxy` process, allocates a free loopback port, waits for readiness with a TCP probe plus an stderr scan, then reuses the shared `tunneledConnection(from:localPort:)` rewrite unchanged. On disconnect, quit, or process death it is torn down, with the same PID-persist and stale-PID sweep the Cloudflare path uses.

Binary provisioning: resolves `cloud-sql-proxy` from PATH, Homebrew, or the gcloud SDK first; otherwise a "Download" button in the pane fetches Google's binary, verifies a hardcoded SHA-256 pin (v2.23.0), and strips quarantine. Google's macOS binaries are not Apple-notarized, so a locally installed one is preferred.

Authentication: Application Default Credentials, a service account key (stored in the Keychain, written to a `0600` temp file only while the proxy runs and never passed on the command line), and IAM database authentication (`--auto-iam-authn`).

## Shared tunnel dispatch (and two bug fixes)

Eight `DatabaseManager` sites used to hand-enumerate "SSH or Cloudflare?" by name. They now share one `TunnelManaging` protocol (both tunnel actors already satisfy it) plus `DatabaseConnection.activeTunnelKind` / `enabledTunnelKinds` and `activeTunnelManager(for:)`. Two of those sites only ever closed the SSH tunnel, so a failed test connection or a cancelled connect leaked the `cloudflared` process; the shared dispatch fixes both. Mutual exclusivity is now an N-way check.

Shared subprocess plumbing was extracted so it isn't written twice: `CloudflaredProcess` became the generic `SupervisedProcessRunner`, and the loopback port helpers became `LoopbackPort`.

## Tests

- Model Codable round-trips and instance-connection-name validation.
- `ConnectionTunnelKind` across every SSH/Cloudflare/proxy combination.
- `CloudSQLProxyManager` against a fake runner: readiness, startup failure, invalid instance, missing binary, service-account-key file lifecycle (written `0600`, removed on close), terminate, and stale-PID sweep.
- `CloudSQLProxyBinaryManager`: checksum match installs, mismatch is rejected, and an installed binary short-circuits without fetching.
- Pane validation.

## Docs and changelog

New `docs/databases/cloud-sql-proxy.mdx`, plus CHANGELOG `Added` (feature) and `Fixed` (the Cloudflare process leak) entries.

## Notes

- `swiftlint lint --strict` is clean on the changed files.
- I have not built this locally; please confirm it compiles before merge.


mdreizin:
@datlechin Thanks for adding that so quickly! You’re rocking it!

Issue #1913, opened 2026-07-19
Delete column or row in CSV


salmonumbrella:
### Problem

When opening a CSV, sometimes I want to delete a column or row. 

### Proposed solution

Right click and delete column, or row. 

### Alternatives considered

Use [Table Cruncher](https://github.com/Tablecruncher/tablecruncher) as an example.

### Related database type

N/A / General

Issue #1469, opened 2026-05-28
feat(csv): structural editing primitives (insert/delete column & row, split/merge, header toggle, CSV properties)


salmonumbrella:
### Problem

TablePro now opens (#1259), navigates and copies (#1330), selects column/cell ranges (#1446), and is gaining client-side filtering (#1454) on CSV files. What's still missing is the ability to mutate the file structure — delete an empty column, insert one to the right, drop a stray row, flip a row from header to data, or fix an auto-detected delimiter that came out wrong. Today these tasks force a bounce to Tablecruncher or Modern CSV, exactly the second-app split #1259 set out to collapse.

The column-header right-click menu already has `Copy Column Values`, `Copy as CSV/Markdown/IN Clause`, `Fill column…` (#1398), and `Hide Column`. It's the natural home for the missing structural verbs.

### Proposed solution

**1. Column-header right-click — add structural items**

- **Insert Column to the Left**
- **Insert Column to the Right**
- **Delete Column**
- *divider*
- **Split Column…** — by delimiter or regex, into N new columns
- **Merge with Next Column…** — joiner string

If multiple columns are selected via #1446, the verbs operate on the selection.

**2. Row-gutter right-click**

- **Insert Row Above**
- **Insert Row Below**
- **Delete Row(s)** — operates on the multi-row selection

**3. Edit menu mirror**

All of the above also appear under `Edit` so they're keyboard-discoverable and shortcut-bindable, mirroring Tablecruncher's [`csvmenu.cpp`](https://github.com/Tablecruncher/tablecruncher/blob/main/src/csvmenu.cpp).

**4. CSV-mode settings**

- **Switch First Row Between Header / Data** — toggle, `⌘⇧H`. Without this, "Delete Column" semantics ambiguate when the header row was wrongly auto-detected.
- **Set CSV Properties…** dialog — delimiter, quote char, escape char, encoding, line ending. Auto-detected at open, manually overridable + `Reload`.

**5. UX rules**

- Pure in-memory mutations; file marked dirty until **Save**.
- One undo entry per invocation.
- `Delete Column` / `Delete Row(s)` show a confirm only if the target contains any non-empty cell.
- `Save` writes through with the currently-configured CSV properties — no silent re-encoding.

### Implementation notes

- CSV-mode only. These verbs mutate the in-memory CSV datasource directly; they do not emit `ALTER TABLE` — that's the DB-mode Structure tab's job.
[...]

Issue #1259, opened 2026-05-13
feat: CSV inspection


salmonumbrella:
Hey — really impressed with what you've shipped here in five months, especially the MCP server.

One thing that would make TablePro a genuinely complete data tool for me (and I suspect a lot of others): native CSV/TSV inspection, without having to import into a SQLite scratch DB first. Right now the ecosystem split is:

- **TablePlus** → DB GUI, no real CSV editing
- **[TableTool](https://tabletool.io/)** → separate app, made by the TablePlus team specifically because TablePlus doesn't do CSVs
- **Modern CSV / Tablecruncher / etc.** → another separate app

You could collapse that split inside TablePro and have the best single-app story on macOS. The "open a .csv and start poking" workflow is enormously common — log dumps, payroll exports, Stripe/Shopify reports, API CSV responses — and right now even paid TablePlus users are bouncing out to a second app.

For prior art / a possible base to fork from: **[Tablecruncher](https://github.com/Tablecruncher/tablecruncher)** went GPL-3.0 in April 2025. It's a mature solo-dev CSV editor (originally a paid macOS app from 2017, then fully open-sourced), C++/FLTK, handles 16M-row / 2 GB files. License is GPL-3 so you'd need to either keep that subsystem AGPL-compatible or rewrite the hot paths in Swift, but the architecture and the years of edge-case handling (encodings, delimiters, quoting weirdness) are all there to learn from.

DuckDB integration you already have for the plugin would also do a lot of the heavy lifting on the query side — `duckdb.read_csv_auto()` over a memory-mapped file gets you pretty far before you need a custom renderer.

Anyway — thanks for building this in the open, happy to test any prototype.

PR #1939, opened 2026-07-21
feat(inspector): insert rows and confirm data-bearing deletes in the CSV editor

datlechin:
Part 1 of 4 toward #1469 (CSV structural editing). Focused on **rows**.

## What
- **Insert Row Above / Insert Row Below** from a row's right-click menu, and mirrored in the Edit menu (where they act on the selected row). Both insert a blank row and scroll to it.
- **Delete confirmation**: deleting rows that contain any non-empty cell now asks first. Deleting blank rows still deletes without a prompt.
- Row insert and delete are each a single named undo step ("Undo Insert Row", "Undo Delete Rows").

## Why
The CSV inspector (from #1259, extended by #1913/#1932) can edit cells and add/remove columns, but the only way to add a row was the toolbar Add Row button, which always appends. There was no way to insert at a position, and deletes had no guard for data-bearing rows. This closes the row half of #1469.

## How (native / HIG)
- Row items ride the existing shared row menu through a new `dataGridRowStructureMenuItems(forRow:)` delegate seam, mirroring the column seam #1932 added. Only the CSV inspector implements it, so database grids are unaffected.
- Edit-menu items dispatch through the responder chain (`NSApp.sendAction(_:to:nil)`) to the focused inspector, gated on `keyWindowIsInspector`, matching the existing Undo/Redo/Find routing.
- Insert reuses `CSVDocument.insertRow(at:)` (already undoable); delete adds a confirm-when-non-empty gate. Per Apple's HIG, undoable deletes normally don't need a prompt, so the prompt is scoped to data-bearing rows and its copy does not claim the action is irreversible.
- Selection is mapped display-position → store index through the inspector's `displayToStore`, respecting the filtered/sorted-view invariant (a display row is not a store row).

## Tests
- `CSVRowStore.insertRow` at a middle index, at the top, clamped past the end, and padding a short row.
- `InspectorRowMenuBuilder` item order, action wiring, and row tagging.
- `InspectorDeleteConfirmation.rowsContainData` / `rowDeleteTitle` decision logic.

## Docs / changelog
- `docs/features/csv-inspector.mdx` Editing section updated.

Needs a local Xcode build to confirm compilation (CI lints `TablePro/` only, which passes `--strict`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Connect to PGlite through its socket server (`pglite-server`), with schema browsing, SQL editing, and row editing like PostgreSQL. Defaults to a loopback host with no TLS, and treats the connection as single-use to match PGlite's one-connection limit. (#1911)
 - In the CSV editor, right-click a column header to rename, insert, delete, or change the type of that column. (#1913)
+- In the CSV editor, insert a row above or below any row from the row's right-click menu or the Edit menu. Deleting rows that contain data now asks you to confirm first. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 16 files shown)


datlechin:
Addressed a code review of this PR. Findings and fixes:
- **Undo naming (both directions)**: the private undo mirrors (`removeRow(suppressUndo:)`, `reinsertRow`, `reinsertRows`) never set the action name, so redo and every later undo degraded to a generic label. Threaded the action name through every `registerUndo` site, matching the `DataChangeManager` pattern.
- **Paste label clobber**: `appendRow()`'s new naming overwrote the "Paste" group name during a paste. Re-assert `setActionName("Paste")` after the append loop.
- Extracted the display→store insertion mapping into a testable `InspectorRowInsertion` helper (with tests), symlinked `CSVDocument` into the test target and added undo-naming tests, corrected the Edit-menu docs, and matched the plain-literal button style.

PR #1940, opened 2026-07-21
feat(inspector): insert columns left/right and delete selected columns in the CSV editor

datlechin:
Part 2 of 4 toward #1469 (CSV structural editing). Focused on **columns**. Stacked on #1939 (base branch `feat/1469-csv-row-editing`).

## What
- The column-header menu now reads **Insert Column Left / Insert Column Right** (renamed from the Before/After wording #1932 shipped unreleased), matching the issue and spreadsheet conventions (Numbers, Sheets, Tablecruncher).
- **Multi-column delete**: select several columns (click a header, Cmd-click others), and the menu reads **Delete Columns**, removing all of them in one undoable step. Deleting a data-bearing column (or columns) asks first.
- Insert Column Left / Right and Delete Column also appear in the **Edit menu**, acting on the selected column.

## Why
#1932 added single-column rename/insert/delete but always operated on the clicked column only, used Before/After labels, and never guarded a data-bearing delete. #1469 asks for Left/Right verbs that operate on the #1446 multi-column selection and confirm destructive deletes.

## How (native / HIG)
- The clicked column plus the current column selection resolve to a delete-target set inside the inspector, read from the shared grid's `selectionController.selection.affectedColumns`. When the clicked column is inside a multi-column selection, the verb operates on the whole selection; otherwise on the clicked column.
- Multiple `removeColumn` calls are wrapped in one `beginUndoGrouping`/`endUndoGrouping` and deleted highest-index-first so lower indices stay valid; the group is named ("Undo Delete Columns").
- Edit-menu items dispatch through the responder chain gated on `keyWindowIsInspector`, same as the row verbs in #1939.
- Delete confirmation is scoped to columns that actually hold data (a short-circuiting scan), and the copy does not claim the action is irreversible, since column deletes are undoable.

## Tests
- `InspectorColumnMenuBuilder`: Left/Right action wiring, singular vs plural Delete label, and the target-columns payload.
- `InspectorDeleteConfirmation.columnDeleteTitle` singular/plural.

## Docs / changelog
- `docs/features/csv-inspector.mdx` column-operations section updated.

Needs a local Xcode build to confirm compilation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Connect to PGlite through its socket server (`pglite-server`), with schema browsing, SQL editing, and row editing like PostgreSQL. Defaults to a loopback host with no TLS, and treats the connection as single-use to match PGlite's one-connection limit. (#1911)
 - In the CSV editor, right-click a column header to rename, insert, delete, or change the type of that column. (#1913)
 - In the CSV editor, insert a row above or below any row from the row's right-click menu or the Edit menu. Deleting rows that contain data now asks you to confirm first. (#1469)
+- In the CSV editor, insert a column to the left or right, and delete several selected columns at once. Column verbs also appear in the Edit menu, and deleting a column that holds data asks you to confirm first. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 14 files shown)


datlechin:
Addressed a code review (including the codex bot note). Findings and fixes:
- **Safety**: column verbs targeted `selection.affectedColumns`, which fires on any multi-column cell drag, so Edit > Delete Column could remove whole columns from an incidental selection (no confirm when empty). Column ops now use genuinely fully-selected columns via a new `GridSelectionController.selectedFullColumns()`.
- **Documented multi-select now works**: Cmd-click on a header was a replace, so accumulating columns was impossible. Cmd-click is now additive (`addEntireColumn`); single-select is unchanged.
- Extracted the target resolution into a pure, tested `InspectorColumnTargets` helper, and corrected the docs.

Issue #1472, opened 2026-05-28
feat(csv): read-only mode for files past a size threshold (memory-mapped / DuckDB-backed)


salmonumbrella:
### Problem

TablePro currently loads CSVs fully into memory. Open a 2 GB Stripe export or a multi-million-row log dump, and the app either pauses for minutes or fails. Tablecruncher opens 2 GB / 16M rows in 32 s on an M2 by skipping the full load; Modern CSV advertises load times "up to 11× Excel" via the same technique. This is the bar for being the only CSV tool on the user's Mac.

### Proposed solution

Read-Only Big-File mode that engages automatically past a configurable threshold (default 1 GB or a few million rows):

- File opens via memory-mapping, not full in-memory load
- Grid scrolls and renders normally — column-header context menu still offers `Copy Column Values`, `Copy as CSV / Markdown / IN Clause`, `Hide Column`, **Column Statistics…**, and #1454's per-column filter
- Disabled in this mode: cell edit, fill column, structural ops, find & replace
- A **Switch to Edit Mode** action loads the file fully into memory and re-enables every mutation surface
- A mode indicator lives in the window header so a greyed-out action is never confusing
- Threshold configurable in Preferences

### Implementation

- The DuckDB plugin's `duckdb.read_csv_auto` over a memory-mapped file is the most direct backing path. Query results stream into the grid via the existing datasource interface.
- Read-only as a primitive is independently useful. Some workflows want a guarantee the on-disk file isn't being silently dirtied during inspection.
- Tablecruncher's parser lives in `src/csvparser.cpp` / `src/csvdatastorage.cpp` and is GPLv3

PR #1941, opened 2026-07-21
feat(inspector): split and merge columns in the CSV editor

datlechin:
Part 3 of 4 toward #1469 (CSV structural editing). Focused on **Split / Merge**. Stacked on #1940 (base branch `feat/1469-csv-column-editing`).

## What
- **Split Column…** splits each value into new columns at every match of a delimiter or a regular expression. The clicked column is replaced by the pieces; short rows get empty cells. New columns are named `<column> 1`, `<column> 2`, ...
- **Merge Columns…** joins the clicked column with the one to its right into a single column, placing a separator you choose between the two values.
- Both appear in the column-header menu and the Edit menu, and each is a single named undo step.

## Why
#1469 asks for split/merge as first-class column verbs, matching Tablecruncher (its `csvmenu.cpp` Split/Merge dialogs), Modern CSV, and the Text-to-Columns / Join conventions from Excel and Sheets.

## How (native / HIG)
- New `CSVRowStore.splitColumn(at:spec:)` / `mergeColumns(at:separator:)` primitives. Because a split changes the column count per row, these materialize the affected rows (the lazy byte-range transform log can't express a split), then resync the header row and clear the transform log. This is edit-mode only; large read-only files are a separate concern (#1472).
- Split accepts a delimiter (`components(separatedBy:)`) or a compiled `NSRegularExpression`, chosen in the dialog via an `NSSegmentedControl`. An invalid regex shows an error and does nothing.
- Undo/redo use a full store-state snapshot captured before and after the change (cheap: Swift arrays are copy-on-write), since split/merge have no simple element-wise inverse. Inferred column types are re-derived and captured in the same snapshot.
- The dialogs are `NSAlert` sheets with an accessory view, matching the existing rename/insert-column prompt. Column layout stays correct because the grid already tolerates a stale `columnOrder` (unknown names skipped, new columns appended).
- `splitColumn`/`mergeColumns` are added to the `InspectorDocument` protocol with default no-op implementations. This is an additive PluginKit change (Library Evolution fills the default for existing conformers), so no `currentPluginKitVersion` bump.

## Tests
- `CSVRowStore.split(_:spec:)` for literal and regex separators; `splitColumn` padding; `mergeColumns` joining; `captureState`/`restore` round-trip.
- `InspectorColumnMenuBuilder`: Split/Merge wiring, and Merge omitted for the last column.

## Docs / changelog
- `docs/features/csv-inspector.mdx` column-operations section updated.

Needs a local Xcode build to confirm compilation. PluginKit change is additive (new protocol requirements with defaults); no version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, right-click a column header to rename, insert, delete, or change the type of that column. (#1913)
 - In the CSV editor, insert a row above or below any row from the row's right-click menu or the Edit menu. Deleting rows that contain data now asks you to confirm first. (#1469)
 - In the CSV editor, insert a column to the left or right, and delete several selected columns at once. Column verbs also appear in the Edit menu, and deleting a column that holds data asks you to confirm first. (#1469)
+- In the CSV editor, split a column into several by a delimiter or regular expression, and merge a column with the one next to it using a separator. Both are single undo steps. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


datlechin:
- **Double parse**: `splitColumn` fetched each row's cells twice (the second pass re-parsed from bytes past the 4000-entry cache). It now parses each row once and reuses the cells.
- **Column layout sync**: split/merge skipped the `columnLayout` maintenance every other structural handler does. Split now replaces the old column's layout key with the new column names (preserving position); merge drops the merged-away column's key.

PR #1942, opened 2026-07-21
feat(inspector): switch the CSV first row between header and data

datlechin:
Part 4 of 4 toward #1469 (CSV structural editing). The **header/data toggle**. Stacked on #1941 (base branch `feat/1469-csv-split-merge`).

## What
- **Edit > Switch First Row Between Header/Data** (`Cmd+Shift+H`) flips whether the first row is a header. Turning the header off moves it into the data and names columns `Column 1`, `Column 2`, ...; turning it back on promotes the first data row to the header. Undoable.
- A file that has no header row is no longer saved with a generated `Column 1,Column 2,...` header line. This also fixes byte round-trip for headerless files.

## Why
Auto-detection guesses whether row one is a header. When it guesses wrong, "Delete Column" and every other structural verb operate on the wrong interpretation, and (before this) saving a headerless file silently added a synthetic header. #1469 calls this out: without a header toggle, delete-column semantics are ambiguous.

## How (native / HIG)
- `CSVRowStore` gains an explicit `hasHeaderRow` flag (set from the same detection used at open) and `toggleHeaderRow()`, which moves the header row in or out of the data set and reindexes column names.
- `CSVWriter` now emits the header row only when `hasHeaderRow` is true, so the on-disk file matches the current header state and headerless files round-trip byte-for-byte.
- Toggle is one undo step through the same snapshot mechanism split/merge use (state captured before and after; `hasHeaderRow` is part of the snapshot). Inferred types are recomputed.
- `Cmd+Shift+H` is free in TablePro and not a macOS reserved shortcut (Finder uses it for Go > Home, which is app-scoped). The Edit-menu item is gated on the CSV window being frontmost.
- `toggleHeaderRow()` is added to `InspectorDocument` with a default no-op (additive PluginKit change, no version bump).

## Tests
- `CSVRowStore.toggleHeaderRow`: header to data, data to header, and a two-toggle round-trip.
- `CSVWriter`: a headerless file writes with no synthetic header row (byte-for-byte round-trip).

## Docs / changelog
- `docs/features/csv-inspector.mdx` new Header row section; `docs/features/keyboard-shortcuts.mdx` adds `Cmd+Shift+H`.

Needs a local Xcode build to confirm compilation. PluginKit change is additive; no version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, insert a row above or below any row from the row's right-click menu or the Edit menu. Deleting rows that contain data now asks you to confirm first. (#1469)
 - In the CSV editor, insert a column to the left or right, and delete several selected columns at once. Column verbs also appear in the Edit menu, and deleting a column that holds data asks you to confirm first. (#1469)
 - In the CSV editor, split a column into several by a delimiter or regular expression, and merge a column with the one next to it using a separator. Both are single undo steps. (#1469)
+- In the CSV editor, switch the first row between header and data with `Cmd+Shift+H` when auto-detection guesses wrong. Files without a header are no longer saved with a generated header row. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


datlechin:
- **Header demote correctness**: demoting read a possibly-stale `.materialized` header, so after an intervening column insert/remove the demoted row had the wrong width (a malformed CSV on save) or resurrected a deleted column's value. Demote now uses `columnNames`, which is always width-correct and value-synced. Added a promote → insert-column → demote regression test.
- **No-op guard**: `toggleHeaderRow` on an empty file marked the document dirty and pushed a dead undo entry. Guarded it.
- **Shortcut system**: Cmd+Shift+H was hardcoded outside the app's `ShortcutAction`/`KeyboardSettings` system (not customizable, not conflict-checked). Added a `toggleHeaderRow` action with the Cmd+Shift+H default and bound the menu item through `shortcut(for:)`.

PR #1943, opened 2026-07-21
feat(inspector): set CSV delimiter, quote, encoding, and line ending with reload

datlechin:
Part 5 toward #1469 (CSV structural editing). The **Set CSV Properties** dialog. Stacked on #1942 (base branch `feat/1469-csv-header-toggle`).

## What
- **Edit > Set CSV Properties…** opens a sheet to set the **delimiter**, **quote character**, **encoding**, and **line ending** by hand, prefilled with the values auto-detected at open. **Reload** re-reads the file with the chosen settings; **Cancel** leaves it as is.
- Reload discards unsaved edits, and asks first when there are any.

(The escape-character field is the last remaining property from the issue; it needs a change to the shared parser and lands in a follow-up so this dialog stays low-risk.)

## Why
Auto-detection can guess the delimiter or encoding wrong. #1469 asks for a manual override with a Reload so the user can fix it without leaving the app.

## How (native / HIG)
- The dialog is a document-scoped sheet (`presentAsSheet`), which the HIG recommends over a floating panel for a text/selection-heavy, document-bound task. It has a Reload primary and a Cancel, per the sheet guidance.
- A new `CSVConfigurableDocument` protocol (refining `InspectorDocument`, in PluginKit) exposes `csvDialect` and `reload(with:)`, so the app-target inspector reaches CSV-specific settings through a protocol instead of importing the plugin. Added as a new protocol, so it is additive (no version bump).
- `reload(with:)` re-parses the originally loaded bytes with the new dialect, re-infers types, clears the undo stack (like a revert), and marks the document edited so Save writes with the new settings.
- Option-to-dialect mapping lives in a pure `CSVPropertyOptions` helper so the picker choices and the resulting `CSVDialect` are unit-testable; the BOM is preserved from the original.

## Tests
- `CSVPropertyOptions`: index lookup from a dialect, unknown-value fallback, and building a dialect from picker indices (with BOM preserved).

## Docs / changelog
- `docs/features/csv-inspector.mdx` auto-detection section notes the override.

Needs a local Xcode build to confirm compilation. PluginKit change is additive (a new protocol); no version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, insert a column to the left or right, and delete several selected columns at once. Column verbs also appear in the Edit menu, and deleting a column that holds data asks you to confirm first. (#1469)
 - In the CSV editor, split a column into several by a delimiter or regular expression, and merge a column with the one next to it using a separator. Both are single undo steps. (#1469)
 - In the CSV editor, switch the first row between header and data with `Cmd+Shift+H` when auto-detection guesses wrong. Files without a header are no longer saved with a generated header row. (#1469)
+- In the CSV editor, set the delimiter, quote character, encoding, and line ending by hand from Edit > Set CSV Properties…, then Reload to re-read the file with those settings. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


datlechin:
Reviewed this PR. The review surfaced only a false positive and one below-threshold edge, so no code changes were needed (the branch was rebased onto the updated stack):
- **False positive**: `reload(with:)` was flagged for not clearing `typeOverrides`, but it calls `recomputeInferredTypes()`, which already resets `typeOverrides = [:]`.
- **Known limitation (niche)**: switching a BOM-detected file to a non-BOM encoding (e.g. Latin-1) and reloading would show the BOM bytes as leading garbage. A correct fix needs BOM-length tracking independent of encoding; deferring it since it only triggers on a deliberate mis-selection of a file whose encoding was already correctly detected from its BOM.

PR #1944, opened 2026-07-21
feat(inspector): choose the CSV escape character (doubled quote or backslash)

datlechin:
Part 6 (final) toward #1469 (CSV structural editing). The **escape character** field, completing the Set CSV Properties dialog. Stacked on #1943 (base branch `feat/1469-csv-properties`).

## What
- Set CSV Properties… gains an **Escape character** picker: a doubled quote (the RFC 4180 default) or a backslash. Files that escape an embedded quote as `\"` now read and save correctly.

## Why
`csvmenu.cpp` in Tablecruncher and DataGrip's format dialog both let you pick the escape mechanism, because real exports (notably MySQL) use backslash escaping instead of RFC 4180 quote doubling. This was the one property from #1469's dialog not yet wired.

## How
- `CSVDialect` gains an `escapeChar` field, defaulting to the quote byte (so the default is unchanged RFC 4180 behavior). It is a new stored property with a default; the existing public initializer is untouched, so this is additive to PluginKit (no version bump).
- `CSVStreamingParser` (indexRows, parseRow, field) honors the escape byte inside a quoted field only when it differs from the quote byte; when they are equal the existing doubled-quote path runs unchanged, so every existing file parses exactly as before.
- `CSVWriter` escapes an embedded quote (and the escape byte itself) with the escape character in backslash mode, and keeps doubling quotes in the default mode.
- The picker choice threads through the pure `CSVPropertyOptions` mapping into the reload dialect.

## Tests
- Parser: backslash escape decodes `\"` to a quote and `\\` to a backslash; the doubled-quote default is unchanged.
- Writer: doubles quotes by default, escapes with a backslash in backslash mode.
- `CSVDialect.escapeChar` defaults to the quote byte; `CSVPropertyOptions` maps the escape choice into the dialect.

## Docs / changelog
- `docs/features/csv-inspector.mdx` notes the escape-character choice.

Needs a local Xcode build to confirm compilation. PluginKit changes (a new `CSVDialect` field and parser method-body changes) are additive; no version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, split a column into several by a delimiter or regular expression, and merge a column with the one next to it using a separator. Both are single undo steps. (#1469)
 - In the CSV editor, switch the first row between header and data with `Cmd+Shift+H` when auto-detection guesses wrong. Files without a header are no longer saved with a generated header row. (#1469)
 - In the CSV editor, set the delimiter, quote character, encoding, and line ending by hand from Edit > Set CSV Properties…, then Reload to re-read the file with those settings. (#1469)
+- In the CSV editor, choose the escape character (a doubled quote or a backslash) in Set CSV Properties…, so files that escape quotes with a backslash read and save correctly. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


datlechin:
- **Doubled-quote escape bug**: the "Doubled Quote" escape option hardcoded byte `0x22` instead of tracking the selected quote character, so pairing **Single Quote** with **Doubled Quote** produced `quoteChar=0x27, escapeChar=0x22` and flipped the parser into backslash mode, corrupting single-quote-doubled files. Fixed: `CSVDialect.init` now defaults `escapeChar` to `quoteChar`, and the options mapping resolves "Doubled Quote" to the selected quote char.
- Added a `field(at:column:)` backslash-escape test and quote-tracking tests.

PR #1949, opened 2026-07-22
feat(mssql): auto-derive Kerberos realm for cross-realm Windows Authentication

### LINKED ISSUE #1947: SQL Server Windows Authentication (Kerberos) fails when the server's realm differs from the Mac's default_realm
## What happened

SQL Server **Windows Authentication (Kerberos)** fails with the generic error

> Failed to connect to `sqlserver.example.com:1433`: Check host, port, credentials, and TLS settings

whenever the SQL Server's Kerberos realm is **different from the Mac's `default_realm`**, even though the setup is valid and works from other clients:

- The user holds a valid TGT in their own realm (`CLIENT.REALM`).
- A cross-realm trust exists (`CLIENT.REALM` ⇄ `RESOURCE.REALM`, via `capaths`).
- The same server connects fine from DataGrip / the MS JDBC driver using `authenticationScheme=javaKerberos`.

## Root cause

FreeTDS builds the service principal **without a realm** — `MSSQLSvc/<host>:<port>`. On macOS, Heimdal assigns the client's `default_realm` to that unqualified principal and does **not** follow a cross-realm referral, so the KDC returns `KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN` (`-1765328377`) and the login packet is rejected.

DataGrip/Java succeed because Java's Kerberos resolves the SPN's realm from `[domain_realm]` (and/or follows AD referrals), so it requests the ticket from the realm where the SPN actually lives.

Evidence from a FreeTDS protocol dump (`TDSDUMP`):

```
login.c:1042: using GSS authentication
gssapi.c:299:  using kerberos name MSSQLSvc/sqlserver.example.com:1433
gssapi.c:417:  gss_init_sec_context: min_stat 2529638919 "Unknown error -1765328377"
gssapi.c:457:  gss_init_sec_context: GSS_S_FAILURE
login.c:670:   login packet rejected
```

Confirmed with Heimdal directly (the app never enters the picture):

```
# unqualified — Heimdal uses default_realm (CLIENT.REALM) and fails, no referral:
$ kgetcred MSSQLSvc/sqlserver.example.com:1433
kgetcred: krb5_get_creds: Server (MSSQLSvc/sqlserver.example.com:1433@CLIENT.REALM) unknown

# realm named explicitly — succeeds via the trust:
$ kgetcred MSSQLSvc/sqlserver.example.com:1433@RESOURCE.REALM   # OK
```

`[domain_realm]` does **not** help, because a service principal parsed from a string gets `default_realm`; the host→realm mapping is only applied to host-based names, not to the `MSSQLSvc/host:port` string FreeTDS constructs.

## Steps to reproduce

1. Mac with `default_realm = CLIENT.REALM` in `/etc/krb5.conf`.
2. A SQL Server whose SPN (`MSSQLSvc/<host>:1433`) is registered in a **different** realm (`RESOURCE.REALM`) with a cross-realm trust in place.
3. `kinit user@CLIENT.REALM`.
4. Create a SQL Server connection, Windows Authentication (Kerberos), principal `user@CLIENT.REALM`, connect by hostname.
5. Connect → fails with the generic "Check host, port, credentials, and TLS settings".

## Expected behavior

Connect succeeds (as it does from DataGrip/JDBC in the same environment). At minimum, the error should name the Kerberos cause (server principal unknown / wrong realm) rather than the generic message.

## Proposed solution

1. **Per-connection Kerberos realm / SPN realm option.** Let the user specify the realm so the SPN is built as `MSSQLSvc/host:port@REALM`. FreeTDS supports a `realm` setting for this. This avoids forcing users to change the machine-wide `default_realm`.
2. **Surface the real error.** A login-phase failure currently shows only the generic "Check host, port, credentials, and TLS settings". The underlying FreeTDS/GSS reason (here `S_PRINCIPAL_UNKNOWN`) never reaches the UI, which made this very hard to diagnose. Propagating the FreeTDS message/GSS minor status would help a lot.

## Workaround

Set `default_realm` to the SQL Server's realm in `/etc/krb5.conf` (downside: bare `kinit` then needs the user's own realm named explicitly, and other services in the client realm need their realm named). Verified this makes both `kgetcred` and TablePro succeed.

## Environment

- **TablePro:** 0.59.0
- **macOS:** 26.5.2, Apple Silicon (M1)
- **Database:** SQL Server (FreeTDS db-lib driver), Windows Authentication (Kerberos)
- **Kerberos:** macOS Heimdal, cross-realm trust between client and resource realms

Seha16:
Fixes #1947.

## Problem

SQL Server **Windows Authentication (Kerberos)** fails for a server whose Kerberos realm differs from the Mac's `default_realm`, even with a valid TGT and a working cross-realm trust. FreeTDS builds an **unrealmed** SPN (`MSSQLSvc/host:port`), and macOS Heimdal resolves that string principal against `default_realm` with no cross-realm referral — so the KDC returns `KRB5KDC_ERR_S_PRINCIPAL_UNKNOWN` and the login is rejected.

The JDBC driver (DataGrip) works in the same environment because Java derives the service realm from the **hostname** via the system `[domain_realm]` mapping — not from the URL, and without changing `default_realm`.

## Change

Make TablePro derive the realm the same way:

- **`MSSQLKerberosRealmResolver`** (macOS) resolves `host → realm` via GSS name canonicalization, which honors the system `[domain_realm]` / `default_realm` configuration (the already-linked GSS.framework; no new dependency).
- The resolved realm produces an explicit realm-qualified SPN (`MSSQLSvc/host:port@REALM`) handed to FreeTDS via `DBSETSERVERPRINCIPAL`. FreeTDS uses `server_spn` verbatim for the GSS target.
- **No new connection field and no `default_realm` change** required — it just works, matching the JDBC driver. When resolution yields nothing, it falls back to today's unrealmed behavior (no regression).
- iOS is unaffected (no GSS/Windows-auth path there); the SPN call is `#if os(macOS)`-guarded.

Also included (surfaced while testing this form):

- **Pin the Authentication method selector to the top** of the connection form. Previously it jumped position when a method showed/hid the built-in username/password (e.g. SQL Server's SQL vs Windows Authentication). The fix pins only selectors whose dependent fields hide credentials, so SQL Server / Snowflake / BigQuery / DynamoDB / etc. become consistent while MySQL/PostgreSQL are unchanged.

## Testing

- Unit tests for SPN construction (`MSSQLKerberosSPNTests`).
- Full app builds (Debug).
- Verified end-to-end against a real cross-realm setup (an AG listener and a standalone instance whose SPNs live in a different realm than the client's `default_realm`): connections that previously failed with `S_PRINCIPAL_UNKNOWN` now authenticate. The FreeTDS protocol dump shows the realm-qualified SPN and GSS completing:
  ```
  gssapi.c: using kerberos name MSSQLSvc/host:1433@RESOURCE.REALM
  gss_import_name: GSS_S_COMPLETE
  login.c: quietly sending TDS 7+ login packet
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - SSH tunnels using the None auth method now work on iPhone and iPad, not just the Mac. A connection synced from the Mac with None auth no longer fails to connect. (#1912)
 - Browsing a Trino table no longer fails with a syntax error. Trino requires `OFFSET` before `LIMIT`, so paging now uses `OFFSET` and `FETCH FIRST`. (#1936)
-- The Authentication method selector no longer changes position in the connection form when you switch between methods that show or hide the username and password (for example SQL Server's SQL vs Windows Authentication). (#1947)
+- The Authentication method selector no longer changes position in the connection form when you switch between methods that show or hide the username and password (for example SQL Server's SQL vs Windows Authentication, or PostgreSQL's password vs AWS IAM). (#1947)
 
 ## [0.59.0] - 2026-07-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, set the delimiter, quote character, encoding, and line ending by hand from Edit > Set CSV Properties…, then Reload to re-read the file with those settings. (#1469)
 - In the CSV editor, choose the escape character (a doubled quote or a backslash) in Set CSV Properties…, so files that escape quotes with a backslash read and save correctly. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
-- SQL Server Windows Authentication (Kerberos) now connects to servers whose Kerberos realm differs from your Mac's default realm. TablePro derives the realm from the server hostname using your Kerberos configuration (the same way the JDBC driver does), so it works with no extra setup and without changing your system `default_realm`. (#1947)
+- SQL Server Windows Authentication (Kerberos) now connects to servers whose Kerberos realm differs from your Mac's default realm. TablePro reads the `[domain_realm]` mapping from your Kerberos configuration, the same way the JDBC driver does, so a server that works in DataGrip works here without changing your system `default_realm`. Hosts with no mapping keep using the default realm. (#1947)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In the CSV editor, set the delimiter, quote character, encoding, and line ending by hand from Edit > Set CSV Properties…, then Reload to re-read the file with those settings. (#1469)
 - In the CSV editor, choose the escape character (a doubled quote or a backslash) in Set CSV Properties…, so files that escape quotes with a backslash read and save correctly. (#1469)
 - Add llama.cpp and MLX as local AI providers. Each preset points at the server's default local endpoint and needs no API key, alongside the existing Ollama and custom OpenAI-compatible options. (#1777)
+- SQL Server Windows Authentication (Kerberos) now connects to servers whose Kerberos realm differs from your Mac's default realm. TablePro derives the realm from the server hostname using your Kerberos configuration (the same way the JDBC driver does), so it works with no extra setup and without changing your system `default_realm`. (#1947)
 
 ### Fixed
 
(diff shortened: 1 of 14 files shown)


Seha16:
Addressed in 01b0027b — the resolver now returns the **canonical host + realm**, not just the realm:

- **Host canonicalization mirrors FreeTDS's own logic**: a dot-less (short) hostname is canonicalized to its FQDN via `getaddrinfo(AI_CANONNAME)`; a name that already contains a dot is used as-is. This restores the FQDN canonicalization FreeTDS performs for short names, which it skips once an explicit SPN is set.
- The **realm** is resolved via GSS name canonicalization (`[domain_realm]` / `default_realm`).
- The SPN is built from the **canonical** host and applied via `DBSETSERVERPRINCIPAL`, so short-name/CNAME and cross-realm hosts both resolve to the FQDN the service is registered under — matching the JDBC driver.

No behavior change for FQDN inputs; the short-name/alias path no longer regresses to `MSSQLSvc/alias:port@REALM`.

Seha16:
Thank you @datlechin for hopping on this PR so fast

PR #1956, opened 2026-07-25
fix(ai-chat): keep  AI chat inside and centered in  the right panel when resize

byteoxo:
Original problem:
telegram-cloud-photo-size-5-6276303433149124938-x

What's include:
keep  AI chat inside and centered in  the right panel when resize
When the right panel shrank, AI chat kept a wider ideal layout (model picker .fixedSize(), unbound messages/tables, composer text container not tracking width), centered in the host, and overflowed into the editor.

Clamp the panel to the host, let the model picker truncate, reflow messages and code, scroll wide tables inside the bubble, and sync the composer text container to the panel width on layout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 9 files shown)


datlechin (approved):
Thank you

PR #1958, opened 2026-07-25
feat(drivers): add SQL Server driver via tiberius (#1957)

Seha16:
## Summary

- Adds a Microsoft SQL Server driver to the Linux port — `crates/drivers/mssql` on `tiberius` (pure-Rust TDS) with rustls TLS — implementing the "MSSQL via tiberius" item from `ROADMAP.md`. Fixes #1957.
- Full `core::Connection`: connect (SQL auth, encryption from `use_tls`, trust-cert), introspection (tables / columns / indexes / FKs via `sys`, with PK / IDENTITY / computed / default detection), `OFFSET … FETCH` pagination, arbitrary + parameterized `query`/`execute`, and atomic `execute_in_transaction`.
- Core `sql_dialect`/`sql_ddl` gain `mssql` arms (`[bracket]` quoting, `@Pn` params, `IDENTITY(1,1)`, `sp_rename`, `ALTER COLUMN`, `DROP INDEX … ON`, `DROP CONSTRAINT`); other engines are byte-for-byte unchanged. Registered in the app — the registry-driven dialog lists "SQL Server" (port 1433).

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo test --workspace --lib` (type/default/error mapping + every dialect/DDL builder)
- [x] `cargo test -p tablepro-driver-mssql --test integration -- --include-ignored` against a real server — connect, introspection (PK/IDENTITY), all-types value roundtrip + `@P` binding, `OFFSET/FETCH` pagination, error mapping — **4/4 pass**
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`

Engine tested against: **SQL Server 2022 (CU14, `mcr.microsoft.com/mssql/server:2022-CU14-ubuntu-22.04`)**.

No new UI widgets or layout — the connection dialog is registry-driven, so "SQL Server" appears automatically; no before/after screenshots apply.

## Known limitations

- SQL Server authentication only (Windows/Kerberos auth is not modelled by `ConnectOptions`).
- Column reorder is unsupported (MySQL-only, matching existing gating); an `ALTER COLUMN` changing only a default is a no-op (MSSQL named-constraint management deferred).
- XML columns surface as text; one connection per session, serialized by a mutex.

PR #1962, opened 2026-07-25
feat(drivers): Windows integrated (Kerberos) auth for SQL Server (#1961)

### LINKED ISSUE #1961: Linux: Windows / Kerberos (integrated) authentication for the SQL Server driver
## Summary

Add **Windows integrated authentication (Kerberos / GSSAPI)** as an option for the Linux SQL Server driver, so users can connect with their current Kerberos ticket (from `kinit`) instead of a SQL Server login.

## Motivation

Many SQL Server deployments only permit Windows authentication. On Linux that means Kerberos: the user runs `kinit user@REALM` and the client authenticates to SQL Server with the ambient ticket. The MSSQL driver currently supports only SQL logins (username/password).

## Proposal

- `core`: `AuthMode { Password, Kerberos }` on `ConnectOptions`; `DatabaseDriver::supports_integrated_auth()` capability (default `false`).
- `drivers/mssql`: enable tiberius `integrated-auth-gssapi`; use `AuthMethod::Integrated` for Kerberos (ambient credential cache; SPN `MSSQLSvc/<host>:<port>`).
- `storage`: persist the auth mode (legacy files default to `Password`).
- `app`: a "Method" selector in the connect dialog, shown only for integrated-capable drivers; Kerberos hides the username/password rows.

## Known upstream-crate limitations (with workarounds)

1. **Cross-realm** — tiberius imports the SPN as a raw Kerberos principal, so it resolves in the client's *default* realm. For a SQL Server in a different realm, set `default_realm` (and `[capaths]` / `[domain_realm]`) so `MSSQLSvc/<host>:<port>` maps to the service realm. Proper fix: host-based SPN (`GSS_C_NT_HOSTBASED_SERVICE`) upstream in tiberius.
2. **Debug builds abort** — libgssapi 0.4.6 wraps GSSAPI's empty final-step token via `slice::from_raw_parts(null, 0)`, which trips Rust's UB precondition check and aborts — in debug builds only. Use a release build (debug-assertions off). Proper fix: null-guard upstream in libgssapi (0.4.6 is the latest `^0.4.5`).

## Acceptance

- Connect dialog offers Windows (Kerberos) for SQL Server; selecting it authenticates with the ambient ticket.
- Verified against a real (cross-realm) AD-joined SQL Server.
- `cargo build` / `clippy` / `fmt` / unit tests green; existing SQL-auth integration tests unaffected.

Seha16:
## Summary

Adds **Windows integrated authentication (Kerberos / GSSAPI)** to the Linux SQL Server driver — connect with your `kinit` ticket instead of a SQL login. Fixes #1961.

- `core`: `AuthMode { Password, Kerberos }` + `ConnectOptions.auth_mode`; `DatabaseDriver::supports_integrated_auth()` (default `false`).
- `drivers/mssql`: tiberius `integrated-auth-gssapi` feature; `connect()` uses `AuthMethod::Integrated` for Kerberos (ambient credential cache, SPN `MSSQLSvc/<host>:<port>`).
- `storage`: `SavedConnection.auth_mode` persisted (legacy files default to `Password`).
- `app`: connect-dialog "Method" selector, shown only for integrated-capable drivers; Kerberos hides username/password; validity / label / dedup updated.
- CI + README: `libkrb5-dev` and `clang` (libgssapi-sys links `gssapi_krb5` and runs bindgen).

## Depends on #1958

This branch is stacked on the SQL Server driver PR (#1958). Until that merges, this PR's diff also contains #1958's commit (`feat(drivers): add SQL Server driver via tiberius`). Please review/merge after #1958 — the diff then reduces to the single Kerberos commit.

## Test plan

- [x] `cargo build --workspace`; `cargo test --workspace --lib` (core + storage `auth_mode` round-trip/legacy default; mssql `supports_integrated_auth`)
- [x] `cargo fmt --all -- --check`; `cargo clippy --all-targets -- -D warnings`
- [x] No regression: mssql integration **4/4** vs. real SQL Server 2022 (SQL auth)
- [x] UI: connect dialog shows **Method = Windows (Kerberos)** for SQL Server, credential rows hidden
- [x] **Live Kerberos connection verified** against a real cross-realm AD SQL Server (client realm ≠ service realm, cross-realm trust)

## Known upstream-crate limitations (with workarounds)

1. **Cross-realm SPN realm** — tiberius imports the SPN as a raw Kerberos principal, so it resolves in the client's *default* realm. For a service in another realm, set `default_realm` (+ `[capaths]` / `[domain_realm]`) so `MSSQLSvc/<host>:<port>` maps to the service realm. Proper fix: host-based SPN (`GSS_C_NT_HOSTBASED_SERVICE`) upstream in tiberius.
2. **Debug builds abort** — libgssapi 0.4.6 wraps GSSAPI's empty final-step token via `slice::from_raw_parts(null, 0)`, tripping Rust's UB precondition check (debug only). Run a **release build** (debug-assertions off) for Kerberos. Proper fix: null-guard upstream in libgssapi (0.4.6 is the latest `^0.4.5`).

Neither limitation affects CI (build + unit tests; no live handshake) or the SQL-login path.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -38,7 +38,10 @@ jobs:
         # ubuntu:25.10 is minimal — git for Swatinem/rust-cache key
         # generation, curl + ca-certificates for rust-toolchain's
         # rustup install, plus the GTK / libadwaita / sourceview /
-        # OpenSSL / libsecret -dev packages the workspace links.
+        # OpenSSL / libsecret / MIT Kerberos -dev packages the workspace
+        # links. libkrb5-dev + clang cover the MSSQL driver's integrated
+        # (Kerberos/GSSAPI) auth: libgssapi-sys links gssapi_krb5 and runs
+        # bindgen (needs libclang).
         # No sudo (container runs as root by default).
         run: |
           apt-get update
@@ -52,7 +55,9 @@ jobs:
             libadwaita-1-dev \
             libgtksourceview-5-dev \
             libssl-dev \
-            libsecret-1-dev
+            libsecret-1-dev \
+            libkrb5-dev \
+            clang
       - uses: dtolnay/rust-toolchain@1.93
         with:
           components: rustfmt, clippy
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -38,10 +38,10 @@ jobs:
         # ubuntu:25.10 is minimal — git for Swatinem/rust-cache key
         # generation, curl + ca-certificates for rust-toolchain's
         # rustup install, plus the GTK / libadwaita / sourceview /
-        # OpenSSL / libsecret / MIT Kerberos -dev packages the workspace
-        # links. libkrb5-dev + clang cover the MSSQL driver's integrated
-        # (Kerberos/GSSAPI) auth: libgssapi-sys links gssapi_krb5 and runs
-        # bindgen (needs libclang).
+        # OpenSSL / libsecret -dev packages the workspace links, plus
+        # libkrb5-dev + clang for the MSSQL driver's integrated
+        # (Kerberos/GSSAPI) auth (libgssapi-sys links gssapi_krb5 and
+        # runs bindgen).
         # No sudo (container runs as root by default).
         run: |
           apt-get update
@@ -70,11 +70,6 @@ jobs:
         run: cargo clippy --all-targets -- -D warnings
       - name: Build
         run: cargo build --workspace
-      # The MSSQL driver's Kerberos support is a default-on cargo feature
-      # so packagers without MIT Kerberos / libclang can drop it. Check
-      # that path here or it rots.
-      - name: Build without Kerberos
-        run: cargo check -p tablepro-driver-mssql --no-default-features --all-targets
       # --bins matters: tablepro-app has no lib target, so --lib alone
       # skips every test in the app crate.
       - name: Unit tests
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -70,8 +70,15 @@ jobs:
         run: cargo clippy --all-targets -- -D warnings
       - name: Build
         run: cargo build --workspace
+      # The MSSQL driver's Kerberos support is a default-on cargo feature
+      # so packagers without MIT Kerberos / libclang can drop it. Check
+      # that path here or it rots.
+      - name: Build without Kerberos
+        run: cargo check -p tablepro-driver-mssql --no-default-features --all-targets
+      # --bins matters: tablepro-app has no lib target, so --lib alone
+      # skips every test in the app crate.
       - name: Unit tests
-        run: cargo test --workspace --lib
+        run: cargo test --workspace --lib --bins
 
   integration:
     name: Driver integration tests (docker)
--- a/linux/Cargo.lock
+++ b/linux/Cargo.lock
@@ -80,6 +80,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -127,7 +136,7 @@ dependencies = [
  "futures-core",
  "libc",
  "portable-atomic",
- "rustc-hash",
+ "rustc-hash 2.1.2",
  "tokio",
  "tokio-stream",
  "xattr",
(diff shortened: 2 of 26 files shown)


Seha16:
Testing evidence on Arch Linux Kernel 7.1.5 with krb5 (Kerberos) with default_realm patched to the needed one:

https://github.com/user-attachments/assets/9570ef62-fc22-4923-a46d-936acabe3418

datlechin:
Pushed `fefd543` with fixes for the review findings. All of it is verified locally with the exact CI commands (`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo build --workspace`, `cargo test --workspace --lib --bins`) plus a `--no-default-features` pass, except the flatpak item called out at the bottom.

### Kerberos through an SSH tunnel was asking for the wrong SPN

`establish()` rewrites `opts.host`/`opts.port` to the tunnel's local forward before `driver.connect()`, and tiberius builds both the SPN and the TLS server name from `Config::host`/`port`. A tunnelled connection was asking the KDC for `MSSQLSvc/127.0.0.1:<ephemeral>`.

The socket and the service name are now separate: `ConnectOptions::service_endpoint` records the endpoint the tunnel replaced, `service_address()` returns it, and the driver configures tiberius from that while opening the socket against `opts.host`. The tuple form of `TcpStream::connect` also handles IPv6 literals, which `config.get_addr()`'s `host:port` string did not.

### Kerberos was writing a password into the keyring

`run_connect` called `store_password` unconditionally, and `AdwPasswordEntryRow` keeps its text after the row is hidden. Type a password, switch to Kerberos, connect, and that credential landed in libsecret against a connection that never uses it. The credential rows are now emptied at collection time under Kerberos, the store is skipped, and `open_saved` no longer reads the keyring back for a Kerberos entry.

### Auth mode no longer lives in widget visibility

`selected_auth_mode()` read `auth_combo.is_visible()`, and `refresh_auth_visibility()` called back into it, so state flowed through GTK flags and back. There is now an `AuthFormState { file_based, supports_integrated, kerberos_selected }` on the model with `mode()` / `shows_method()` / `shows_credentials()`, and the widgets are driven from it. Five unit tests cover the combinations, including a stale Kerberos selection surviving a driver switch and file-based drivers asking for nothing.

That also fixes two things the visibility coupling had introduced:

- `DriverChanged` never called `refresh_validity()`. With SQL Server + Kerberos the form was valid with an empty username; switching back to PostgreSQL brought the username row back but left Connect enabled.
- Dropping the `file_based` visibility lines meant `refresh_validity()` put the `.error` class on a hidden username row for SQLite.

### Kerberos is a cargo feature now

`integrated-auth-gssapi` was on the workspace `tiberius` dependency, so `libkrb5-dev` and `libclang` were mandatory for anyone building the workspace. It moved to `tablepro-driver-mssql`'s `kerberos` feature, default on, so `--no-default-features` drops both. `supports_integrated_auth()` is `cfg!(feature = "kerberos")` so the UI never offers a mode the binary cannot perform, and a Kerberos request in a build without it returns the new `DriverError::Unsupported` instead of silently doing a SQL login. CI checks that configuration so it does not rot.

### Debug builds no longer abort

```toml
[profile.dev.package.libgssapi]
debug-assertions = false
```

The UB precondition check is emitted per crate, so turning it off for that one package makes `cargo run` usable for Kerberos without weakening the check anywhere else. Worth dropping once the null guard lands upstream.

### Also

- `cargo test --workspace --lib` was skipping the app crate entirely, since `tablepro-app` has no lib target. 97 existing tests had never run in CI. The step is `--lib --bins` now; all 97 pass headless.
- README documents `kinit`, the SPN, the cross-realm `krb5.conf` setup, and the no-Kerberos build. Those limitations only existed in the PR description before.
- `AuthMode::Kerberos`'s doc said drivers without support "fall back to password auth". None did; the driver now rejects it, and the doc says so.
- The connection label keys off `auth_mode` rather than an empty username.

### Two things I did not do

**The flatpak manifest is unverified.** `org.gnome.Sdk` 47 ships gcc, not clang, and bindgen needs libclang, so I added the `llvm18` extension with `LIBCLANG_PATH`. I could not run `flatpak-builder` from here, and nothing in CI builds the flatpak. Whether the freedesktop 24.08 SDK carries krb5 headers also needs a real build to confirm. Please run it once before merging.

**`po/tablepro.pot` is untouched.** "Method" and "Windows (Kerberos)" are still unextractable, but `POTFILES.in` already lists `ui/app.rs`, `ui/edit_dialog.rs`, and `ui/insert_dialog.rs`, none of which exist any more, so `xtr` cannot run at all. That predates this PR and fixing it properly is its own change.

Cargo.lock is unchanged: `kerberos` is default-on, so resolution is identical.

Seha16:
Testing evidence on Arch Linux Kernel 7.1.4 after rebasing onto the `main` branch

https://github.com/user-attachments/assets/8ee73337-f2cc-448e-97e2-45af3ad6f00b

Seha16:
Testing evidence after refactoring of Kerberos/Windows Auth

https://github.com/user-attachments/assets/5438e359-d2df-4ae1-ba62-767ebcac1dbc

datlechin:
Hi @Seha16 could you please resolve the conflict?

Seha16:
Testing evidence after rebase and conflict resolution @datlechin 

https://github.com/user-attachments/assets/0e141036-4c18-49b6-b68f-f7ba72ac7872

PR #2093, opened 2026-08-11
fix(connections): scope the browse cursor to the window, not the connection

### LINKED ISSUE #2088: Mixed use of tabs across multiple workspaces
### What happened?

When I opened two database workspaces. During the switching process between the two databases. The tabs may not follow the workspace switch, which may lead to confusion.

### Steps to reproduce

1. Open a db
2. create a tab(query tab or db table)
3. Open another db
4. tab may working in before workspace
5. create a tab(query tab or db table)
6. switch workspace, tab will confusion with workspace

### Expected behavior

_No response_

### Database type

None

### TablePro version

0.64.0

### macOS version & chip

macOS 26.6 / Apple Sillicon

### Screenshots / Logs

_No response_

buiducnhat:
## Problem

Two databases of one connection share one native macOS tab bar, because `WindowManager.tabbingIdentifier(for:)` keys only on the connection. Switch between those tabs and the window subtitle correctly names the tab's database, but the sidebar lists the other one, the workspace rail highlights the other one, and Cmd+T opens a tab on the other one.

That last part is the substantive half: a new tab created from the window showing `shop` lands on `analytics`.

## Root cause

"Where am I browsing" was stored once per connection, as `ConnectionSession.browseDatabase`/`browseSchema` in `DatabaseManager.activeSessions`. The real cardinality is once per window. Every reader of that field is window-scoped code, and every write broadcast `connectionStatusChanged`, which every other window answered by re-reading the shared value. So one window's switch did not merely fail to notify the others, it overwrote them.

Nothing realigned the cursor when a window became key either: `handleWindowDidBecomeKey` syncs the sidebar's table *selection*, never the container.

Existing tabs always queried the right database, because execution already routes through `DatabaseScope`. This was a browsing-consistency bug, not data corruption.

## What changed

Five commits, each building and lint-clean on its own.

1. **`refactor(connections)!`** — `browseDatabase`/`browseSchema` become `driverDatabase`/`driverSchema`, `browseDatabaseName`/`browseScope` become `driverDatabaseName`/`driverScope`. Their honest meaning is where the one shared driver is pinned and what a reconnect restores, which is genuinely per connection. Atomic: every caller and both test doubles in the same commit.

2. **`feat(connections)`** — new `WindowBrowseState`, owned per window by `MainContentCoordinator` and mutated only through `applyBrowseState`. `switchDatabase`/`switchSchema` move their own window's cursor and roll it back on failure. `SessionStateFactory` seeds a new window from the container it was opened for. `ConnectionToolbarState.syncFromSession` no longer resolves the chip from the shared session, which is what let one window repaint every other window's chip.

3. **`fix(workspace-rail)`** — `browsedWorkspace(ofWindow:)` replaces `browsedWorkspace(for: connectionId)`, and `resolveEntries` gains a `browsedContainers:` closure so each window contributes its own row. The rail reads its own window through the same `coordinator(forWindow:)` lookup `activate` already used for others.

4. **`refactor(connections)`** — `SchemaService` and `SchemaProviderRegistry` rekeyed from `UUID` to `DatabaseScope`, so two windows no longer share one table list or one autocomplete provider. Teardown stays connection-wide: `invalidate(connectionId:)` clears every scope. `DatabaseTreeMetadataService` and `MetadataConnectionPool` were already scope-keyed and are unchanged; they were the reference shape.

5. **`fix(launch)`** — the persisted browse container becomes per window group, so two windows saved on different databases come back on their own. Additive JSON field decoded with `decodeIfPresent`; the old single value remains the fallback for state written by older builds, and is still written so a rollback lands somewhere real.

Engines that genuinely cannot hold two containers at once (`requiresReconnectForDatabaseSwitch` and no pooling, i.e. PGlite) keep the shared cursor behind one boolean gate. No new user-facing strings anywhere.

## Invariants respected

- *A refresh never clears the cache it is refreshing* — the `.loading` guard, `markLoadFailed` refusing to overwrite `.loaded`, and the load-generation checks keep their exact shape, one key level finer.
- *Schema loading* — `SQLSchemaProvider` is now one instance per scope, so its in-flight `loadTask` dedupe is still correct rather than racing two databases through one set of fields.
- *An emptied tab manager is not the same as the user closed every tab* — an automatic save still never deletes, and a group whose coordinator has torn down keeps its last known cursor instead of writing nil.

## Tests

`swiftlint lint --strict` clean, Debug build clean.

New: `WindowBrowsePersistenceTests` (9 cases) covering two groups restoring onto their own containers, the legacy single-value fallback, and round-trip encoding. Extended `WorkspaceRailStoreTests` with the three cases that would have caught this: two windows browsing different containers each earning a row with no tabs open, a window's cursor outranking the driver position, and the pre-cursor fallback to the saved default.

Green: `SchemaServiceTests`, `SchemaServiceRefreshTests`, `SchemaServiceRoutinesTests`, `SchemaServiceHierarchicalTests`, `SchemaProviderRegistryTests`, `SchemaRefreshServiceTests`, `TableStructureLoaderScopeTests`, `WorkspaceRailStoreTests`, `WorkspaceAnchoringTests`, `TabPersistenceCoordinatorTests`, `WindowGroupAssignmentTests`, `DataRefreshScopeTests`, `MultiConnectionTests`, `HealthMonitorReconnectTests`, `SwitchDatabaseTests`, `SwitchSchemaTests`, `SwitchContainerTests`.

No UI automation added: the flow needs two windows of one connection focused in sequence, and multi-window focus timing does not run deterministically here.

## What happened to the original approach

This PR started as `selectTab(in:target:)`, selecting a tab matching the target container after a rail switch. Tracing it showed it could not take effect: `activate` resolves its window through `window(showing:target:)`, which already filters to a coordinator holding a tab whose container matches; one window holds effectively one tab (every `addTab` call site is guarded by `tabs.isEmpty` and otherwise routes to `WindowManager.openTab`, which mints a new window); and every `tabs.append` sets `selectedTabId`. So the match was already selected and the assignment was a no-op. In the `mostRecentWindow` fallback branch no tab matched and it returned early. Codex's P2 on the `current == workspace` guard was real, but moving the call above it would not have helped for the same reason.

It also targeted the one direction that already worked. The rail switch path was fine; the gap was that switching native tabs never moved the cursor at all. `selectTab` is removed here because a per-window cursor makes it unnecessary. @buiducnhat's two commits remain at the base of the branch.

## Known follow-ups

- `SidebarViewModel` is still one instance per connection, so its filter caches now key on a connection-wide generation token. Two windows on different databases whose table lists happen to have the same count could hit a stale filter cache. Making that view model per window is a separate change.
- `SharedSidebarState.searchText` / `favoritesSearchText` are still shared between windows of a connection. Not part of the reported bug, so left alone deliberately.

Closes #2088


datlechin:
Thanks for digging into this, and for the trace that localised it to the rail switch path.

I took the fix a different way in #2096, and I want to explain why rather than just redirect you.

I traced `selectTab(in:target:)` and it can't take effect in practice. `activate` resolves the window through `window(showing:target:)`, which already filters to a coordinator holding a tab whose container matches. One window holds effectively one tab (every `addTab` call site is guarded by `tabs.isEmpty` and otherwise routes to `WindowManager.openTab`, which mints a new window), and every `tabs.append` sets `selectedTabId`. So in that branch the matching tab is already selected and the assignment is a no-op. In the `mostRecentWindow` fallback branch no tab matches, so it returns early. Codex's P2 about the `current == workspace` guard is real too, and moving the call above it wouldn't have helped for the same reason.

The deeper issue is that the rail switch path was the one direction that already worked. The browse cursor lived once per connection on `ConnectionSession`, so switching native tabs never moved it, and every window's sidebar, rail highlight and autocomplete read that one shared value. Cmd+T from the window showing `shop` opened a tab on `analytics`. #2096 makes the cursor per window and keys the schema and autocomplete caches by `DatabaseScope`.

Your two commits are at the base of that branch, so your authorship is preserved in the history. Closing this in favour of #2096.

datlechin:
Correction to my comment above: I said I was closing this in favour of #2096. I've done the opposite. The full fix is now pushed onto this branch, #2096 is closed, and your two commits are still at the base of the history here. Sorry for the churn.

The reasoning about `selectTab` in that comment stands, and the PR description is updated to match what the branch now contains.

datlechin:
Hi @buiducnhat, thanks for the contributing. I'll going to close this PR because I've a big refactor on the connection/workspace/rail and so may it will fix the main issue of this PR trying resolve.

Could you may please run the latest commit on your local then test again see the main issue is fixed or not? Thank you.

PR #2097, opened 2026-08-12
feat(connections): host every open connection in a single window

datlechin:
## Why

A user opened two connections and got two separate windows, each showing the full workspace rail listing every open connection, and asked whether they could be merged into one window.

That is not a stray bug. Since 0.64.0 every connection gets its own window, and `WindowManager.tabbingIdentifier(for:)` keys the native tab group on the connection id, so two connections can never share a window. Every window still renders the rail listing all of them, so the app names all your connections in N places and makes you window-switch between them. No mature client does both: tools with a persistent connection sidebar have one window, tools with a window per connection put the connection list behind an on-demand switcher.

The tell that this was already known: `TabWindowController.frameAutosaveName` is one shared slot for every connection, so the windows stack and a rail click "reads as the window changing content rather than a different window being raised". That illusion survives one interaction. Every window writes that key on resize and move, only new windows read it back, and `recomputeWindowMinSize()` resizes windows on its own.

This is the first of three PRs moving the app to a single-window model. It is scaffolding, with no user-visible change.

## What this changes

The connection was bound into the object graph at the wrong layer. `MainSplitViewController` fixed its connection at `init` and held the session, phase, and panel state as scalars, because a window served exactly one connection for its whole life.

- `ConnectionWorkspace` holds everything a window needs to present one connection: phase, attempt token, session, session state, right-panel state.
- `ConnectionWorkspaceRegistry` holds the set of connections a window hosts and which one it shows, with ordering, neighbour selection on removal, and wrap-around cycling.
- `MainSplitViewController` reads those fields through the registry instead of storing them. Workspace construction moved out of `init` into `adoptWorkspace(payload:autoConnect:)`, so a window can call it more than once.
- Status reconciliation fans out over every hosted connection instead of only the visible one. A background workspace that loses its session still reaches the right phase, so switching to it later cannot show content for a connection that is already gone.
- `WindowManager.connectionIdsRetainingRestoreIntent()` collects per workspace rather than per window.

The registry holds one entry throughout this PR, so behaviour is identical to before.

## Two decisions worth reviewing

**A late connect attempt discards itself.** `finishAttempt` now takes the connection id and returns early when the registry no longer holds that workspace. The missing entry is the generation check, so a connect that outlives a connection the user closed cannot write a phase back and resurrect it. `transition(to:for:)` repaints only when the target workspace is the one on screen.

**Each workspace owns its undo stack.** `NSWindow.undoManager` was the right source while a window meant one connection. One window shared between connections would let Cmd+Z in one roll back an edit made in another, so `ConnectionWorkspace` carries its own `UndoManager`. `DataChangeManager` still routes through the window for now and moves over in PR 2, when a window can actually hold more than one connection.

## Testing

`ConnectionWorkspaceRegistryTests`, 10 cases: insertion order and selection, idempotent insert, phase and attempt-token isolation between workspaces, distinct undo managers, neighbour selection on removal, clearing selection when the last workspace goes, ignoring selection of an unhosted connection, a removed workspace being unreachable, and cycling in both directions.

No UI automation, because this PR has no user-visible behaviour to drive. The UI automation that would have caught the original report lands in PR 3, which is where two connections start sharing a window.

Build succeeds and `swiftlint lint --strict` is clean.

## Not in this PR

The `TabWindowController` and `MainSplitViewController` renames the design called for are left out. They touch around 40 files with no behaviour change and would bury this diff. Worth doing as separate cleanup.

## Next

- PR 2: editor tabs stop spawning windows. A custom tab strip on `NSTabViewController` with `tabStyle = .unspecified`, `newTab()` appending to the existing tab manager, and the four window-reconciliation files deleted.
- PR 3: connections consolidate into one window. The rail switches content in place, the frame hack goes, and window-level tabbing becomes native and shared so Merge All Windows and drag-out keep working.


datlechin:
## Sidebar rebuilt on `NSOutlineView`, plus the PR #2104 regressions

One commit, 94 files. It is one body of work: a SwiftUI `List` hosted in a bare `NSHostingController` cannot draw an emphasized selection or answer an arrow key, so every sidebar list became an outline, and the chrome PR #2104 changed went back to what AppKit draws.

### Sidebar

- The flat, hierarchical and Favorites lists are all `NSOutlineView` now. Measured cause: hosted in a bare `NSHostingController` with no SwiftUI scene above it, the backing `SwiftUIOutlineListView` returns **false** for `responds(to: #selector(NSResponder.moveDown(_:)))`. No keyboard, no emphasis, and no way to opt in.
- Section titles are real source list group rows. `isGroupItem` is what stops AppKit indenting their children, which is why a table now lines up with a database the way a package lines up with the project in Xcode's navigator. `rowSizeStyle = .small` and `indentationPerLevel = 13` are AppKit's own numbers, not picked by eye.
- Clicking a table opens it. It used to wait out `doubleClickInterval` first. Double-click is disclosure only; **Open in New Tab** on the contextual menu pins a preview tab.
- Holding an arrow key opened a tab and ran a query for **every row it passed**. Arrowing through 20 tables fired 20 queries. Keyboard opens now coalesce on `NSEvent.keyRepeatInterval`.
- Focus: `SidebarOutlineView` states the contract both lists share. It claims first responder on `mouseDown`, which is what `KeyHandlingTableView` already does for the data grid, and fills the first-responder vacancy a window is left with because a SwiftUI-hosted list does not exist when AppKit consults `initialFirstResponder`.

### Two defects worth calling out

**`NSEvent.keyCode` read on a mouse event.** `isArrowNavigation(type: event.type, keyCode: event.keyCode)` evaluates both arguments before the type check inside, so a mouse-driven selection raised inside AppKit's own click handling. The raise unwound through `NSTableView`'s mouse tracking: no open, no mouse up, and a tracking loop that only ended when the next click arrived, which is why clicking a table appeared to need three clicks. The type check now lives with the `keyCode` read so no caller can bypass it. Present on `main`, latent until single click became the open gesture.

**No way to add a table to Favorites.** `TableRow` gates the star on `onToggleFavorite != nil` and the outline never passed it, leaving `FavoriteTablesStorage.toggle` with zero call sites. A fresh install could not create one.

### From PR #2104

Alert key equivalents, the toolbar (tooltips, Import, Results state), the tab strip back in the content area rather than a titlebar accessory, the theme slot, tag filter contrast, grid cell selection fill, nested connection groups, and drag-and-drop onto a window.

### How this was verified

Behaviour was measured, not inferred. AppKit harnesses under `xcrun swiftc` established the source list geometry and the emphasis rules; the running app was driven with synthesized events and read back through Accessibility to settle where the keyboard actually was. Several early diagnoses were wrong and were discarded when the measurement disagreed.

### Not done

- **No UI automation.** Rule 4 asks for it and this PR does not have it. Every one of the four highest-severity defects here was of a kind unit tests cannot reach, which is exactly the argument for adding it.
- Full test target not run locally; the affected suites pass (340+). CI is the check.
- No command moves the keyboard from the sidebar into the grid. Removing the gesture left the gap; Xcode has ⌘J for this.
- `pendingGridFocusOnOpen` is still a latch that can outlive the open that set it. Reachable via Show Structure on a tab with no grid attached.

PR #2099, opened 2026-08-12
refactor(editor): make completion follow the tab's own database, and every metadata read follow the type

J2TeamNNL:
## Summary

TablePro already shows a per-tab database picker in the query editor toolbar (`QueryContainerPicker`), and its own doc comment says "the tab owns the new database for the rest of its life and the sidebar's browse cursor stays where the user left it."

Autocomplete ignored that picker. This makes completion honour it, and fixes the second half of the same defect class in plugin metadata.

## Root cause

One mechanism, two surfaces: the app resolved "which database is this work for?" from **ambient** state rather than from the identity that owns the work.

- `SchemaProviderRegistry` keyed providers by `connectionId` and sourced every fetch through `withBrowseMetadataDriver`, which is the **sidebar's** cursor. Every tab of a connection shared one provider bound to whatever the object browser last pointed at.
- Metadata reads keyed by `pluginTypeId`, the **primary plugin's** id rather than the type's own. A variant (Redshift, CockroachDB, PGlite) was served the primary's facts.

Both are a key that cannot express a distinction the app needs.

## What changed since the first version of this PR

The metadata half is no longer a 6-site patch. `snapshot(forTypeId:)` is renamed `snapshot(forRegisteredTypeId:)`, so every site that passed a `DatabaseType` stopped compiling until it was classified: **83** now ask `snapshot(for:)` and **6** remain genuine plugin-identity lookups, each commented as such.

Letting the compiler enumerate the sites rather than a regex found **8 multi-line calls that a line-based grep missed**, including `DatabaseManager.canPool()` reading `postConnectActions`, which decides pooling eligibility. The real count was 70 collapsed sites, not the 57 first estimated.

Three corrections came from measured probes rather than from reading the code:

- Cancelling a caller that awaits a shared `Task.value` neither cancels that task nor returns early, so a superseded resolution always runs its tail to the commit. Refusing to cache was half a guard; the stale value must not be **returned** either.
- `@Observable` tracks the whole stored property, and `dict[k, default: 0] &+= 1` notifies through the ungated `_modify` even when nothing changed. The per-scope `revisions` dictionary bought nothing over one counter while costing a body re-evaluation per editor, so invalidation now uses one observable box per scope inside a plain container.
- SwiftUI re-runs a `.task` at an **unchanged** id when its host view is unparented and reparented, which is what every connection switch does. That is why a failed resolution is no longer cached: an uncached failure heals for free on the next switch, a cached one never does.

`serverVersion` left the cache key entirely. Keying on it meant reading it from a SwiftUI body through `DatabaseManager.driver(for:)`, and `PluginDriverAdapter` is not observable, so a version learned after connect reached no body and the profile stayed cached under a nil-version key for the life of the tab. It changes only across a reconnect, which already clears the cache.

## Collateral, found while investigating and shipped here

- **PGlite** was offered SSH, SSL, Cloudflare Tunnel and SOCKS panes, plus `~/.pgpass`, the AWS IAM block and a password field, for an embedded WASM engine on loopback. Its own curated entry declares none of them, and three shipped docs pages say it has none.
- **Redshift's `padb_harvest` and CockroachDB's `system`** listed in the sidebar as ordinary user databases, because `systemDatabaseNames` resolved to PostgreSQL's empty list.
- **The AI schema context listed no columns for any table named with a capital letter.** It wrote `columnsByTable[table.name.lowercased()]` while `AISchemaContext` read `columnsByTable[table.name]`. The two other producers in the repo are each self-consistent; this was the only mismatched pair.

## One thing worth a reviewer's attention

Narrowing `additionalConnectionFields` for a variant would also have narrowed the **secret-redaction and Keychain-deletion lists** that read it. A PGlite connection saved while the form was still wrongly offering PostgreSQL's fields can hold a secret under a field PGlite never declares, so the exact list would have orphaned it in the Keychain and stopped redacting it on export.

`PluginManager.secureConnectionFieldIds(for:)` returns the union of the variant's fields and its serving plugin's. All five deletion and redaction paths use it; rendering a form stays exact, which is the opposite requirement.

Checked separately: this downgrades TLS nowhere. Redshift, CockroachDB and MariaDB keep `supportsSSH` and `supportsSSL` true; ScyllaDB and Turso have no curated entry so they fall back to their primary exactly as before; PGlite is the only type that flips, correctly.

## Caught in self-review

A `code-review` pass over the finished diff found eight issues; seven are fixed here.

The one that mattered most was mine and would have shipped a broken upgrade. `DatabaseDriver.buildAdditionalFields` loads a connection's secure fields from the Keychain by the **rendered** field list. Once a variant resolves its own curated entry, Redshift's list no longer carries the AWS IAM fields, so a Redshift connection saved on a shipped build with `awsAuth` set would keep that flag (it lives in `additionalFields`) while `awsSecretAccessKey` was never loaded: the connect fails, and the AWS section no longer renders, so there is no way to turn it off. It now uses the superset, like the other four paths.

Two more were real fences that did not hold:

- `QueryCompletionProfileRegistry.clear(connectionId:)` removed the generations it had just bumped, so `default: 0` answered 0 again and a resolution still blocked in the metadata lease could commit a profile from the closed session into the emptied cache. The next connection's scope keys are identical, so it would have been served that profile. The generation is now monotonic across teardown, with a test.
- `SchemaProviderRegistry`'s failure path cleared `loadedScopes` unfenced, so a populate that lost the race and then failed could discard the flag for the union `SchemaRefreshService` had just written.

Also fixed: eviction no longer runs while a connection has no browse cursor (the window a reconnect passes through, where the browse provider is protected by neither the exemption nor `liveScopes`), `getOrCreate` no longer evicts the provider it just created, and `verify.sh` now recognises all three spellings xcodebuild uses for a failing case rather than only the Swift Testing one, which had made every XCTest-named quarantine entry read as a new failure.

Reported and not fixed: `SchemaProviderRegistry.populate` coalesces a refresh that arrives mid-fetch into the older in-flight fetch, so a table created during a tab's initial load will not appear in that tab's completions until the next refresh. Completion stays correctly scoped, and the fix needs a re-run mechanism with its own race surface, so it is better as its own change.

## Test plan

- `verify.sh build`: **PASS**.
- `verify.sh test` over 24 suites: **PASS**, 359 cases, 358 passed. The one failure is `SQLCompletionProviderTests/testCommaFromScopesColumnsToAllTables()`, already on line 37 of `.github/macos-test-quarantine.txt`.
- `verify.sh lint` over 14 changed paths: **0 violations**.
- `verify.sh docs`: **PASS**.
- `verify.sh abi`: see below.
- `verify.sh plugins`: could not run here. The pinned `oracle-nio` fork fails to build locally with `unknown attribute 'usableFromInlinenonisolated'` from a `@TaskLocal` macro expansion, in target `OracleNIO`, with no error in any file this branch touches. CI's `AllPlugins` step covers it.

20 new test cases across three suites:

- `PluginManagerVariantAccessorTests` pins the accessors rather than the registry, which is the end that was wrong. It includes a general guard asserting every accessor agrees with `snapshot(for:)` for every known type, so a future accessor written against `pluginTypeId` fails on the variants instead of shipping silently.
- `QueryCompletionProfileRegistryTests` covers failure-not-cached, recovery-is-cached, the fence on return as well as on commit, and per-scope revision isolation. One case previously asserted the opposite behaviour, in a test named "returns **and caches** the conservative base profile".
- `SchemaContextForAITests` covers the capitalised-table column mismatch, and that the context names the scope's database rather than the browse cursor.

## UI automation

The two-tabs-on-different-databases flow has no deterministic coverage. The only sample database is `Chinook.sqlite`, and SQLite declares `supportsDatabaseSwitching: false`, so the picker renders a static indicator and the schema half is nil: a test there would exercise none of this change. The flow needs a live multi-database MySQL or PostgreSQL server, which CI does not have.

## Docs

`docs/features/autocomplete.mdx` now describes the binding in terms of the control the user can see, rather than as a caching rule.


datlechin:
Reviewed this and pushed six commits fixing what the review turned up. Of the 15 findings, 5 were confirmed as written, 8 were partly confirmed with corrections, 1 was unprovable, and 1 was refuted. Two regressions nobody had named turned out to matter more than anything on that list.

## The two that mattered

**A variant type was completing against the app's stub dialect.** `registerVariant` keeps the curated built-in entry for a variant and took only `explainVariants` from the plugin. The curated `postgresqlDialect` is a placeholder for the window before any plugin loads: **84 keywords, 43 functions, 28 data types, no operators**. `PostgreSQLDialect.descriptor` ships **559 keywords, 90 data types, and an operator set**. Before this branch, `sqlDialect(for: .redshift)` read `snapshot(forTypeId: "PostgreSQL")` and got the plugin's. The new rawValue-first lookup fixed `caseSensitivityStyle` and in the same move cut Redshift, CockroachDB and PGlite completion down to the stub.

Fixed at `registerVariant`: a variant takes the plugin's editor config, and the curated entry keeps only the facts it states deliberately, meaning the ones where it differs from the curated *primary*. Today that is case-insensitive matching alone, which is why Redshift is spelled `postgresqlDialect.withCaseSensitivityStyle(.caseFoldFunction)`. It is a named single-fact adoption next to the existing `adoptCuratedCaseSensitivity`, not a value diff: `SQLDialectDescriptor` is not `Equatable`, and a whole-descriptor comparison would report "differs" for Redshift and hand it the stub straight back. The primary type id is now passed explicitly, because `registerTypeAlias` runs on the line *after* `registerVariant`, so `reverseTypeIndex` cannot be relied on there.

**The filter preview and the executed query disagreed on Redshift.** `metadataSnapshot(for:)` was wired into three readers, so `FilterCoordinator.generateFilterPreviewSQL` resolved Redshift's `.caseFoldFunction` while the base query still built through `resolveSQLDialect` → `snapshot(forTypeId: pluginTypeId)` and got PostgreSQL's `.ilikeOperator`. The preview showed one filter and the grid ran another.

Fixed by moving the lookup onto the registry as `snapshot(for: DatabaseType)` and routing every reader of those editor fields through it: `sqlDialect`, `statementCompletions`, `editorLanguage`, `autoLimitStyle`, `columnTypesByCategory`, `DialectQuoteHelper.resolveSQLDialect` and `SQLDialectFactory.createDialect`. It lives on the registry rather than in `PluginManager` because `resolveSQLDialect`'s callers are not `@MainActor`.

The remaining ~57 `pluginTypeId` metadata reads stay as they are, deliberately. The curated variant capability tables were written when nothing read them and are wrong in both directions, so converting them is a separate audit of the values themselves rather than of the lookup.

## The review findings

**Two writers on one provider.** `SchemaProviderRegistry.refresh` and `SchemaRefreshService.syncAutocompleteProvider` both repopulated the browse-scope provider on one `refreshData`, with different table sets: `loadSchema` writes the current schema, `resetForDatabase` writes the union across every expanded schema. The registry now excludes the browse scope, which `SchemaRefreshService` owns. It keeps its own subscription rather than handing population over, because `SchemaRefreshService` subscribes only to `currentSchemaChanged` and the broadcast route into it is gated on `reachesBrowsedDatabase`, so it has no trigger that reaches a non-browsed scope at all.

**Qualified completion off the browse scope.** `setNamespaces` had one caller, for the browse scope only, so a tab bound to another database had `knownSchemas == []`: `SCHEMA.` after FROM returned nothing and the FROM/JOIN namespace list was empty. `prepare` now fetches the object list and the schema list on one `.bulk` lease and sets namespaces, the database list and the connection.

**A refresh that landed during a load was dropped.** `loadSchema` coalesces onto an in-flight task and issues no new fetch, so the provider kept the pre-change table list with nothing scheduled to correct it. Population no longer goes through `loadSchema`.

**An empty catalog was refetched forever.** `prepare` gated on `isSchemaLoaded()`, which is `!tables.isEmpty`, so a database that legitimately has no tables never satisfied it and re-ran a catalog fetch on every activation of that tab. The registry tracks loaded scopes itself.

**Duplicate catalog fetches.** Two tabs on one scope both passed the guard, then serialised on the same pooled entry, so the second issued a full duplicate `fetchTables` after the first had finished. There is now an in-flight task per scope.

**Providers accumulated.** Keyed by scope but evicted per connection, so every database or schema a session visited left a provider holding up to 300 tables of column metadata until disconnect. Eviction now drops only scopes nothing renders, asking `CoordinatorLiveScopeProvider` at eviction time rather than counting retains: a query tab reads its provider from a SwiftUI body, and appearance is not lifetime here, so a retain keyed on the view would churn on every pane reparent.

**AI inline suggestions had no schema.** `connectionInfo` was written only by the quick switcher's `loadSchema`, which was skipped whenever the provider was already filled, so `buildSchemaContextForAI` returned nil and the prompt fell back to the schema-less template. Every scope now gets its connection. `resetForDatabase`'s `database` parameter was declared and never stored, which mattered once tab scopes existed: `buildSchemaContextForAI` named the database with `browseDatabaseName`, so it would have labelled one database's tables with another's name.

`AIChatInlineSource` also latched the first `SQLSchemaProvider` instance and never rebuilt it. That was harmless while one instance per connection was mutated in place; with an instance per scope it pins the first scope's tables, and it goes live exactly when `connectionInfo` starts being set. It is now refreshed per request, matching `connectionPolicy` beside it.

**The completion popup closed under the user.** `configure` rebuilt the service unconditionally, and the editor's `.task(id:)` folds in a profile revision that any window's refresh bumps, so a Cmd+R elsewhere dropped an open popup mid-keystroke. `configure` is now idempotent on provider identity, database type and profile revision. `initializeEditor` and the connection change also configured with `profile: nil`, so whichever ran after the resolver put the editor back on the app's own dialect with nothing scheduled to correct it; there is now one configure path holding the resolved profile, and it also runs on a scope change so a rebound tab stops completing against the previous database.

**`QueryCompletionProfileRegistry`.** `init()` is `#if DEBUG` like its sibling, which documents the rule in code. A fenced resolution returns the freshly computed base rather than the profile the refresh existed to replace, on the joining path as well as the originating one. `clear(connectionId:)` handles teardown, keeping `revisions` monotonic while the connection is open because the editor's `.task(id:)` keys on it.

**Unshipped PluginKit surface.** `tokenCasingPolicy` was read nowhere and the `@_disfavoredOverload` init preserved no existing mangled symbol, since the type ships here for the first time. Both are gone while that is still free; wave 2 adds casing back additively with a new overload, which is the pattern CLAUDE.md prescribes. `revision` earns its place as the identity `configure` compares. The ABI check against the merge base is **purely additive** (a new requirement with a default and a new non-frozen struct, no symbol removed), so no `currentPluginKitVersion` bump.

**Test isolation.** `registerTypeAlias` had no inverse, so the fake alias leaked into the process; there is now a `removeTypeAlias` and a `defer`. And because no plugin loads under XCTest, the existing guard test could only ever assert over curated data and passed identically with the dialect regression present. The variant tests now hand `registerVariant` a synthetic snapshot carrying a distinctive keyword, which is what actually catches it.

Finding 8 is refuted: the coordinator used to call `getOrCreate` in its init, so the "no provider, skip" guard was unreachable in the app, and the doubled lease count was a test-fixture artefact.

Also removed `plans/20260812-sql-completion-singlestore-worktrees/`, which looked like local working notes rather than something meant to ship; `plans/` is untracked on main. Say the word if you want it back.

## Verification

`build` PASS, `lint` 0 violations, docs clean, `abi` additive against the merge base, and 354 unit cases across the suites owning the changed types, all passing bar 3 already in `macos-test-quarantine.txt`.

Two things I could not verify here. `AllPlugins` fails on this machine inside the vendored `oracle-nio` `@TaskLocal` macro expansion, with 0 non-Oracle errors across 150 compiled plugin sources, which is a known local-toolchain issue; CI compiles it on a different toolchain. And no UI automation: the flows here need a live multi-database server, and the sample database is single-database SQLite, so no deterministic XCUITest can express a tab bound to a database other than the browsed one.

The UI test failures currently on this PR are also on `main` at the merge base (the Compare & Sync cases and the sample-database ones), so they are not from this branch.

One thing I deliberately left alone, worth a look separately: `MetadataConnectionPool` keys entries by `(scope, workload)` against `maxPerConnection = 6`, and per-scope completion wants both workloads. Tabs across several schemas plus the browse scope can reach that cap. The eager `fetchAllColumns` preload running only for the browse scope and the selected tab's scope would be the cheap mitigation.

PR #2100, opened 2026-08-13
fix(sidebar): keep the workspace rail steady when switching tables

tuananhht94:
## Environment

- macOS: 26.5.1
- TablePro: 0.64.0
- Databases: MySQL 8 and Redis (the bug is UI-level, not database-specific)

## Reproduction steps

1. Open two or more connections or databases so the workspace rail is visible.
2. Open a table and apply a filter or sort, so the current tab holds active work.
3. Click other tables in the sidebar.

## Expected

The rail stays in place while tables change.

## Actual

Two problems on every table switch:

1. The rail flickers: every store event rebuilt all rail cells with `reloadData()`, which blinked the selection even when the entries had not changed.
2. The rail disappears for about half a second, then slides back in. With active work on the current tab, each table click opens a new native window tab. The new window's rail starts at width 0, and its first reload runs while the sidebar view loads, before the window has assigned the entry-count callback. Visibility is only applied at `viewWillAppear`, at which point the window exists, so the width animates in. The first frames of the new tab paint without the rail.

## Fix

- Diff the old and new entries before reloading: unchanged entries touch nothing, in-place changes reconfigure the existing cells, and `reloadData()` runs only when the set of workspaces itself changed.
- Apply rail visibility synchronously in `viewDidLoad` after the split items are added. The window is nil at that point, so the width lands unanimated before the first frame.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The workspace rail no longer flickers when you switch tables or focus another window, and no longer disappears for a moment when a table opens in a new window tab.
 - Right-clicking a table that is not part of the current selection now acts on that table. It used to act on the selected tables instead, including for Delete.
 - Database icons and the current-database checkmark stay visible on a selected row in the database switcher. They were drawn in the accent colour, which vanished against the accent-coloured selection.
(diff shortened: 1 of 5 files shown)


sophiathedev (commented):
please add a short before/after video preview.

datlechin:
Hello @tuananhht94, please try run the app from latest main commit, see the issue fixed or not?

PR #2112, opened 2026-08-14
feat(explain): visualize MySQL tree plans

sophiathedev:
## Summary

- parse MySQL `EXPLAIN FORMAT=TREE` and `EXPLAIN ANALYZE` output into the existing diagram and tree viewer
- preserve MySQL/MariaDB JSON parsing and Raw fallback for unsupported, malformed, oversized, or excessively deep plans
- surface estimated cost and rows, actual timing, row and loop counts, relation names, and index details
- support trackpad pinch-to-zoom across all parsed EXPLAIN diagrams while retaining the existing zoom controls
- document the supported MySQL variants and include a full-window preview

## Why

Typed MySQL `EXPLAIN ANALYZE` statements were routed to the viewer by #1488, but remained Raw-only because the MySQL parser accepted JSON plans only.

Follow-up to #1488.

## Safety

- does not add a new SQL execution path or toolbar action
- limits TREE input to 2 MB, 10,000 nodes, and 128 levels before falling back to Raw
- rejects invalid or overflowing numeric metrics without trapping
- clamps button and trackpad zoom to 25%–300% and ignores invalid gesture values
- leaves MariaDB ANALYZE behavior unchanged

## Verification

- 24 focused app tests across `MySQLPlanParserTests`, `CockroachDBPlanParserTests`, `ExplainResultRouterTests`, and `QueryPlanDiagramZoomTests`
- 181 tests in 27 suites with `swift test --package-path Packages/TableProCore`
- `swiftlint lint --strict` with 0 violations
- Debug `TablePro` build with Xcode 26
- manual end-to-end validation through TablePro's MySQL driver against MySQL 8.4.11 using a multi-join `EXPLAIN ANALYZE` query

The full app test suite was also attempted, but an existing `MCPBridgeIntegrationTests` local-server test did not complete. SwiftFormat 0.62.1 stops on the repository's unsupported `--ifdefindent` option before formatting.

## Preview

MySQL EXPLAIN ANALYZE diagram


datlechin:
Thanks @sophiathedev. This work has shipped in #2117, which is now merged.

Your two commits are included there unchanged, with your authorship intact:

- `8fb76e6` feat(explain): visualize MySQL tree plans
- `651fdb1` feat(explain): add trackpad zoom

#2117 started as a review of this PR and grew into a rewrite of the whole feature, so rather than ask you to rebase onto a moving target we carried your commits forward. The MySQL TREE and ANALYZE parsing you added is what the docs screenshot on the EXPLAIN page now shows.

What changed around your work:

- The plan format moved onto `ExplainVariant` in PluginKit, so a parser is chosen by output format instead of a hardcoded switch on `DatabaseType`. MySQL's two variants are tagged `mysqlComposite` and still resolve to your composite JSON-then-TREE parser.
- The three EXPLAIN execution paths collapsed into one gated entry point. That closed a Safe Mode bypass and made `EXPLAIN ANALYZE` cancellable.
- Pinch zoom moved onto `NSScrollView.allowsMagnification`, so the diagram gets anchored zoom, smart magnify and Cmd with scroll from AppKit. The zoom bounds you factored into `DiagramZoom` are still the source of truth for both diagrams.

A few things from the review of this PR were fixed on the way in, in case they are useful next time: cost never rendered because every display site required both a startup and a total cost and MySQL only reports the total; the diagram derived each row's Y from that node's own height, so siblings drifted apart; and the parser dropped any continuation line that was not a metrics fragment, which lost node text when a query literal contained a newline.

Closing this one in favour of #2117. Thanks again for the contribution.

Issue #2157, opened 2026-08-17
UUID still displayed as Hex


DeckardCain2014:
### What happened?

In https://github.com/TableProApp/TablePro/issues/2086 the UUID display should have been fixed. But my table is still shown in Hex values instead of UUID.

<img width="326" height="301" alt="Image" src="https://github.com/user-attachments/assets/a7fc2c07-eec8-439c-8884-871a589c3ed2" />

<img width="584" height="236" alt="Image" src="https://github.com/user-attachments/assets/baaac4e2-c8a9-414b-9a40-a6cc7dc179d7" />

<img width="432" height="294" alt="Image" src="https://github.com/user-attachments/assets/0b3fc1f1-c338-417c-8f81-32aa72e601e0" />

### Database type

None

PR #2159, opened 2026-08-17
fix(datagrid): honor binary UUID display

sophiathedev:
## Summary

- honor `Display As > UUID` for generic 16-byte SQL binary cells in the grid,
  row inspector, VoiceOver, and single-cell copy
- persist an explicit `Raw Value` choice and isolate formats by duplicate column
  occurrence and pinned result set
- keep mounted cells, the inspector, and active value filters consistent when a
  display format changes
- leave undecoded MongoDB binary values in hex so BSON subtype and legacy byte
  order remain owned by the MongoDB driver

## Root cause

The typed binary formatting path returned compact hex before consulting the
column's display format. The same early return also affected direct cell copy.
The surrounding display state was keyed only by column label and tab, so format
changes could leave mounted cells, filters, the inspector, or pinned results
using stale state.

## MongoDB boundary

Issue #2157 led to this investigation, but MongoDB subtype 3 values cannot be
safely interpreted without their configured Java, C#, or Python legacy byte
order. This change does not guess that order. If the driver has not decoded a
MongoDB binary UUID, the host keeps it as hex and the connection's **Legacy UUID
Encoding** setting remains authoritative.

Related to #2157.

## Preview

The preview uses an isolated SQLite fixture with two generic 16-byte BLOBs.

### Before

Binary cells use compact hexadecimal output.

Binary UUID values before formatting

### After

`Display As > UUID` renders the same stored bytes as lowercase UUIDs.

Binary UUID values after formatting

## Testing

- focused formatter, storage, grid cache, live reload, copy, filter,
  accessibility, inspector-trigger, and MongoDB codec/flattener suites
- neighboring result-pinning, layout, width, keyboard-copy, display-mapping,
  interaction, filter-state, and tab-switch suites
- strict SwiftLint on all 25 changed Swift files with zero violations
- unsigned TablePro Debug build
- `git diff --check`
- final independent AppKit/state and plugin-boundary reviews with no findings


datlechin:
I pushed a commit to this branch (maintainer edits are on) with fixes from a review pass. Happy to drop any of it if you disagree with a call.

### Fixed

- `DataGridView` assigned `coordinator.delegate` one line *after* `syncDisplayFormats`, so the new `dataGridDisplayFormatChanged()` was dropped whenever the delegate was nil, which is exactly the state `releaseData()` leaves behind. Moved the assignment above the sync in both `makeNSView` and `updateNSView`.
- `CellDisplayFormatter`'s `.text` branch applied any non-`.raw` format without the `isApplicable` check the `.bytes` branch gained. That left the MongoDB carve-out open for a driver-decoded hex string arriving as `.text` on a blob column.
- `remapValueFilters` called `valueFilterState.prune` but derived its result only from the loop below it, so a prune that cleared every filter left `valueFilteredIDs` stale with no filter showing. `prune` now reports whether it removed anything and that feeds `didRemap`.
- `remapValueFilters` also set `didRemap = true` for any filtered column whose format changed, even when no selected value actually moved. Callers then cleared the cell selection and ran a full reload for nothing. It now skips the write when `selectedValues` is unchanged.
- `setDisplayFormat` built an array of length `columnIndex + 1` while the SwiftUI recompute builds a full `columns.count` array, so `displayFormatsChanged` was true on the next update and one menu click cost two full `reloadData` passes. Extracted `DisplayFormatArray.setting` to pad to the column count.
- The inspector read `columnDisplayFormats` from the window's single `dataTabDelegate`, which can still point at the previously selected tab. Added `InspectorValueDisplayFormatResolver.activeFormats`, which hands back the grid's formats only while it still holds this tab's columns, and otherwise falls through to the stored format.
- `effectiveFormat` was evaluated for every column and then discarded whenever `activeFormats` was non-nil, costing a `UserDefaults` read plus a JSON decode per column per inspector refresh. It is now computed only on the fallback path.
- `storageKeys` used `Dictionary(grouping:)`, which materializes an array per distinct label just to count occurrences. Swapped for a `reduce(into:)`.

Docs for the MongoDB carve-out, the persisted Raw Value, and the single-cell copy rule went into `json-viewer.mdx` and `data-grid.mdx`. Seven test cases added.

### Left alone, on purpose

- Persisting `.raw` rather than removing the entry. That is the point of the change, and it fixes Raw Value handing the column straight back to detection.
- Single-cell copy applying the format while a block copy does not. `copyKeepsRawBytesForMultiCellGridSelection` locks that in and the split is defensible. Worth flagging that it also means a 1x1 Cmd+C on a Unix timestamp column now copies a formatted date where it used to copy the integer.
- The format not persisting on a query tab with no table scope. That `if let scope = tableScope` guard predates this PR.

Verified on this branch in a separate worktree: project generation, 8 test suites (86 cases, including the new ones), a Debug build, and `swiftlint --strict` on the changed files.

PR #2178, opened 2026-08-18
fix(sidebar): show returned object kinds

### LINKED ISSUE #2173: Flat sidebar hides object kinds the driver returned, tree layout shows them
The sidebar's two layouts now disagree about which object kinds exist for the same connection.

PR #2151 changed the tree layout to list whatever the driver returned. The flat layout was left on the old rule, which hides an object kind whose capability flag the plugin did not declare. Toggling View > Sidebar Layout on such a connection changes the set of object kinds on screen.

## Where the two rules live

Flat layout still gates on the capability flag, in `SidebarViewModel.sectionShouldRender` via `SidebarObjectKind.shouldRender(itemCount:capabilities:)`:

```swift
func shouldRender(itemCount: Int, capabilities: PluginCapabilities) -> Bool {
    if self == .table { return true }
    if let capabilityFlag, !capabilities.contains(capabilityFlag) { return false }
    return itemCount > 0
}
```

Tree layout counts contents only, in `DatabaseTreeObjectGroupResolver.groups`:

```swift
guard itemCounts[kind, default: 0] > 0 else { return nil }
```

The gate only ever suppresses. It never causes an empty section to appear, so for a kind with items it is the difference between showing them and hiding them.

## Why this hides real objects

Two paths reach it.

A driver can return `MATERIALIZED VIEW` or `FOREIGN TABLE` rows from `fetchTables` while its plugin omits `.materializedViews` or `.foreignTables`. `DatabaseType` is a string-backed struct and the plugin domain is open, so a registry driver only has to forget the flag. The flat list then drops those tables with no section, no `No items` row and no error.

`SidebarViewModel.capabilities(for:)` also returns `[]` whenever the driver is not a `PluginDriverAdapter`:

```swift
guard let adapter = DatabaseManager.shared.driver(for: connectionId) as? PluginDriverAdapter else {
    return []
}
```

Every gated kind disappears from the flat list on that path.

## Steps to reproduce

1. Use a connection whose driver returns views, materialized views, foreign tables, procedures or functions that its plugin does not declare a capability for.
2. Browse it in tree layout. The folders appear.
3. Switch to View > Sidebar Layout > Flat. The sections are gone.

## Suggested fix

Apply the tree's rule to both: a kind with items always renders, and the capability flag stops deciding whether existing objects are visible. That makes `capabilities` unused in `shouldRender`, so removing the parameter is part of the change and touches `SidebarViewModel.sectionShouldRender` and its callers.

Worth deciding at the same time whether an empty section for a declared kind should render in flat layout. Today it does not, and that is the only thing the flag could still usefully express.

Tests: `SidebarViewModelTests` pins the current flat behavior in four cases, so those change with the fix. `DatabaseTreeNodeTests.objectGroupResolution` already pins the tree rule.

Found while reviewing #2151. The tree side is fixed there; this is the half that was deliberately left alone because it predates that PR.

sophiathedev:
> [!NOTE]
> Maintainer PR #2177 was opened while this draft's preview builds were running. Both pull requests address #2173. This pull request remains a draft until the overlap is resolved.

## Summary

- show every nonempty object-kind section in Flat sidebar layout
- remove capability metadata from section visibility decisions
- keep the empty Tables section while continuing to hide empty optional sections
- update sidebar tests, the changelog, and database navigation documentation

## Root cause

Flat layout required both returned items and the matching plugin capability before rendering Materialized Views, Foreign Tables, Procedures, or Functions. Tree layout already used returned item counts as its source of truth.

That mismatch hid real objects when a plugin omitted a capability flag. It also hid every gated kind while `DatabaseManager` temporarily had no active driver, because that path resolved capabilities to an empty set while cached schema objects remained visible.

Closes #2173

## Preview

Both screenshots use the same isolated fixture. The driver returns `RevenueSummary` as a materialized view without declaring the materialized-view capability, then the sidebar filters for `Revenue`.

### Before

Flat sidebar before the fix, with the returned materialized view missing

### After

Flat sidebar after the fix, showing the returned materialized view

## Verification

- [x] 176 focused sidebar, filtering, outline, and type-select tests
- [x] TablePro Debug build
- [x] strict SwiftLint on changed production Swift files, 0 violations
- [x] changed test code linted with existing file-wide baseline rules excluded, 0 additional violations
- [x] `git diff --check`

## Test notes

A deterministic XCUITest is not included because the checked-in SQLite UI fixture does not produce an optional object kind with missing capability metadata. The regression is covered at the visibility boundary for all six object kinds, while existing native outline tests cover Flat and Tree hierarchy behavior.

The verification wrapper's documentation add-on reports an existing `upstream/main` reference to `/Applications/Xcode-beta.app`, which is not installed on this host. SwiftLint itself completed with zero violations on the changed production files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Flat sidebar layouts show every object kind returned by the driver, matching Tree layout even when plugin capability metadata omits a kind. (#2173)
 - Sixteen-byte SQL binary columns respect their Display As setting in the grid, the row inspector, VoiceOver, and single-cell copy. Raw Value stays selected after refresh, filtered rows keep their identity when the format changes, and MongoDB binary UUIDs remain under the driver's Legacy UUID Encoding setting. (#2157)
 - Viewing a very long single line with word wrap on could use enormous amounts of memory and stall or kill the app. This hit the JSON value viewer, chat code blocks and the SQL review sheet, which always wrap, and the SQL editor when Word Wrap is on. Wrapped text is now laid out once instead of once per wrapped row, so a long line stays fast no matter how long it is.
 - An open cell overlay in the data grid kept its old border and background colour if the appearance changed while it was on screen, such as an automatic switch to dark at sunset. It repaints now.
--- a/TablePro/Models/Sidebar/SidebarObjectKind.swift
+++ b/TablePro/Models/Sidebar/SidebarObjectKind.swift
@@ -1,5 +1,4 @@
 import Foundation
-import TableProPluginKit
 
 struct DatabaseTreeObjectGroup: Hashable, Sendable {
     let database: String
@@ -56,16 +55,6 @@ enum SidebarObjectKind: String, CaseIterable, Sendable, Hashable {
         }
     }
 
-    var capabilityFlag: PluginCapabilities? {
-        switch self {
-        case .table, .view:     return nil
-        case .materializedView: return .materializedViews
-        case .foreignTable:     return .foreignTables
-        case .procedure:        return .storedProcedures
-        case .function:         return .userFunctions
-        }
-    }
-
     var isRoutine: Bool {
         self == .procedure || self == .function
     }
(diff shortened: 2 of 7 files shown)


sophiathedev:
Closing because #2177 landed the same fix while this draft’s preview builds were running. The implementation and issue are now resolved on main.

PR #2204, opened 2026-08-18
fix(json): stabilize tree filtering

sophiathedev:
## summary

- preserve JSON tree node identity while filtering so disclosure state targets the rows that are actually rendered
- build one filtered projection per search or root update instead of recursively rebuilding it during SwiftUI redraws
- reveal matching ancestor paths during search, then restore the user's prior disclosure state when the search is cleared
- reset cached projection state when result JSON changes and add accessible labels to the expand and collapse controls

## root cause

Filtering rebuilt every matching node with a new UUID each time the view evaluated the projection. Expansion IDs therefore referred to discarded rows, which could leave nested matches hidden. The projection was also recomputed from multiple view properties, repeating traversal work on large JSON values.

## verification

- [x] project generation
- [x] 234 focused regression tests across tree state, parsing, results JSON, number normalization, row conversion, reindent, content detection, and cell interaction
- [x] TablePro Debug build with code signing disabled
- [x] strict SwiftLint on all changed Swift files
- [x] native AppKit full-window render
- [x] near-limit regression with 4,900 nodes

A release-optimized local probe over a 4,900-node tree measured 8.19 to 8.34 ms per search update across 200 iterations.

## preview

Searching for a nested value reveals only its matching path. The full-window screenshot below was rendered with the production tree view.

JSON tree filter preview


datlechin:
Thanks for this. I pushed a follow-up commit (`db7d69d4b`) onto the branch with fixes from a full review, keeping your commit as-is. Here is what changed and why.

## The core problem

`visibleNodes` is a pure derivation of `(rootNode, searchText)`, but it was stored in `@State` and synced with `.onChange`. `.onChange` runs after `body`, so the derived half was always one frame behind. That one mistake produced three of the findings, including the frame that rendered the discarded tree's rows.

`State(initialValue:)` also evaluates its argument on every view init and then discards it, so the projection ran once per init, was thrown away, and `.onChange` recomputed it. Net effect was roughly twice the work per keystroke plus a full recompute on every unrelated parent redraw.

The instinct behind the PR was right: I measured a full projection over a max-size 5,000-node tree at **6.1-6.5 ms**, so per-redraw recomputation genuinely is not affordable. Only the direction was inverted.

## What the follow-up changes

- The projection is memoized in a plain non-Observable class held in `@State` and read from `body`, the pattern `MarkdownView.swift` already ships. It computes at most once per `(rootNode, searchText)`, is never stale, and costs nothing on unrelated redraws. `TreeProjectionCacheTests` asserts the actual computation counts.
- Disclosure state holds only user intent and is keyed by `keyPath` rather than `id`. `id` is a fresh UUID on every parse, so expansion could never survive a re-parse no matter how the snapshot was handled.
- Both `.onChange` handlers driving the projection are gone.
- Expanding or collapsing during a search now sticks for the rest of that search and is dropped when the field clears. This follows the rule the sidebar already settled in a bug fix (`DatabaseTreeOutlineCoordinator.swift:649`), so **Expand All** during a filter deliberately applies to the filtered view only.
- The filter is now one implementation shared by the JSON and PHP trees. `PhpTreeView` was a verbatim copy of the pre-PR `JSONTreeView` and still had every defect this PR fixed for JSON.
- Matching goes through `SidebarNameFilter`, the matcher the rest of the app uses. It trims whitespace, so a query of one space no longer empties the tree.
- A container matched by its key keeps its full contents and stays expandable, so you can search for a key and then read what is under it.

## Two corrections to the review that prompted this

- `localizedStandardContains` is accent-insensitive but **not** width-insensitive. Measured on this toolchain: `"ＡＢＣ".localizedStandardContains("ABC")` is `false`, and so is the half-width katakana case. The CHANGELOG claims accent-insensitivity only.
- Preserving expansion across a tree rebuild cannot be done by preserving `id`, because the parser mints new UUIDs. Keying on `keyPath` is what actually fixes it.

## Two defects found while doing this

Both are fixed here because the first one makes the new empty state lie.

- Filtering a truncated document reported no results for keys that are in the document. Both parsers also shorten `displayValue` (JSON at 300 characters, PHP at 80) while the filter matched only that, so any longer string was unsearchable past the cap. Matching now uses the whole value, and a truncated document says so instead of claiming the value does not contain what you typed.
- Copy Value on an object or array copied the summary label `{3 keys}`. JSON containers now copy that part of the document, and long strings copy in full.

## Verification

Build passes. 31 tests across 4 suites pass. SwiftLint reports 0 violations on the changed paths.

No UI automation: this flow needs a live connection with a JSON column and an open cell popover, and the UI test target has no fixture for that.

One known limitation worth recording: expansion is keyed by `keyPath`, so two containers that resolve to the same path (a key containing a dot, as in `{"a.b": {}, "a": {"b": {}}}`) would open together. It is rare, and it is the trade that lets expansion survive a re-parse.

PHP objects and arrays still copy their summary label, because the PHP serialized format cannot be rebuilt from the display tree. Leaves copy in full.

datlechin:
Also merged current `main` in (`a7902dfa2`) to clear the conflict. The only clash was `CHANGELOG.md`, where both sides had added sections under `[Unreleased]`; they are merged into the Keep a Changelog section order. No source file conflicted. Build, tests and lint were re-run after the merge.

Issue #2107, opened 2026-08-13
Give Starter value that does not depend on sync


datlechin:
### Problem

Everything behind the Starter paywall today does the same thing: move data between machines.

| Starter feature | What it actually is |
|---|---|
| iCloud Sync | multi-device |
| 2 Mac activations | multi-device |
| Encrypted connection export | sharing |
| Environment variables in connection fields | sharing |
| Linked Folders | sharing |

So a developer who works on one Mac, and who does not want connection metadata leaving that Mac, has nothing to buy. Free already covers every supported database, the SQL editor, the AI assistant, the MCP server, SSH tunnels, ER diagrams, EXPLAIN visualization, the server dashboard, Users & Roles, backup and restore, Vim mode, query history with full-text search, Safe Mode with Touch ID, and XLSX export, on unlimited Macs.

This came out of user feedback:

> I think you need to revise your pricing strategy, like mole.fit, with a basic lifetime pricing for the individual version. This can ensure revenue from the basic version, while also adding some features for individuals. As for the professional version, honestly, no one would feel comfortable storing database account-related information on a third-party cloud for synchronization.

Two separate points in there, and only one is a product gap:

1. **Lifetime pricing already exists.** Starter is $24/year or $59 one time, and the app has supported it since day one (a license with no `expiresAt` shows as Lifetime in Settings > Account). Someone who read the pricing page closely enough to send feedback still missed it. That is a pricing page problem, not a pricing problem. Tracked as a side item below.
2. **"Adding some features for individuals" is the real gap.** For a user who declines sync, Starter is worth $0 at any price. Lowering the price does not sell it. Changing what is in it does.

Goal of this issue: give Starter value that a single-Mac user wants, without taking anything out of Free.

### Proposed solution

Five candidates. All run entirely on the local machine, need no account and no sync, and build on layers that already exist. None of them exist in the codebase today.

**Ground rule: nothing currently free moves behind the paywall.** Repackaging free features as paid costs more trust than it earns revenue. Every item here is new work.

---

- [ ] **1. Schema Diff and migration SQL**

  Compare the structure of two databases (local against staging, or two schemas on one server) and show the differences by table, column, index, foreign key and trigger. Then generate the SQL that brings one side to the other.

  Why it sells: people search for this by name. Redgate SQL Compare is an entire product built on it, and DataGrip charges for the same thing. It is also the clearest answer to the feedback above, because nothing leaves the machine.

[...]

Issue #1837, opened 2026-07-09
Faulty Details for Json column


likeme:
### What happened?

When have json column, even though data is present in table format, on select of row in right side fields level you can see data is missing. Like in shared image, delivered timestamp not visible. It led to misleading and removal of software to better and reliable one.

<img width="1509" height="782" alt="Image" src="https://github.com/user-attachments/assets/15947355-d81c-4b85-a9ad-30c85565fd45" />

### Database type

MySQL / MariaDB

PR #2222, opened 2026-08-19
feat(results): add native result charts

sophiathedev:
## Summary

- add a native Chart result mode with bar, line, area, and scatter views
- provide typed X/Y and optional series controls over the active loaded result buffer
- show exact raw X, Y, and series values on hover without changing chart scope
- adapt chart controls to narrow result panes and preserve readable plot space
- bound projection work and reject values Swift Charts cannot represent without precision loss
- keep data-grid-only commands and licensed metadata outside Chart mode

## Preview

<table>
  <tr>
    <td><strong>Bar</strong><br>Bar chart full-window preview</td>
    <td><strong>Line</strong><br>Line chart full-window preview</td>
  </tr>
  <tr>
    <td><strong>Area</strong><br>Area chart full-window preview</td>
    <td><strong>Scatter</strong><br>Scatter chart full-window preview</td>
  </tr>
</table>

All previews are full-window captures using the bundled Chinook sample database. The query charts invoice date, revenue, and billing country from 40 loaded rows.

## Scope

Charts visualize loaded rows only. They do not fetch, aggregate, sample, or persist configuration. Pivot tables remain follow-up work.

Related to #2107. This implements the result-chart slice of item 4; pivot tables remain follow-up work.

## Validation

- generated `TablePro.xcodeproj` from `project.yml`
- passed chart, selection, toolbar, projection, license, result-mode, status-bar, and pinning regression suites
- rendered categorical and numeric X data across all four chart types in light and dark appearances
- passed strict SwiftLint on all 13 changed Swift files
- verified the string catalog and chart localization placeholders for Turkish, Vietnamese, Simplified Chinese, and Traditional Chinese
- passed an unsigned Debug TablePro build
- manually verified all four chart types in full-window Chinook flows

The chart UI test now uses stable accessibility identifiers for the Chart segment and license gate. It compiles locally; the local UI runner was killed before bootstrapping, so GitHub Actions remains the authoritative runtime result for that test.


datlechin:
Pushed `eb5a9d3` onto this branch: it fixes the 15 review findings plus 4 more that came out of investigating them. Nothing was rewritten, it is one commit on top of your work.

Three root causes account for most of it, so the fix is a refactor at three boundaries rather than 19 patches.

**1. The chart's vocabulary was narrower than Swift Charts'.** `AxisKind` had two cases where Charts has three primitive plottables (`String`, `Double`, `Date`).

- `.date` / `.timestamp` / `.datetime` now resolve to a real time axis instead of unordered string bands. Cell text is parsed by a new `DatabaseDateParser`, extracted from `DateFormattingService` so there is one grammar, plus three measured patterns it was missing (PostgreSQL `timestamptz` with a space before the offset, and microsecond timestamps).
- `y` and numeric `x` are now `Double` rather than `Decimal`. This was the worst finding and it was not in the original review: `NSDecimalNumber(decimal:).doubleValue` is not correctly rounded, so `isExactlyPlottable` was rejecting **2806 of the 10000 values of the form `d.dd`** (measured, reproduced twice). A `DECIMAL(10,2)` price column silently lost over a quarter of its bars: `19.99`, `0.07`, `1.07`, `0.11`, `0.21` were all dropped. Charts converts to `Double` anyway, so parsing with `strtod` and validating against the value Charts will actually plot removes the second, differently-rounded conversion entirely.
- Bar grouping passes a discrete slot to `.position(by:)` instead of an `Int`. Charts reads an `Int` as a **continuous offset**, so a repeated category's second bar rendered inside the *next* category's band at a different width. Measured with an `ImageRenderer` pixel harness: for `(paid,100),(paid,60),(pending,30)` the second `paid` bar centred at x=398 with the band boundary at x=363. The two-series grouped bar chart in this PR's own preview was affected. Also measured that Charts orders a discrete position scale by first appearance, not alphabetically, so the ordinal's string needs no zero padding.
- `Resolved.xAxisKind` is now carried on the projection and used, instead of the canvas sniffing `points.first`.

**2. A bound was treated as a terminal error.** Every cap returned `points: []`, so a result one row over any limit showed an error card instead of a chart.

- Caps now truncate and report: `limits: [Limit]` replaces the all-or-nothing `Issue`, and the toolbar says "Showing the first 2,000 points of 8,431 loaded rows". Every other tool does this (Metabase's "Showing first 2000 rows", DBeaver's row-count cap); none refuses.
- `showsDataChrome` split into per-affordance predicates on `ResultsViewMode`, so Chart mode keeps the row count, `Fetch All` and the pagination controls. The toolbar's "More rows available" warning is gone: it had no remedy on screen, and `hasUnloadedRows` was forward-looking only, so on the **last** page of a 5-page table it reported the chart complete while 4,000 of 5,000 rows were missing. The status bar's row range already tells the truth, so the second, wrong signal was removed rather than corrected.
- A skipped row now records a line break on every path. When the series cell is the unreadable one the row cannot be attributed, so every series breaks.
- `ResultChartView` has an explicit state enum, so no combination renders a blank pane. `project()` is `async throws(CancellationError)`, which makes "cancellation is the only failure" a compiler guarantee rather than a comment, so no unreachable retry UI was added.

**3. A three-way router was collapsed into a bool.** `GridSelectionOwner` already existed; `dataGridOwnsSelection` flattened it.

- Every row command switches over the owner exhaustively. This fences `resolvedRowSelection()` itself, so a stale grid selection carried into Chart mode can no longer delete rows, while the sidebar's table-deletion path still works.
- Structure mode gets Copy back. `Copy Rows with Headers` and `Copy Rows as JSON` are gated separately on `hasDataGridRowSelection`, because the structure grid has no handler for them and feeding it structure-grid indices is the display-position bug from #1837.
- Ownership only counts where a handler exists, so a Create Table tab no longer swallows `Cmd+Delete` and `Cmd+C` into nil handlers.
- JSON mode keeps its row commands. That change was unrelated to charts and undocumented, so it was reverted rather than kept.
- Chart configuration moved from `ResultSet` to `QueryTab`, beside `sortState` and `pagination`, and is keyed by column **name** rather than index. It now survives a page turn, a sort and a re-run, and a re-ordered `SELECT` cannot silently chart a different column. A choice whose column is missing is kept, not erased.
- The default Y axis skips the primary key. On the bundled Chinook `Track` table the first chart was Row Number vs `TrackId`: a straight diagonal.

**Smaller ones:** per-series dashes come from `.chartLineStyleScale(range:)` (the constant `.lineStyle` was overriding `.lineStyle(by:)`, so series differed by colour alone); hover uses an index built once instead of two O(n) scans per frame, and the deep `onChange(of: projection)` comparison is gone; the accessibility summary no longer reads "with 1 points"; `docs/features/licensing.mdx` gained the tier row.

**On finding 2** (the segment accessibility identifier): it does not reproduce. A measured AX probe shows a segmented `Picker` passes a per-segment identifier through to the `AXSegment`; the container does not clobber it. The UI test now addresses the segment by its title anyway, matching the two CI-green precedents in `QueryPlanResultUITests` and `OpenQuicklyCommandUITests`, and it passes locally: the runner logs `Click "chart.xyaxis.line" RadioButton`.

**Verified:** Debug build passes. 160 unit tests pass across every touched suite, including new ones for the money-value regression, the chronological date axis, cap truncation, config survival across a result replacement, the default-Y rule, the bar banding, and the plural. `ResultChartUITests` passes. `swiftlint --strict` is clean on every changed file.

One pre-existing lint error is untouched and not from this branch: `ExecutionAuditLog.swift:40` (`storage_environment_directory`), which is already on `main` at the merge base.

datlechin (approved):

PR #2264, opened 2026-08-20
fix(localization): standardize Korean UI copy

devy1540:
## 목적
issue: #2219 
Follow up on the maintainer feedback from #2237 by making Korean UI copy consistent across macOS and iOS.

## 내용(의도 포함)

- Use a consistent formal register for Korean instructions and confirmation prompts.
- Align common commands with macOS terminology and correct numeric particle handling.
- Add regression tests for the agreed register and platform terms.
- Leave behavior, source strings, and other localizations unchanged.

## 성공기준

- `scripts/localization.py verify` passes.
- Both edited catalogs compile with `xcstringstool`.
- Strict SwiftLint passes.
- `KoreanLocalizationSourceTests` passes on macOS.
- The iOS app builds for the iOS 26.5 simulator.
- The Korean language option, formal help text, and updated Edit menu terms render correctly in the macOS app.


datlechin (approved):
Thanks, i'll continue follow up this pr

PR #2267, opened 2026-08-20
fix(plugin-dameng): reconnect after a stopped query instead of failing forever

### LINKED ISSUE #2262: dameng bugs
### What happened?

macOS26，达梦数据库连接尚存些许bug。切换模式的时候报错（Schema 切换失败），打开表需要很长时间有时候甚至打不开

### Steps to reproduce

1.连接达梦数据库
2.打开表

### Expected behavior

_No response_

### Database type

None

### TablePro version

_No response_

### macOS version & chip

_No response_

### Screenshots / Logs

_No response_

datlechin:
Fixes #2262.

## What was wrong

The Dameng driver's own documentation and its own code comment both promise this:

> Stopping a query closes its connection. DM8 has no out-of-band cancel request, so TablePro
> stops reading the reply **and reconnects on your next query**.
> (`docs/databases/dameng.mdx`, and the same claim at `DamengPlugin.swift:171`)

The reconnect was never written. Measured against a live DM8 8.1 server through the exact C ABI
the plugin calls, using the committed `libdameng_bridge.a`:

```
--- cancel a statement that IS in flight ---
  [thread] tp_dm_cancel at t+200ms
  slow query + cancel           19338 ms  FAILED   error: cancelled

--- is the connection still usable afterwards? ---
  next statement                    0 ms  FAILED   error: Dameng connection is closed
  statement after that              0 ms  FAILED   error: Dameng connection is closed
  ping                       tp_dm_ping -> false
```

`is_recoverable` in the bridge is false for `Cancelled`, `Timeout`, `Protocol`, `Io`,
`ConnectionFailed`, `AuthFailed` and `NotConnected`, so `tp_dm_execute` disposes the client and
`TpDmConnection.client` stays `None` for the life of the handle. Swift never learned:
`DamengConnection.execute` rethrew without releasing `rawConnection`, so `connectedPointer()`
kept handing out the stale pointer and every later call reported a closed connection.

Cancellation is routine flow control in TablePro, not an exceptional path.
`MainContentCoordinator.supersedeExecution` runs on every tab retarget, which is every time the
user clicks a second table, and it cancels the in-flight query, the table-load task and the
row-count task. So opening one table while another was still loading killed the connection, and
everything after it failed until `ConnectionHealthMonitor`'s 30 second ping noticed. That window
is the reporter's "opening a table takes a very long time, sometimes it cannot be opened at all",
and a schema switch inside it surfaces through `MainContentCoordinator+Navigation.swift:540` as
the reported **"Schema Switch Failed"** alert.

## The fix

The driver reconnects and runs the statement once more, which is the shape MySQL
(`MySQLPluginDriver.executeWithReconnect`), PostgreSQL (`LibPQDriverCore.executeWithReconnect`)
and Redis (`RedisPluginConnection.executeCommandSyncRetrying`) already ship. Dameng was the only
driver missing it.

- **`Native/DamengBridge`**: `TpDmError` carries a `connection_lost` flag, taken from the
  `is_recoverable` the bridge already computes at the failure site, exposed as
  `tp_dm_error_is_connection_lost` next to the existing `tp_dm_error_is_cancellation`. The
  message text cannot separate "the server rejected this SQL", which leaves the connection
  usable, from "the transport died". `transaction_operation` now reports through
  `set_driver_error` as well, so a stopped commit or ping is classified the same way a stopped
  query already was.
- **`DamengConnection`**: retires its handle when the bridge says the connection closed, so the
  Swift state is honest and the driver can tell a dead connection from a rejected statement.
  Cancellation is now scoped to the statement actually on the wire: `tp_dm_cancel` is
  connection-wide, so a cancellation aimed at a statement still queued behind another used to
  interrupt the wrong one, and a queued statement that was cancelled ran anyway because the next
  `tp_dm_execute` cleared the flag on its way in. A connect is fenced by an epoch so one that
  lands after a disconnect cannot revive a connection the caller already closed.
- **`DamengPluginDriver`**: `executeWithReconnect` restores the active schema and the text
  escaping mode. Only a read is sent again, using the same rule MySQL applies, because a write
  may already have been applied before its reply was abandoned. `switchSchema` recovers too,
  since `SET SCHEMA` is idempotent. However many statements find the connection dead, they share
  one rebuild rather than each starting its own, the shape `SQLSchemaProvider` uses. A connection
  lost inside a transaction is never rebuilt under it: the loss is reported, because the rebuilt
  connection never saw the `BEGIN` and the later commit would report success having written
  nothing.

## Verified

| Step | Result |
| --- | --- |
| `verify.sh generate` | PASS |
| `verify.sh build` (app) | PASS |
| `DamengDriver` + `DamengDriverTests` build | PASS |
| `cargo test` in `Native/DamengBridge` | 8 passed, 1 ignored (live) |
| Dameng suite against a live DM8 8.1 server | 24 of 25 pass, see below |
| `verify.sh lint Plugins/DamengDriverPlugin Plugins/DamengDriverPluginTests` | 0 violations |
| `verify.sh plugins` (AllPlugins) | Dameng compiled and signed; the aggregate fails locally on OracleNIO's `@TaskLocal` macro, which this branch does not touch |

`Skill(code-review)` at high effort found six real defects in the first draft of this change,
all fixed here and covered by tests: retiring a dead handle used to invalidate the rebuild that
was recovering it, concurrent rebuilds fought each other, `tp_dm_cancel` could read a handle
another thread was freeing, a transaction could be silently replaced, and a failed escaping probe
permanently downgraded the mode. `Skill(security-review)` found nothing.

The regression test is `testLiveDM8RecoversFromAStoppedStatement`. It was checked both ways
against a live server: with the retry removed it fails with the exact error the issue reports,
`DamengError(message: "The Dameng connection is closed.")`, and it passes with the fix in place.

The one failing test in the Dameng suite is `testLiveDM8Workflow`, which fails at "create parent"
with `-2004: 不支持的语句类型`. That is pre-existing: it fails identically on `main`, and it is a
real defect described below, not a regression from this change. It had never been seen because
the live tests need `TABLEPRO_DM8_INTEGRATION=1` and the environment does not reach the test
process through `xcodebuild`; they run under `xcrun xctest`.

There are no screenshots because nothing on screen changed. The user-visible difference is the
absence of an error, and staging it by hand means racing a cancel against a query, which is the
non-deterministic thing the regression test replaced.

## Found while investigating, not fixed here

- `generateCreateTableSQL` emits `CREATE TABLE` and `CREATE INDEX` as one statement string, and
  DM8 rejects a multi-statement batch with `-2004`. Creating a table with an index fails.
  Measured directly against the server.
- `ALL_COL_COMMENTS.OWNER` is the owning **user**, while every other `ALL_*` view TablePro reads
  uses the **schema**, so the join in `fetchColumns` and `fetchAllColumns` drops every column
  comment whenever a schema is not named after its owner. Measured: the same query returns rows
  for `OWNER = 'SYSDBA'` and none for `OWNER = 'APP'`.
- A result set larger than one inline response is silently truncated: 662 rows of a 20,000 row
  table, and 1,159 of 2,000 tables in the sidebar, with `is_truncated` still false.
- A result set containing a `CLOB` column comes back with the wrong number of columns and rows.
- Stopping a Dameng query still takes several seconds to take effect. There is a fix for this in
  the pinned Rust fork that measures 9.80s to 0.00s at the crate level, but it made no difference
  through the shipped static library and I could not explain why, so it is not in this PR.


ekovovo:
plz release

datlechin:
@miracleloveWT dm8 plugin is releasing, please wait https://github.com/TableProApp/TablePro/actions/runs/32329126061

PR #2310, opened 2026-08-20
feat(connections): add Edit and Delete swipe actions to connection row

subpop:
Add swipe actions to the rows in connection list to edit and delete connections.


datlechin:
Thanks for this. I kept the feature and pushed a commit that reshapes it, plus merged `main` in to clear the CHANGELOG conflict. Writing up what I found, because the interesting part is that I had to measure it rather than reason about it.

**The gesture is genuinely native, which I was not sure of going in.** I built a small SwiftUI probe against the macOS 14 target and looked at what `swipeActions` actually produces: SwiftUI's `List` is backed by a private `NSTableView` subclass whose delegate implements the public AppKit method `tableView(_:rowActionsForRow:edge:)` and hands back real `NSTableViewRowAction` objects. AppKit's own header calls that `a "swipe to delete" feature` and it has been there since 10.11. It also works through `DisclosureGroup`, so grouped connections get the actions too, and this list's `.listStyle(.inset)` is the style where the reveal works. So the idea is sound.

Three things changed:

**1. A full swipe could delete a connection outright.** `allowsFullSwipe` defaults to `true`, and Apple's docs say a full swipe performs the first-declared action, which here was Delete. I checked whether that default actually survives the trip into AppKit, and it does: `NSTableViewRowAction` has a private `allowsFullSwipe` property and SwiftUI forwards the value into it, measured as `1` by default and `0` when you pass `false`. So one over-enthusiastic swipe could reach the delete path with no button press. It now passes `allowsFullSwipe: false`.

**2. The buttons had no names.** A bare `Image(systemName:)` label produces `NSTableViewRowAction.title == ""`, which I confirmed by reading the returned actions directly. That empty string is what accessibility reads, so VoiceOver had nothing but the SF Symbol name to go on. `Label(String(localized: "Delete"), systemImage: "trash")` sets a real title and keeps the image, and AppKit's header actually asks for the image anyway ("Prefer using an image over text for the row action button"), so this is strictly better rather than a trade. It also brings the two strings into the localization catalog.

**3. The buttons are now shared with the context menu.** `WelcomeContextMenus.swift` already built an Edit and a Delete button for the same two commands, so there were two copies that could drift. Both now come from shared builders.

One thing I deliberately did not do: the actions are not added to Linked or Team Library rows. Those rows currently reuse the source connection's UUID as their list tag, so a row action on them resolves to the *local* connection instead. That is a pre-existing bug I am fixing separately, but it meant extending the swipe there would have been a one-gesture path to deleting the wrong thing.

I also folded in a fix for something this feature made worse. The delete confirmation was being presented before its "does this connection have saved queries" check finished, so it usually showed the shorter message and never warned that saved queries get deleted too. That was survivable when delete took a right-click and a menu; with a swipe it becomes routine. The check now completes before the alert appears.

Worth knowing for future work in here: this cannot get UI test coverage. XCUITest on macOS has no two-finger swipe, and `NSTableView.rowActionsVisible` throws if you set it to `true`, so there is no way to reveal the actions programmatically at all. The new unit tests cover the confirmation ordering, which is the part with real logic, and the gesture itself stays manual QA permanently. If you have a trackpad handy, the things worth eyeballing are that both buttons show their titles, that a long uninterrupted swipe does nothing, and that VoiceOver reads them by name.

subpop:
I debated whether or not to include labels on the buttons. I omitted them to keep the buttons smaller, but your point about accessibility makes a lot of sense. I knew full swiping would trigger delete; I figured the confirmation dialog was a reasonable enough guard against accidental deletion. Setting `allowsFullSwipe` to `false` seems totally solid. Thanks for accepting the change!

PR #2321, opened 2026-08-21
fix: remove rectangular quick switcher outline

shuvroroy:
## Summary

Remove the thin rectangular frame around the rounded Open Quickly panel while preserving its native macOS drop shadow.

## Root cause

The panel is a transparent, borderless `NSPanel` whose visible surface is rounded by SwiftUI's Liquid Glass shape. WindowServer derives the native window shadow from the root `NSHostingView`, however, and that root layer remained rectangular. The result was a faint square perimeter outside the rounded glass corners.

Disabling `NSPanel.hasShadow` removes the perimeter, but also removes the panel's native depth. The fix instead masks the hosting layer with the same continuous corner radius as `QuickSwitcherSurface`, allowing WindowServer to retain the native shadow and follow the visible rounded surface.

Screenshot 2026-08-21 at 1 54 06 PM

## Changes

- Apply `QuickSwitcherMetrics.cornerRadius`, `.continuous`, and `masksToBounds` to the panel's root hosting layer.
- Add a regression test that verifies the native shadow remains enabled and the root shadow mask matches the rounded surface.
- Remove the stale rectangular edge pixels from the light and dark Open Quickly documentation screenshots without changing their UI content, dimensions, or color data.
- Document the fix under `[Unreleased]` in `CHANGELOG.md`.

## Screenshots

| Light | Dark |
| --- | --- |
| Open Quickly in light appearance without the rectangular outline | Open Quickly in dark appearance without the rectangular outline |

## Verification

- Reproduced the square frame in a standalone AppKit/SwiftUI panel on macOS 26.6.2.
- Captured the panel before and after applying the hosting-layer mask: the rectangular frame disappears while the native shadow remains.
- The TablePro scheme builds successfully.
- `swiftlint lint --strict` passes with no violations.
- `QuickSwitcherPanelControllerTests` passes: 11 executed, 11 passed.
- `git diff --check` passes, and changed Swift lines stay within 120 characters.
- Both documentation assets decode as 1280 × 1070 RGBA PNGs with fully transparent outer corners.

## Checklist

- [x] Regression test added
- [x] Documentation screenshots updated
- [x] `CHANGELOG.md` updated under `[Unreleased]` with #2321
- [x] No user-facing strings added
- [x] Build, strict SwiftLint, and Quick Switcher panel tests pass


datlechin:
Thanks for this. I could build and run it here, so I reproduced the artifact and measured it before changing anything. I have pushed a commit to this branch.

## What the outline actually is

It is not the window shadow. I captured the panel window with `screencapture -o`, which omits the shadow entirely, and the artifact is still there:

| build | corner alpha outside the rounded surface |
| --- | --- |
| before | 5 to 11 across the full square window bounds |
| after | 0 |

So `glassEffect(.regular, in: shape)` tints the hosting view's whole bounds, past the shape it is handed. That faint square wash is the outline. The shadow was never involved, which also means the `hasShadow` assertion was not testing the reported symptom.

Three things I measured while narrowing it down:

- `clipShape(shape)` after `glassEffect`, in `QuickSwitcherSurface`, changes nothing. The glass backdrop is a layer beneath the SwiftUI render tree, so SwiftUI cannot clip it. This is why your AppKit layer mask is the right altitude and my first instinct, that it was clipping at the wrong level, was wrong.
- Removing the `GlassEffectContainer` changes nothing either.
- Forcing the pre-26 branch gives corner alpha 0 with no mask at all. The surface already clips itself there, so this is a macOS 26 defect only.

## What I changed

- Gated the mask to macOS 26, since below it the corners already measure 0 and the mask would only add the offscreen compositing pass that rounding a layer with sublayers forces. macOS 14 and 15 now render exactly as they do today.
- Moved the mask into a named `maskContentToSurfaceShape` with the rationale as a `///` doc comment. The repo asks for no `//` comments describing what code does, and the surrounding files use doc comments for the why.
- Made `surfaceCornerRadius` an init parameter instead of the panel reading `QuickSwitcherMetrics` for itself. The panel is generic over its content, so the shape has to travel with the content rather than be re-derived in two places.
- Rewrote the test to assert the radius the panel was handed, using 19 rather than the metric. The old assertion compared the layer against the same constant the production line assigned, so it passed for any radius, including one the surface no longer used.
- Recaptured both documentation screenshots from a build carrying the fix, at the same 1280x1070. The previous PNGs were the old screenshots with a rounded alpha mask applied: the pixel diff was alpha only, zero RGB change, confined to the corners. The shipped images measure corner alpha 9 to 17, so they were showing the bug, and the mask also clipped the real corner antialiasing.
- Reworded the CHANGELOG entry, which described the shadow explanation.

## One thing for you

The PR description has an em dash in the checklist line. The repo bans them everywhere including PR bodies, and ships the grep that catches it.

Everything builds, `swiftlint lint --strict` is clean, and all 11 tests in `QuickSwitcherPanelControllerTests` pass.

PR #2352, opened 2026-08-21
chore(scripts): give two orphaned checks a job, and clean up after the ones that leak

datlechin:
Four items from the CI audit, all in the "this exists but does nothing" category.

## Two checks that guarded real invariants and ran nowhere

`check-mongodb-filter-shapes.sh` and `audit-refactor-health.sh` are referenced by nothing. The audit's suggestion was to delete them, but running both showed they work: the first reports `all shapes parse`, the second's `--check` mode reports `drift gates passed` against `.github/duplicate-contract-baseline.txt`.

So they are not dead, they are unwired, and deleting a working check to tidy up is the wrong trade. Both join `repo-hygiene.yml`. Neither needs anything macOS-only, so they run on the free Linux runner in about a second each.

The workflow's paths filter is widened to match, because a check that only runs when `scripts/` changes would not have run when the thing it checks changes: `TablePro/**/*.swift`, `Packages/**/*.swift` and the duplicate-contract baseline are added.

## A gate branch nothing can reach

`check-pluginkit-abi.sh` carried:

```sh
if [ "${ABI_ACKNOWLEDGED_ADDITIVE:-}" = "1" ]; then
    ... "The PR carries the abi-additive label ... the gate passes." ...
```

Set by a workflow that no longer exists. Its failure message then told the reader to add an `abi-additive` label and re-run, and nothing reads that label either. Both were instructions to do something with no effect. The branch is gone and the message now describes the additive-versus-breaking decision the script is actually there to support, per CLAUDE.md.

The larger question of whether to restore the deleted CI gate stays open; this only removes the plumbing that pretends it is still there.

## Two scripts that leaked on failure

`create-dmg.sh` attaches a volume and writes a temp image in its `hdiutil` fallback, with no trap. A failure between the attach and the detach left both behind, and the next run then failed on a volume name that was already mounted. It has a trap now that detaches and removes only the temp image.

`build-freetds.sh` used a fixed `/tmp/freetds-build` root and created four more `mktemp -d` directories that nothing removed, so a run leaked five directories. It uses one private root with a trap covering all five.

That fixed path was also doing something useful: caching the downloaded tarball across runs. Moving to `mktemp -d` would have turned every run into a fresh download, so the tarball now lives in a named cache directory outside the per-run root, and its checksum is verified on every run rather than only after a download.

## Stale instructions that bypass the new baseline

#2349 made `download-libs.sh` verify `Libs/ios` against a committed baseline. Two scripts still printed the old manual publish recipe as their closing "next steps":

```
tar czf /tmp/tablepro-libs-ios-v1.tar.gz -C Libs/ios .
gh release upload libs-v1 ... --clobber
```

Following that would publish an archive whose baseline nobody regenerated, and every developer and CI job would fail on the next pull. Both now point at `publish-ios-libs.sh` and the commit that goes with it. There are no `gh release upload libs-v1` instructions left in `scripts/`.

## Verification

`actionlint` is clean across all six workflows and `shellcheck -x --severity=warning` clean across all 43 scripts. Both newly wired checks were run locally and pass. Every touched script parses.

https://claude.ai/code/session_013MEaba8K1HQcyDNeq5wEFk

### DIFF
--- a/.github/workflows/macos-tests.yml
+++ b/.github/workflows/macos-tests.yml
@@ -186,6 +186,12 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: scripts/download-libs.sh
 
+      # Compiles a C probe against Libs/libbson and parses every filter document the MongoDB
+      # query builder can emit. It guarded that invariant and ran nowhere; it needs clang and the
+      # vendored libraries, so this is the first job where it can run at all.
+      - name: Check the MongoDB filter shapes against the query builder
+        run: scripts/check-mongodb-filter-shapes.sh
+
       # Cargo and rustup need access outside Xcode's user-script sandbox. Build the
       # native bridge explicitly before compiling the driver target.
       - name: Build Dameng native bridge
--- a/.github/workflows/repo-hygiene.yml
+++ b/.github/workflows/repo-hygiene.yml
@@ -75,10 +75,9 @@ jobs:
       - name: Validate the registry update script
         run: python3 .github/scripts/test_update_registry.py
 
-      # Two checks that guarded real invariants and ran nowhere. Both are pure bash and grep, so
-      # they belong on the free Linux runner rather than in the macOS suite.
-      - name: Check the MongoDB filter shapes against the query builder
-        run: scripts/check-mongodb-filter-shapes.sh
-
+      # A check that guarded a real invariant and ran nowhere. Pure grep over Swift sources, so it
+      # belongs on the free Linux runner. The MongoDB filter-shape check is the other one that was
+      # orphaned, but it compiles a C probe against Libs/libbson, so it lives in the macOS build
+      # job in macos-tests.yml where the toolchain and the libraries already are.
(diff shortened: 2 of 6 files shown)


datlechin:
The hygiene job failed on `check-mongodb-filter-shapes.sh`:

```
missing /home/runner/work/TablePro/TablePro/Libs/libbson_x86_64.a (run scripts/download-libs.sh)
```

My error. I checked what tools each orphaned script needed with a grep that did not include `clang`, concluded both were pure bash and grep, and put both on the Linux runner. That one compiles a C probe and links `Libs/libbson_${ARCH}.a`, so it needs macOS, a toolchain and the vendored libraries.

It moves to the `build` job in `macos-tests.yml`, immediately after `Download static libraries`, which is the first point in CI where it can run at all. `audit-refactor-health.sh --check` stays on the Linux runner; it is grep over Swift sources and passed there.

Verified locally: the shape check passes with `Libs/` present, and the drift gates pass.

Issue #2342, opened 2026-08-21
SQLite: Clicking a table immediately after opening DB hangs on "Executing…"


nano9g:
### What happened?

Opening a SQLite database and clicking any table in the left sidebar will display "Executing…" forever. Stopping the operation, selecting another table, then clicking back displays the data immediately.

### Steps to reproduce

1. Open a SQLite database
2. Click any table
3. Note that "Executing…" is displayed forever in the toolbar
4. Click the ⊗ next to Executing…
5. Select a different table and observe that rows are displayed immediately
6. Click back to the original table and its rows load as well

### Expected behavior

Tables should load normally.

### TablePro version

0.66.0

### macOS version & chip

macOS 15.8 / Apple Silicon

### Screenshots / Logs

Let me know if you need any more information. I've tested this will multiple SQLite databases and observed the same behavior. The number of rows in the table that won't load initially also does not seem to matter; an empty table triggers the issue just like one with a high number of rows. Thanks!

datlechin:
Thanks for the detailed report, the steps made this much easier to chase.

I traced it as far as I can from here. The "Executing…" label and the ⊗ beside it are driven by one window-level flag, and there is exactly one way for them to stay up: the flag was raised when the query started and nothing lowered it. A fix for that is on the way: the indicator will read the execution registry directly instead of a separately stored flag, so it cannot outlive the work it describes, and Stop, ⌘. and the disconnect warning will follow the same value.

What I could not reproduce is the stall itself. I built 0.66.0 and current `main` and drove both against SQLite: a restored window with no tabs, a fresh connection from the welcome list, opening a `.sqlite` file into a running app, cold launching with the file, a 400-table database, and clicking a table with a blank SQL editor tab open. Every first click finished in 68 to 140 ms.

So I would like to know whether your query actually ran and never came back, or ran and had its result discarded. TablePro traces the whole path already. If you can capture it, that answers it in one line.

In Terminal, before you open the database:

```
log stream --level debug --predicate 'subsystem == "com.TablePro" AND category == "TableLoad"'
```

Then open the database, click a table, wait until "Executing…" is clearly stuck, and paste what the terminal printed. Each navigation prints one numbered block. The last stage in the block is the answer:

- ends at `executeStarted` with no `driverFetchBegin` after it: the query never reached SQLite
- ends at `driverFetchBegin` with no `driverFetchEnd`: SQLite has it and is not returning
[...]

PR #2355, opened 2026-08-21
fix(plugin-sqlite): report a query that stops early instead of returning the rows it got

datlechin:
Found while investigating #2342. This does **not** close that issue: the reported hang did not reproduce here, and what it fixes is a different, measured defect on the same path. See the note at the end for where #2342 stands.

## The defect

All three `sqlite3_step` loops in the SQLite driver were written as `while sqlite3_step(stmt) == SQLITE_ROW`, so the code that ends the loop was never looked at. `SQLITE_ROW` is one of six things a step can return, and every other one exits the loop the same way a clean finish does.

The result is that a query that failed halfway, or never started, is reported as a successful result containing whatever rows were read before it stopped:

- A database locked by another process returns zero rows and no error, so the grid renders an empty table.
- An I/O error partway through a scan on a network volume returns the rows read so far as the complete table.
- A corrupt page does the same.

Measured against the system SQLite (3.54.0, which is what `project.yml:526` links), with a second handle holding `BEGIN EXCLUSIVE` while a prepared statement steps:

```
PROBE rows=0 terminatingCode=5 isDone=false
PROBE oldBehaviour: returns a successful result with 0 rows and no error
PROBE newBehaviour: throws 'database is locked'
```

The `prepare` call was already checked, which is why this only shows up when the lock is taken after the statement is prepared, or when the failure happens mid-scan.

## The fix

Each of the three loops keeps its terminating code and reports anything that is not `SQLITE_DONE` as the error SQLite gives for it. The buffered paths throw; the streaming path finishes its continuation with the error rather than completing normally, after yielding what it had. Hitting the row cap is still a clean truncation, not an error.

## Verification

- `xcodebuild -scheme TablePro build`: passes. The SQLite driver is a bundled plugin, so this compiles it.
- The probe above, before and after.
- New UI test `SQLiteFirstTableLoadUITests`: opens the sample SQLite database, clicks a table, and asserts the rows arrive and the "Executing…" indicator clears. It passes on `main`, which is part of why #2342 is still open; it is here so the flow the issue describes has a test at all.

No unit test: the SQLite driver has no test target, its sources are not in `TableProTests`, and the actor that owns these loops is private to the plugin. Making it testable means putting the plugin's sources in the test target, which is a change worth making on its own rather than inside this one.

## Where #2342 stands

Not reproduced, so not fixed. What was ruled out, all by measurement:

- **The driver's connect path.** `sqlite3_open` is lazy and took 0.1ms; the file is untouched until the first statement.
- **A second connection blocking the first.** TablePro opens a pooled metadata connection alongside the browse query, so this was the leading theory. An uncommitted writer does not block a second handle's reads in rollback-journal mode, and WAL never blocks them.
- **Row count.** Nothing in the driver treats an empty table differently, which matches the reporter saying an empty table hangs the same way.

What is still open, and what would settle it: whether the stall is in the app's execution-ownership state (`retireQueryTask` clears the toolbar only for the claim that installed it) or in a wait that never returns. Both would need the reporter's own database, or the `TableLoadTracer` output from a session where it happens.

Two things worth fixing separately, both confirmed while investigating:

1. `sqlite3_interrupt` does not break a busy wait, measured. Stop therefore cannot end a lock wait, only the app's own state, so the driver keeps waiting out the busy timeout with nothing to stop it. A `sqlite3_busy_handler` that checks a cancel flag would make Stop reach it.
2. Every non-parameterized DML statement reports "0 rows affected": `executeUserQuery` returns a hardcoded `rowsAffected: 0` for the streaming path, and `sqlite3_changes` is only called on the two buffered ones. MySQL and PostgreSQL report the true count, so SQLite is the only driver that tells a user their `DELETE` did nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed a crash on macOS 26 and later when the editor redrew a diagnostic underline or search highlight whose text had been edited away.
 - Fixed a crash when an input method, dictation or Look Up asked the editor about text that had already been edited away. (#2339)
 - Fixed crashes when the editor's layout, syntax highlighting or accessibility read text that a newer edit had already removed. (#2340)
+- A SQLite query that stops early, because the database is locked or the volume returns an error, now says so instead of showing an empty table.
 - The XLSX, MQL and SQL Import plugins linked to a documentation page that did not exist. They now point at Import & Export.
 
 ## [0.67.0] - 2026-08-21
--- a/Plugins/SQLiteDriverPlugin/SQLitePlugin.swift
+++ b/Plugins/SQLiteDriverPlugin/SQLitePlugin.swift
@@ -173,7 +173,8 @@ private actor SQLiteConnectionActor {
         var rowsAffected = 0
         var truncated = false
 
-        while sqlite3_step(statement) == SQLITE_ROW {
+        var stepResult = sqlite3_step(statement)
+        while stepResult == SQLITE_ROW {
             if rows.count >= PluginRowLimits.emergencyMax {
                 truncated = true
                 break
@@ -200,6 +201,11 @@ private actor SQLiteConnectionActor {
             }
 
             rows.append(row)
+            stepResult = sqlite3_step(statement)
+        }
+
+        if !truncated, stepResult != SQLITE_DONE {
+            throw SQLitePluginError.queryFailed(String(cString: sqlite3_errmsg(db)))
         }
 
         if columns.isEmpty {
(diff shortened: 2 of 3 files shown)


datlechin:
One more finding, which invalidates the coverage this area already had.

`SampleDatabaseLauncher.open` finishes by routing `.openTable(table: "Track")`, so opening the sample database opens a table for you. Every UI test built on `launchWithSampleDatabase` therefore clicks its table against a connection that has already connected, fetched metadata and run a query: it tests the second query, never the first. `MainContentCoordinator.swift:1249-1292` only spawns the concurrent schema fetch when the table's metadata is uncached, which is exactly the first-open state #2342 describes and the one that route skips.

I could not get a cold session in the harness. A relaunch restores the last session rather than showing the welcome window, and after File > Close Connection the welcome window's connection row would not resolve by value, label, identifier, or a descendants-of-any predicate. Rather than leave a red test in this PR, the suite keeps the case that passes and the finding is written into its doc comment, so the next person does not spend the same three runs discovering that the launch route cannot reach the reported state.

PR #2366, opened 2026-08-21
test(ci): quarantine test cases instead of whole suites, and check the entries mean something

datlechin:
The macOS gate skipped 39 whole suites. I ran them: 483 of those cases pass, 53 fail.

Nine tenths of what the gate was skipping was skipped because a sibling case in the same file was broken. Nine of the 39 suites were entirely green: `QueryExecutorTests` (39 cases), `SQLStatementGeneratorCompositePKTests` (21), `CoordinatorEditorLoadTests` (14), `HostKeyStoreTests` (11), all four `EtcdQueryBuilder*` suites (23 between them), `EtcdQueryBuilderBrowseTests` (5).

The quarantine is now one **case** per line. 54 lines instead of 39 suites, and the local run is:

```
Test run with 12149 tests in 1420 suites passed after 131.755 seconds.
```

482 cases rejoin the gate.

## The entries had to be checkable, because one form of them silently does nothing

Before writing 54 lines I checked that a per-case line works at all. It does not, in the form the old parser accepted:

```
-skip-testing:TableProTests/EtcdPrefixRangeEndTests/allMaxBytes     -> ran 7 cases, 1 failed
-skip-testing:TableProTests/EtcdPrefixRangeEndTests/allMaxBytes()   -> ran 6 cases, 0 failed
```

A Swift Testing case is only skipped with its parentheses. Without them xcodebuild accepts the argument, skips nothing, and says nothing. The line reads exactly like a working one. The old parser's regex `^[A-Za-z_][A-Za-z0-9_]*(/[A-Za-z_][A-Za-z0-9_]*)?$` accepted the bare form and rejected the working one, so every per-case entry anyone wrote would have been inert.

That is the same failure `FreeTDSClassifierTests` had for 479 commits, and a grep of the sources cannot catch it: the function exists, it is the argument form that is wrong. So `quarantine-args.sh` now checks entries against the identifiers **xcodebuild itself enumerates** rather than against a grep. Enumeration costs 2 seconds for 13,199 identifiers, and it subsumes the old check.

Mutation-tested, all four cases:

```
missing parens     exit=1  did you mean 'EtcdPrefixRangeEndTests/allMaxBytes()'? A Swift Testing case needs its parentheses
bad case name      exit=1  'EtcdPrefixRangeEndTests' exists but has no case named 'noSuchCase()'
deleted suite      exit=1  nothing by that name is enumerated; delete the line
whole suite        exit=0  -skip-testing:TableProTests/EtcdPrefixRangeEndTests
```

A whole-suite entry still works, so nothing has to be rewritten per case to stay quarantined.

The UI quarantine had the same hole from the other direction: `list_tests.py` filters by name and an entry matching nothing was silently ignored, so the case it was meant to hold back ran anyway. It now fails naming the inert entries. The file is empty today, which is why this never bit.

## The four reasons in the file were mostly wrong

The old file put 20 suites under "bundled .tableplugin drivers do not activate in the headless test host". That is not what is happening. The test host **is** `TablePro.app` and all 14 bundled plugins are in its `Contents/PlugIns`.

What is actually happening, read off the failures:

- **21 cases assert a fallback that was deliberately deleted.** `TableOperationSQLBuilder.foreignKeyDisableStatements()` is `adapterProvider()?.foreignKeyDisableStatements() ?? []`. There is no built-in `DatabaseType` switch any more. `TableOperationsPluginTests` still opens with "the coordinator falls back to built-in DatabaseType switches. These tests verify that fallback." The empty result is correct; the expectation is stale. These need rewriting against a stub adapter.
- **8 assert a string the app no longer produces**, e.g. `DatabaseType.scylladb.iconName` is `scylladb-icon` and the test wants `cassandra-icon`.
- **3 are stale fixtures**: the decoded model gained a field the fixture JSON lacks, so decoding throws before the assertion (`Key 'agentSocketPath' not found`, `expected Int. Path: rowHeight`).
- **12 are change-tracking drift**: `reloadVersion` no longer increments where they expect, undo restores a different working set.
- **3 need a registered driver**: both duplicate-ID checks report "an error was expected but none was thrown", which is what an empty registry gives you.
- **3 are environment-coupled** (login keychain, an app installed on the machine).
- **3 are genuine behaviour differences** needing investigation.

Each group now says which of those it is, so the next person does not start from a label that sends them looking at plugin loading.

## What this does not do

It does not fix the 53. Each needs a judgement about whether the test or the code is wrong, and for the largest group the answer is clearly the test, but "clearly" is not "verified". They are still skipped, now individually, with an accurate reason each.

## Verification

Full local run with the new file: `xcodebuild exit=0`, 12,149 cases, 1,420 suites, 131.755s.

Getting there took two passes. The first run left exactly one case failing, `ConnectionURLFormatterSSHProfileTests/noProfileFallbackUsesInlineConfig()`, which my per-suite sweep had missed; it is in the file and the second run is green.

CI is not this machine, so if a case that passes here fails on the runner, this PR's own run will name it and it joins the list with that evidence.

`actionlint` and `shellcheck -x --severity=warning` clean.

### DIFF
--- a/.github/macos-test-quarantine.txt
+++ b/.github/macos-test-quarantine.txt
@@ -1,51 +1,92 @@
-# Quarantined macOS test suites (skipped by the macOS Tests CI gate).
+# Quarantined macOS test cases (skipped by the macOS Tests CI gate).
 #
-# These suites fail or hang in the headless XCTest host for reasons documented below,
-# not because the app code is wrong. The gate runs every other suite. Burn this list
-# down: fix a suite, delete its line, and it rejoins the gate. See
-# specs/claude-code-refactor-roadmap.md and the macOS-test gotchas note.
+# One test case per line, as xcodebuild enumerates it: Suite/case(). '#' starts a comment.
+# quarantine-args.sh checks every line against the real enumeration and fails the job on any entry
+# that would skip nothing, so a renamed case cannot leave an inert line behind and a Swift Testing
+# case cannot lose its parentheses and silently rejoin the run.
 #
-# Format: one Swift Testing suite type name per line. '#' starts a comment.
+# This file used to name whole suites. 39 suites were listed, and running them measured 483 passing
+# cases against 53 failing ones: nine suites in the list were entirely green and the rest were
+# failing in ones and twos. Nine tenths of what the gate skipped was skipped because a sibling case
+# in the same file was broken.
+#
+# Burn this list down: fix a case, delete its line, and it rejoins the gate.
 
-# --- Driver-dependent: bundled .tableplugin drivers do not activate in the headless test
-#     host (only plugin metadata loads), so driver-backed SQL generation returns empty.
-TableOperationsPluginTests
-TableQueryBuilderFilteredQueryTests
-SQLCompletionProviderTests
-SQLStatementGeneratorPKRegressionTests
-SQLStatementGeneratorCompositePKTests
-EtcdQueryBuilderBrowseTests
-EtcdQueryBuilderCountTests
-EtcdQueryBuilderFilteredTests
-EtcdQueryBuilderTagTests
-EtcdPrefixRangeEndTests
-DataChangeManagerClickHouseTests
-QueryExecutorTests
-ClickHouseDialectTests
-ValidateDriverDescriptorTests
-PluginCapabilityTests
-DatabaseTypeCassandraTests
-RowOperationsManagerTests
-StructureGridDelegateAddRowTests
-SaveCompletionTests
-ChangeReapplyVersionTests
+# --- Asserts a fallback that no longer exists.
+#     TableOperationSQLBuilder delegates every statement to the connected plugin adapter and returns
+#     nothing when there is none: `adapterProvider()?.foreignKeyDisableStatements() ?? []`. The
+#     built-in DatabaseType switches these expect were deleted, so the empty result is correct and
+#     the expectation is what is stale. Each needs rewriting against a stub adapter.
+TableOperationsPluginTests/combinedMySQLWithFK()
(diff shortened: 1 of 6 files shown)


datlechin:
Pushed a second commit. Testing the UI half turned up the same bug from the other direction.

The UI quarantine file's header has always said entries are `"Suite/testMethod"`. That form never filtered anything:

```python
def is_quarantined(identifier):
    parts = identifier.split("/")
    suite = parts[1] if len(parts) > 1 else identifier
    return suite in skipped or identifier in skipped
```

`identifier` is target-qualified (`TableProUITests/Suite/case()`), so it only ever matched a bare suite name or a fully-qualified line. The documented per-case form matched neither. The file is empty, so nobody had hit it, and my new inert-entry check passed it happily because the entry *does* name a real case: it is the matching that was wrong, not the entry.

Both now go through one `forms()` helper, so validation and filtering cannot disagree about what a line means. Measured on the real bundle:

```
no quarantine                  69 cases
per-case entry                 68 cases
whole-suite entry              68 cases
entry matching nothing         exit=1, names the inert entry
```

I also used it for the one thing it was obviously for. `InspectorToolbarPlacementUITests` needs a 1512pt-wide screen and the runner is 1024x768, so it has an `XCTSkipUnless` and has never run on CI. It has also never failed, which is the problem: a skip inside a test is invisible to anyone reading the gate. It is now a line in the quarantine file with its reason, which makes "this gates nothing" reviewable, and with the way to get it back written down: the assertion is geometric, an `NSWindow` can exceed the display as long as it is not ordered front, so building the window and toolbar in a `TableProTests` case would run at any resolution. That is a rewrite rather than a configuration change, so it is not in this PR.

PR #2400, opened 2026-08-24
feat(plugin-beancount): expose pad directives

### LINKED ISSUE #2399: Expose Beancount pad directives as a table
## The situation

TablePro projects Beancount directives into read-only SQL tables, but pad directives are not queryable as directives.

## What works today, and why it is not enough

The balancing transaction generated by a pad is visible in `transactions`. That does not expose the pad's target account, source account, or original source location as one queryable record.

## Ask

Add a read-only `pads` table with the date, target account, source account, and source location. Keep the rustledger and Python backends aligned, including pads declared in included files.

salmonumbrella:
## What changed

- Add a read-only `pads` table with target and source accounts and source locations.
- Keep the rustledger and Python projections aligned, including pads declared in included files.
- Document the 18-table Beancount surface.

## Why

Pad-generated balancing transactions are visible today, but the original directive and its source account are not queryable. This exposes that missing record without changing the ledger or the existing transaction projection.

## Usage

```sql
SELECT date, account, source_account, source_location
FROM pads
ORDER BY date;
```

Closes #2399


datlechin:
Pushed `3fc2caba2` to this branch after a review pass. Summary of what changed and why, since it replaces the rledger half of the feature rather than patching it.

## The problem

On the rledger backend the branch recovered a pad's two accounts by opening the ledger file, indexing to the `lineno` rledger reported, and tokenizing that line. Four things went wrong there, all measured against rledger 0.22.0 and Python Beancount 3.2.3:

- **CRLF drops every pad.** Swift treats `\r\n` as a single `Character`, so `contents.split(separator: "\n")` returns **1 element** for a whole Windows-saved ledger. Every `lines[safe:]` lookup past index 0 is nil.
- **Slash dates drop every pad.** Beancount's lexer is `[0-9]{4,}[\-/][0-9]+[\-/][0-9]+`, so `2024/01/30` is valid and rledger reports its date back as ISO `2024-01-30`. The dash-only `dateComponents` then failed the guard.
- **`;` comments leak.** A comment needs no leading whitespace, so `... Equity:Opening-Balances;opening` put `Equity:Opening-Balances;opening` into `source_account`, which breaks the join against `accounts.name` the docs recommend.
- **The date guard could return the wrong accounts.** It compared two `Optional`s, so when neither side parsed, `nil == nil` passed. Measured: a garbage line returned `(Assets:Wrong, Equity:Wrong)` rather than being rejected.

Two more the review found: a BOM-prefixed ledger drops its first-line pad (rledger accepts the BOM and reports the pad at line 1), and an account containing U+00A0 (legal Beancount) is shredded by `\.isWhitespace`.

## Why the scanner is gone rather than fixed

You were right that rledger exposes no pad accounts as query columns. `directive_to_entry_row` in rustledger-query fills `flag/payee/narration/tags/links/accounts` only for `Directive::Transaction`, `accounts` comes back `[]` for a pad, and there is no `#pads` table.

But they are reachable through **`PRINT`**, which renders each directive from rledger's own parsed AST (`execution.rs:1221` is `Directive::Pad => format!("{} pad {} {}\n", pad.date, pad.account, pad.source_account)`). Measured on CRLF, slash-dated and `;`-commented ledgers, it returns the same normalized line every time:

```
"2024-01-30 pad Assets:Cash Equity:Opening-Balances\n"
```

So the ledger file is never opened, and all six defects above stop being reachable instead of being defended one at a time.

`padsQuery` now selects the `id` it orders by and orders by `id` alone, which is the order `PRINT` iterates (`execute_print` walks `resolved_directives()`, the same iterator whose index becomes `#entries.id`). The two lists are zipped by position and every pair's ISO date must match; any mismatch writes a row to the existing `diagnostics` table with `phase` `projection` instead of being dropped into an OSLog warning.

`FROM FALSE` is load-bearing: `PRINT FROM type = 'pad'` does **not** filter (measured, returns everything), and bare `PRINT` is 6.7 MB on a 50k-transaction ledger versus 2.1 KB for `PRINT FROM FALSE`, which runs in 0.18s, slightly cheaper than the pads `SELECT` already being made.

## Other fixes in the same commit

- `pads.source_file`, `line` and `source_location` are now nullable, matching `transactions` and `postings`. A plugin-generated pad carries `filename: '<generated>', lineno: 0` (measured on Beancount 3.2.3) and was previously dropped whole despite its date and both accounts being known.
- The pad path uses the `stringValue()` helper rather than raw `as? String`.
- Removed the dead `enriched["location"]` assignment, which duplicated what `sourcePosition` already derives.
- `(#2399)` added to the CHANGELOG entry.
- Docs record that rledger derives the accounts from `PRINT`, verified against rustledger 0.22.0.

## Tests

The correlation is now a pure function, `padProjection(entries:directives:)`, so it is covered by six tests that **run in CI with no backend installed**. That was the real coverage gap: both integration tests are `.enabled(if:)`-gated and no workflow installs `rledger` or Beancount, so the whole rledger path was shipping untested.

The shared integration fixture now writes its pad ledger as CRLF with a slash date and a no-space comment, so the existing assertions became the end-to-end guard.

## Two things worth knowing

- The positional correlation is an implementation property of rustledger 0.22.0, not a published contract. It is guarded by the per-pair date check, and I verified it stays stable across runs including two pads sharing one date (both sides keep source order).
- `PRINT FROM FALSE` renders every non-transaction directive. On a ledger with a large price database that is real output: 50,000 `price` directives produce 3.45 MB. Still 0.18s, and only when the ledger has at least one pad, but it cannot be narrowed server-side.

Separately, and not touched here: a ledger containing an **unused pad** cannot be opened at all on the Python backend, because the projection script does `if errors: sys.exit(1)` and Beancount reports `PadError: Unused Pad entry`. rledger opens it fine. Worth its own issue.

PR #2401, opened 2026-08-24
fix(connections): give the connection color its own surface instead of the engine glyph

### LINKED ISSUE #2398: The color of the connections could be more vivid.
### Problem

Currently, when choosing the connection color, it is reflected in the left sidebar and the top connection icon, but it is not very obvious because the icons are originally colored. Would you consider adding some more prominent color cues, such as the background color of the connection name and  the database capsule on the top, possibly with transparency attached to the selected color?

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Related database type

None

datlechin:
Fixes #2398.

The reporter says the connection color "is reflected in the left sidebar and the top connection icon, but it is not very obvious because the icons are originally colored", and asks for a background behind the connection name.

Three separate causes, all confirmed in the code.

## 1. The color was spent recoloring an already-branded glyph

`DatabaseConnection.displayColor` was `color.isDefault ? type.themeColor : color.color`. One property answered two different questions, "which engine is this" and "which connection is this", and every consumer painted the answer onto the same pixel: the database type icon. Picking a color therefore *replaced* the engine's brand color rather than adding a signal, so the only visible change was a hue shift on a 14pt icon, and the engine lost its own color to pay for it.

Four consumers: `ConnectionStatusView`, `WelcomeConnectionRow`, `WorkspaceRailCellView`, `ConnectionSwitcherPopover`.

Split into `brandColor` (always the engine's) and `identityColor: ConnectionColor?` (nil when the user picked none, never falling back to the brand color).

## 2. The window-chrome tint has been dead code since April

`dc0093243` ("refactor: AppKit-native window/toolbar management for main editor", #768, 2026-04-18) removed `.modifier(ToolbarTintModifier(connectionColor:))` and left the comment "Connection color tint is not yet ported". It was never ported. `ToolbarTintModifier` had zero call sites for four months, so there was no tint at all, at any opacity, and raising its `0.12` would have shipped no visible change.

It could not have been revived as written either: `.toolbarBackground(_:for: .windowToolbar)` styles a toolbar SwiftUI itself installs and reaches it by flowing up to a `Window` scene's root view. The app runs the AppKit lifecycle with no SwiftUI scene and installs a real `NSToolbar` in `MainWindowToolbar`. Deleted, with a source-scanning guard test so it cannot come back silently.

## 3. Neither the toolbar nor the rail repainted while connected

This one would have made the whole fix invisible at the moment the user picks a color.

- `ConnectionToolbarState` is built once per session at `SessionStateFactory`. Its only writer, `update(from:)`, is reachable from `init` and from `initializeToolbar()`, whose sole caller `initializeView()` has no callers at all. `handleConnectionRecordChange` only reassigned SwiftUI root views and never touched it.
- `DatabaseManager.observeConnectionUpdates` reconciled `safeModeLevel` and nothing else, while `WorkspaceRailStore.resolve` reads `session.connection` for any live session.

So editing a live connection's color, or its name, changed nothing until reconnect. Both paths now take the stored record; the session keeps only the two fields it actually owns at runtime, the browsed database and the safe-mode level.

## The design

The reporter's literal ask, a translucent wash over the toolbar background, is what the HIG rules out and what measurement rules out.

- Toolbars: "Reduce the use of toolbar backgrounds and tinted controls. Any custom backgrounds... might overlay or interfere with background effects that the system provides."
- Color: "Avoid using the same color to mean different things." "Avoid relying solely on color to differentiate between objects."

And a wash sits over a material whose backdrop is the wallpaper, so no contrast ratio can be guaranteed for it, and it collapses under Reduce Transparency.

What Apple ships for this exact problem is Safari Profiles: the profile's color and symbol in a named toolbar button, with the toolbar material and window frame left alone. TablePlus puts the color on its toolbar status item and Postico "highlights the status display" the same way. TablePro already shipped the shape too, in `tagBadge`.

So the color now fills a capsule behind the connection name in the toolbar, and a band behind the label in the workspace rail. In the welcome list and the connection switcher it is a dot beside the name, the Finder-tag idiom, because those rows can be selected and a fill of their own would fight the selection band. The scope chips stay neutral: one filled shape per window.

`WorkspaceRailCellView`'s header comment argued the opposite ("the colour is the icon"), so this PR refutes it in place rather than reversing it silently: that premise holds for a Finder icon, a neutral document or folder shape with no color of its own to lose, and fails for a brand mark that is already a color meaning something else.

Color is never the only channel. `ConnectionColor.displayName` now joins the tooltip and the accessibility label on every identity surface.

## Contrast

Text on a fill needs 4.5:1 (WCAG 2.2 SC 1.4.3). Measured against the palette with the label `legibleForeground(on:)` picks, five of the eight only reach 3.2:1 to 4.2:1:

| | light | dark |
|---|---|---|
| red | 3.57 | 3.43 |
| blue | 3.52 | 3.23 |
| purple | 4.17 | 3.63 |
| pink | 3.65 | 3.52 |
| gray | 3.26 | 7.31 |
| orange / yellow / green | 9.09 / 13.89 / 9.46 | 9.41 / 14.87 / 10.39 |

Also measured: `accessibilityHighContrastAqua` resolves `systemRed` and the rest identically to `.aqua`, so the palette does not self-adjust for Increase Contrast.

`ConnectionColor.labelledFill` lowers brightness in HSB, holding hue and saturation, only as far as the derived label needs. Orange, yellow and green already clear the bar with a black label and are returned untouched. The rest keep 83% to 96% of their brightness, so they still read as the color that was picked:

| | fill → tuned | brightness kept |
|---|---|---|
| purple | `#CB30E0` → `#C22ED6` | 96% |
| pink | `#FF2D55` → `#E2284B` | 89% |
| red | `#FF383C` → `#E03135` | 88% |
| blue | `#0088FF` → `#0076DE` | 87% |
| gray | `#8E8E93` → `#76767A` | 83% |

`ConnectionColor.color` is untouched, so the picker swatch, dots and glyph tints keep full saturation. Only a fill that carries text uses `labelledFill`. The existing tag badge adopts it too, because it had the identical defect at 10pt.

## Also fixed, because the model change makes them unrepresentable

`ConnectionColor.none` renders as `Color.clear`, so a render site that filled with it painted a transparent hole. `indicatorColor` and `labelledFill` return `nil` instead, which fixes two shipped instances found while investigating:

- The query history drawer painted an invisible dot for every default-colored connection. Its `.secondary` fallback was unreachable because the directory always passed a non-nil color.
- The compare status strip did the same, while its own sibling `CompareEndpointPicker.swatch` guarded `.none` correctly. The strip is the surface that says which database is about to be written to.

## Verification

- `build` PASS
- `test` PASS, 81 of 81 across `ConnectionIdentityColorTests`, `ToolbarBackgroundGuardTests`, `LegibleForegroundTests`, `WorkspaceRailStoreTests`, `ConnectionToolbarStateTests`, `SafeModeMigrationTests`, `HistoryRowTintTests`, `GroupMenuEntriesTests`, `DataGridCellSelectionFillTests`, `CompareEndpointPickerModelTests`, `DatabaseTypeBrandColorTests`, `MultiConnectionTests`
- `swiftlint lint --strict` clean
- `docs/scripts/check-writing-style.sh` and `check-docs-against-source.py` both pass

New tests: every `labelledFill` clears 4.5:1 in all four appearances; tuning only ever dims and never shifts hue; an already-legible fill is returned unchanged; `brandColor` does not move when a color is picked; neither accessor hands back a paintable color for `.none`; a color or name picked while connected reaches the live session, and the session's switched database survives the reconcile; no source file reaches for `.toolbarBackground(`.

## Screenshots

Not captured. The change is visible in the toolbar, the workspace rail and the welcome list, and a faithful before-and-after needs the app driven to an open connection on both this branch and a build of `main`. Rendering the views offscreen would show the new state but not an honest "before". Worth capturing before merge if a reviewer wants them.

## Not done here

- The workspace rail band and the toolbar capsule are covered by unit tests, not `TableProUITests`. Both surfaces need a live connection and the rail only appears when a window hosts more than one, so neither runs deterministically without a seeded fixture.
- Mobile parity. `ConnectionColor` has two declarations (`DatabaseConnection.swift` and `TableProModels/ConnectionColor.swift`) and `TableProMobile` carries a third color mapping of its own that maps `.none` to grey rather than clear. Left alone deliberately; noted so it is not mistaken for an oversight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The data grid draws its own column separators. (#2381)
 - The inline cell editor scrolls a long line instead of wrapping it. (#2381)
 - Row inspector fields, cell popovers and the Compare row diff follow the data grid font. (#2393)
+- The connection color as a filled badge behind the connection name, with the database icon back to its engine color. (#2398)
+- Deeper fills on tag badges and the connection name badge. (#2398)
 
 ### Fixed
 
+- A connection's color and name not reaching the toolbar and workspace rail until the next reconnect. (#2398)
+- An invisible connection dot in the query history drawer and the compare status strip. (#2398)
 - The JSON viewer keeping its old font after a font, theme or text-size change. (#2393)
 - Hex dumps wrapping mid-line instead of keeping their columns aligned. (#2393)
 - A second of delay opening the inline editor on a result with hundreds of columns. (#2381)
--- a/TablePro/Core/Database/DatabaseManager+Sessions.swift
+++ b/TablePro/Core/Database/DatabaseManager+Sessions.swift
@@ -512,17 +512,41 @@ extension DatabaseManager {
         connectionUpdatedCancellable = AppEvents.shared.connectionUpdated
             .receive(on: RunLoop.main)
             .sink { [weak self] connectionId in
-                self?.reconcileSafeModeLevel(for: connectionId)
+                self?.reconcileStoredRecord(for: connectionId)
             }
     }
 
-    func reconcileSafeModeLevel(for connectionId: UUID?) {
+    func reconcileStoredRecord(for connectionId: UUID?) {
         let targetIds = connectionId.map { [$0] } ?? Array(activeSessions.keys)
         for id in targetIds {
-            guard activeSessions[id] != nil,
+            guard let session = activeSessions[id],
                   let stored = connectionStorage.loadConnection(id: id) else { continue }
+            adoptStoredRecord(stored, keepingRuntimeStateOf: session, for: id)
             setSafeModeLevel(stored.safeModeLevel, for: id)
         }
+    }
+
+    /// The stored record owns every field the connection form can edit. A live session owns only
+    /// what it switched at runtime, which is the browsed database and the safe-mode level, so those
+    /// two are carried over and the rest is taken from storage.
+    ///
+    /// This used to reconcile `safeModeLevel` alone, which left everything else on the session
+    /// frozen at connect time. `WorkspaceRailStore.resolve` reads `session.connection` for any live
+    /// session, so renaming or recolouring a connection changed nothing in the rail until the next
+    /// reconnect (#2398).
+    private func adoptStoredRecord(
+        _ stored: DatabaseConnection,
+        keepingRuntimeStateOf session: ConnectionSession,
+        for connectionId: UUID
+    ) {
+        var reconciled = stored
+        reconciled.database = session.connection.database
+        reconciled.safeModeLevel = session.connection.safeModeLevel
+        guard reconciled != session.connection else { return }
+
+        var updated = session
+        updated.connection = reconciled
+        setSession(updated, for: connectionId)
     }
 
     func setSafeModeLevel(_ level: SafeModeLevel, for connectionId: UUID) {
--- a/TablePro/Core/Database/DatabaseManager+Sessions.swift
+++ b/TablePro/Core/Database/DatabaseManager+Sessions.swift
@@ -521,27 +521,31 @@ extension DatabaseManager {
         for id in targetIds {
             guard let session = activeSessions[id],
                   let stored = connectionStorage.loadConnection(id: id) else { continue }
-            adoptStoredRecord(stored, keepingRuntimeStateOf: session, for: id)
+            adoptDisplayFields(from: stored, into: session, for: id)
             setSafeModeLevel(stored.safeModeLevel, for: id)
         }
     }
 
-    /// The stored record owns every field the connection form can edit. A live session owns only
-    /// what it switched at runtime, which is the browsed database and the safe-mode level, so those
-    /// two are carried over and the rest is taken from storage.
+    /// Carries the fields a live session only ever *displays* across from storage, and nothing else.
     ///
-    /// This used to reconcile `safeModeLevel` alone, which left everything else on the session
-    /// frozen at connect time. `WorkspaceRailStore.resolve` reads `session.connection` for any live
-    /// session, so renaming or recolouring a connection changed nothing in the rail until the next
-    /// reconnect (#2398).
-    private func adoptStoredRecord(
-        _ stored: DatabaseConnection,
-        keepingRuntimeStateOf session: ConnectionSession,
+    /// This used to reconcile `safeModeLevel` alone, so everything else stayed frozen at connect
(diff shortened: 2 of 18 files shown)


datlechin:
Second commit, `e2f4eb076`, from a self-review pass. Two of these were real defects in the first commit.

**A live session's connect target could move under it.** `adoptStoredRecord` adopted the whole stored record, and `reconnectOntoDatabase` builds its reconnect from `session.connection`. So editing host, port, username or SSH config on an open connection, or an iCloud pull carrying another Mac's edit through `connectionUpdated.send(nil)`, would let the health monitor silently reconnect an open window with its tabs to a different server. Narrowed to an allowlist of the fields a session only ever displays (`name`, `color`, `tagIds`) and renamed `adoptDisplayFields`. A new test asserts host, port and username do not follow a storage edit.

**`labelledFill` froze half the palette.** `tunedForLegibleLabel` ends at `NSColor(hue:saturation:brightness:alpha:)`, a concrete colour in whatever appearance was current, and `Color(nsColor:)` carried that fixed value. Measured, built once and read under both appearances:

| | light | dark |
|---|---|---|
| red (goes through the search) | `#DB393B` | `#DB393B` |
| orange (hits the early return) | `#FF8D28` | `#FF9230` |

So the five tuned colours stopped following a Light/Dark switch while the three untouched ones kept doing it. Now built with `NSColor(name:dynamicProvider:)`, which AppKit resolves against the appearance it draws in, so no call site has to remember to observe the colour scheme. `WorkspaceRailCellView` already avoided this by deferring resolution into `performAsCurrentDrawingAppearance`; the SwiftUI call sites did not.

Three smaller ones:

- The welcome list dot only rendered when a colour was set, so a coloured row's name sat 14pt right of an uncoloured one and the column had a ragged left edge. An uncoloured connection now gets a neutral dot, matching the connection switcher.
- `ConnectionToolbarState.update(from:)` wrote six `@Observable` properties with no equality guard, on a path that fires on every connection save. Guarded field by field, like the rest of that class.
- `tunedForLegibleLabel` returned pure black if no candidate passed, silently losing the hue. Unreachable at 4.5:1, but it takes the ratio as a parameter, so it now falls back to the colour it was given.

Not changed, and worth a reviewer's opinion: `.fixedSize(horizontal: true)` on the connection name at `ConnectionStatusView.swift:68` means `lineLimit(1)` and `truncationMode(.tail)` can never fire, so the toolbar item is always the name's full ideal width. That predates this PR; the capsule adds 16pt of padding on top of it. Fixing toolbar sizing properly is a separate change.

Re-verified: build PASS, 82 of 82 tests, `swiftlint lint --strict` clean.

PR #2429, opened 2026-08-25
test: burn the macOS quarantine down to 6, and let main report on itself

datlechin:
Takes `.github/macos-test-quarantine.txt` from 14 entries to 6. Every removal was run green first; nothing was removed on the strength of a reading.

## What the failures actually were

Six groups, and none of them was a bug in the app. In each case the test held a lever the app had moved, or the test's own setup could not reach the code it asserted.

| Suite | Why it failed |
| --- | --- |
| `SaveCompletionTests` (3) | No connected driver, so `assemblePendingStatements` produced nothing and `saveChanges` bailed at "Could not generate SQL for changes." before touching the `inout` parameters. The reason recorded in the quarantine was wrong: `scope(for:)` falls back to a scope built from the tab, so `selectedTabScope` is never nil once a tab exists. |
| `ValidateDriverDescriptorTests` (2) | Assumed MySQL was registered. It never was: discovery ends at "will load on first use" and the bundles do not register in the xctest host at all, so the duplicate check had nothing to collide with. |
| `SQLCompletionProviderTests` (1) | Two bugs in the test. It expected a bare `user_name`, but `allColumnsInScope` qualifies whenever more than one table is in scope, so two tables always produce `u.user_name`. And `SQLSchemaProvider.cachedDriver` is `weak`, so the mock deallocated and `getColumns` returned nothing at all. |

`driverPlugins` is `internal(set)`, so a `@testable` test can register its own occupant; `DatabaseManager.injectSession` takes a `ConnectionSession` with a driver. Both seams already existed.

## Coverage added, not just restored

- An alias-prefix case for completion. `WHERE u.` takes the dot-prefix branch and offers bare labels for that table alone, which nothing covered. The two shapes are asserted together so a later change cannot collapse them.
- `testFewerThan20Returned` asserted `items.count >= 0`, which is true of every possible result and so could not fail. It now asserts the keyword it is named for and the cap it is paired with.

## CI

A push to main is the only verdict that commit will ever get, and 13 of the last 30 main runs of the macOS suite were cancelled by the next merge, so whole days passed with no conclusive result. `cancel-in-progress` is now `${{ github.ref != 'refs/heads/main' }}` on both the macOS and iOS suites.

This does not queue up: with `cancel-in-progress` false GitHub still replaces the *pending* run in the group, so a burst of merges holds at most one running plus one queued, and the queued one is the newest.

## Verification

292 cases across every suite touched, all passing, on the merged tree:

- `SQLCompletionProviderTests SaveCompletionTests PluginValidationTests` — 135/135
- the ten remaining suites — 157/157

`actionlint` clean on both workflows. Two pre-existing swiftlint violations in `SQLCompletionProviderTests.swift` (sorted imports, empty count) are fixed here since the file was already open.

## Still quarantined, and why

Three undo-grouping cases: `NSUndoManager` leaves `groupsByEvent` on, so operations without a run loop turn between them land in one group. Injecting the undo manager was tried and hangs the test host; that is recorded in the file so the next attempt does not repeat it.

Three read as environment-coupled, but only one is. `KeychainHelperTests/writeOverwritesExistingValue` genuinely needs a keychain that accepts writes. The other two are misfiled and are noted in the PR discussion rather than changed here, because one of them is a product question rather than a test fix.

https://claude.ai/code/session_013MEaba8K1HQcyDNeq5wEFk


datlechin:
Two entries in the quarantine's "environment-coupled" block are misfiled. Neither is changed in this PR; both are recorded here so the block's reason stops being wrong.

**`TablePlusImporterTests/testImportConnections_mapsDriverCorrectly`** is a stale expectation. It expects TablePlus's `CockroachDB` to map to `PostgreSQL`, but CockroachDB is a first-class `DatabaseType` now, with its own default port 26257, icon, URL scheme and `.cockroachText` EXPLAIN format. `mapDriver` returning `"CockroachDB"` is correct and the test is what drifted. A one-line test fix, and it belongs in whatever PR next touches that suite.

**`SequelAceImporterTests/testIsAvailable_whenFileExists_returnsTrue`** is a product question, which is why I left it alone.

#1318 moved foreign-app detection from data files to LaunchServices and deleted SequelAce's `isAvailable()`, which had been `FileManager.default.fileExists(atPath: favoritesFileURL.path)`. Review feedback on that PR added a data-file arm back for two importers:

```swift
// DBeaverImporter
func isAvailable() -> Bool { installedAppURL() != nil || findDataSourcesFile() != nil }
// DataGripImporter
func isAvailable() -> Bool { installedAppURL() != nil || !locations().isEmpty }
```

SequelAce did not get one, so it takes the bare protocol default. A user who still has `Favorites.plist` but has uninstalled Sequel Ace is told the importer is unavailable, with their connections sitting right there. macOS usually leaves the container behind on uninstall, so this is reachable rather than theoretical.

Whether that is a bug or the intended narrowing is your call. If it is a bug, the fix is one line matching the other two, plus the `resolveAppURL` seam `TablePlusImporter` already has so the test can assert both directions hermetically:

```swift
var resolveAppURL: @Sendable (_ bundleIdentifier: String) -> URL? = {
    NSWorkspace.shared.urlForApplication(withBundleIdentifier: $0)
}
```

Worth noting either way: `testIsAvailable_whenFileMissing_returnsFalse` currently passes on CI only because Sequel Ace is not installed on the runner. It asserts nothing about the file.

datlechin:
Correction to my comment above, since it is now merged and someone may act on it.

I said the SequelAce importer was missing the `|| data present` arm that DBeaver and DataGrip carry, and that a user with a favorites file but no installed app would be told the importer is unavailable. **That was wrong, and it is not a bug.**

DBeaver and DataGrip have that arm because a JetBrains Toolbox or portable install can be invisible to LaunchServices, so `urlForApplication(withBundleIdentifier:)` returns nil for an app that is genuinely installed. A Sequel Ace install is an ordinary `.app` that LaunchServices always finds. The asymmetry is deliberate, not an oversight, and no change is wanted.

The quarantined case was simply stale against #1318, which moved detection to LaunchServices and deleted the file-based `isAvailable()`. Fixed in #2431 by driving an injectable `resolveAppURL`, the same seam `TablePlusImporter` already has, with a default identical to the protocol's own behaviour.

The TablePlus CockroachDB point in that comment stands and is also fixed in #2431.

PR #2430, opened 2026-08-25
perf(plugins): stop fetching query rows once the row cap is exceeded (#2427)

### LINKED ISSUE #2427: Use server-side pagination instead of automatically injecting a LIMIT
### Problem

TablePro currently adds a LIMIT to queries that do not already define one, based on the configured maximum number of rows to display.

However, this does not seem to prevent the database from processing the full query efficiently. For example, a query without an explicit LIMIT can take around 10 seconds, even though TablePro eventually displays only 1,000 rows according to the configured limit.

This suggests that limiting the displayed result set is not behaving as expected from a performance perspective.

Ideally, TablePro should avoid relying on modifying the original query with a LIMIT and instead implement server-side pagination, similar to DBeaver. This would allow TablePro to fetch only the rows needed for the current page and request additional rows as the user navigates through the result set.

### Proposed solution

Implement server-side pagination for query results:

Fetch only the first configured batch of rows (e.g. 1,000 rows).

Request subsequent batches from the database only when the user scrolls or navigates to the next page.

Avoid executing or materializing the entire result set before displaying the first batch.

Preserve explicit LIMIT clauses provided by the user.

Ideally, use database-specific pagination mechanisms where appropriate (LIMIT/OFFSET, cursors, fetch size, etc.).

This should improve the time-to-first-results and make large queries behave more like DBeaver.

### Alternatives considered

_No response_

### Related database type

None

datlechin:
Fixes #2427.

## The reported problem, and what is actually wrong

The issue asks TablePro to stop injecting a `LIMIT` and adopt DBeaver-style server-side pagination. Both halves of that premise turn out to be wrong, and the truth is worse than reported.

TablePro has never injected a `LIMIT`. `QueryExecutor.resolveRowCap` produces a cap and hands it to the driver, and `PluginDatabaseDriver.executeUserQuery`'s default implementation runs the query to completion, materializes every row, then keeps `Array(raw.rows.prefix(cap))`. The server produces every row, every row crosses the network, every row is decoded into `[[PluginCellValue]]`, and only then are 1,000 kept. That is the reporter's 10 seconds.

Meanwhile 18 driver plugins already ship a genuinely incremental `streamRows`, and PostgreSQL's already uses `PQsendQuery` + `PQsetSingleRowMode` with a `PQcancel` abort. Its only callers in the whole app are export and compare. The incremental read path and the query path were never joined.

## Measured

A C probe compiled against the real `Libs/libpq_arm64.a` (`PQlibVersion()` = 170004) against PostgreSQL 17.11 and 16.15:

| | |
|---|---|
| 3,000,000-row `PQexec` | 731 ms |
| Single-row read of 1,000 + `PQcancel` + drain to NULL | 4 ms, connection back to `IDLE` |
| Abandoning the read **without** the cancel | 775 ms (PG17) / 1,952 ms (PG16) charged to the user's *next* query |

Stopping the read without cancelling also leaves the backend `state=active`, `wait_event=ClientWrite`, holding `backend_xmin`. The cancel is correctness, not an optimization.

## What this does

A new additive `PluginDatabaseDriver` requirement:

```swift
func executeBoundedQuery(query: String, rowCap: Int) async throws -> PluginQueryResult?
```

`nil` means "I cannot bound the fetch at my source", matching this protocol's own convention (`injectRowLimit`, `buildExplainQuery`, `defaultExportQuery`). A new requirement with a default is ABI-additive, so no `currentPluginKitVersion` bump and no plugin re-release. `scripts/check-pluginkit-abi.sh` confirms it.

`PluginBoundedStream` is the shared collector, and `boundedQueryFromStream` is the one-line opt-in for a driver whose `streamRows` yields incrementally and whose producer stops when its consumer does.

**The default `executeUserQuery` is deliberately untouched.** Routing it into the bounded path would have been unsafe: `MCPConnectionBridge` sets `shouldCap = tier == .safe || hasReturning`, so `INSERT … RETURNING *` arrives with a row cap under `.protectedWrite`, a policy that exists because "a half-applied write cannot be undone by retrying". A bounded read would `PQcancel` mid-INSERT. Instead the app opts in at the call sites it controls, whose caps come from `resolveRowCap` and therefore only ever exist for reads. This also means the 17 registry-only plugins that will not be rebuilt are bit-identical to today.

### Drivers

| Bounded at the source | How |
|---|---|
| PostgreSQL, CockroachDB, Redshift, PGlite | new `boundedQuery` on `LibPQPluginConnection`: single-row mode, stop at cap+1, `PQcancel` + drain. Uses no stream cancellation at all. |
| SQLite | `Task.isCancelled` in its `sqlite3_step` loop; the actor method inherits the caller's task, so the cancel lands |
| Redis, DynamoDB, BigQuery | `try Task.checkCancellation()` inside the paging loop, producer not nested |
| MySQL, MariaDB, MongoDB | already did (`mysql_use_result` + `killQueryOnServer` + drain; `find` limit = cap+1). Unchanged. |

**Opting a driver in requires proving its producer stops, and five drivers do not.** Terminating the stream only cancels the task `streamRows` created, so the producer has to be reachable from it and has to poll. MSSQL (`MSSQLPlugin.swift:599`) and Oracle (`OraclePlugin.swift:425`) run theirs inside a *nested* unstructured `Task {}`, which `streamTask.cancel()` cannot reach because a plain `Task {}` inherits context but is not a child. Cassandra's `CassandraConnection.streamQuery:688` is a synchronous `cass_session_execute`/`cass_future_wait` paging loop with no cancellation check at all. Those three would have returned the first page instantly while the connection kept pulling the rest, so the user's next statement blocks behind a read nobody wanted, which is worse than not opting in. Snowflake and Trino carry no cancellation check either, and propagation through their awaited sequences cannot be verified without a live server, so they stay out too. Each is listed in the follow-up notes with what it needs.

Trino does gain the `onTermination` handler its stream was missing, which is a real defect on the export path regardless of the bounded read: without it a cancelled export left the producer running.

**The cancel is withheld inside a transaction block.** Cancelling a statement while `PQtransactionStatus` reports `PQTRANS_INTRANS` puts the transaction into the aborted state, and every later command fails until `ROLLBACK`. Browsing a large table inside a transaction the user opened would have thrown that transaction away. The bounded read drains instead, so it still skips decoding the tail but pays the transfer. PGlite already withheld the cancel for its own reason (`suppressServerSideCancel`), so it is not in the table above.

**Also not converted:** ClickHouse (its `streamRows` fires an extra `LIMIT 0` round trip to derive columns, so a bounded read would be two HTTP requests where today it is one, net-negative on small results until that is restructured); DuckDB (its stream materializes a `duckdb_result` first and has no `onTermination`; needs `duckdb_pending_prepared_streaming`); Cloudflare D1, LibSQL remote, Etcd, Elasticsearch, SurrealDB, Beancount, Dameng (one call returns the whole result, nothing to stop).

### Honest limit

The speedup is plan-dependent and the PR does not claim otherwise. Measured on the same SQL: `ORDER BY` with an index 3,415 ms → 5 ms; without an index 6,027 ms → 3,324 ms; `GROUP BY` 77 ms → 78 ms, no gain. A blocking plan node has to finish before it can emit row 1. What this always removes is the transfer and decode of every row past the cap, and the hidden drain charged to the next query.

### Not done, deliberately

No cursor-held pagination over arbitrary SQL. DBeaver does not do it either: `SQLQueryDataContainer.readData()` closes the statement and `ResultSet` per read, `RESULT_SET_MAX_ROWS_USE_SQL` and `RESULT_SET_USE_FETCH_SIZE` are both off by default, and `RESULT_SET_REREAD_ON_SCROLLING` defaults true, so "next page" re-runs from row 0 and is quadratic in reads. Holding a cursor would pin a connection and a snapshot across user idle time.

No SQL rewriting. TablePlus appends the limit to the user's text, which shipped as TablePlus#1683: a query ending in `-- where` ran as `-- where LIMIT 1000;` and returned the whole table. The dead `injectRowLimit` requirement stays in place per the ABI rule against removing a published requirement, and stays unused.

No UI change. `.partialLoad` already renders "Showing 1,000 rows" and `PaginationCoordinator.fetchAllRows` already offers Fetch All behind a confirm.

## Also fixed

Three defects found while investigating this subsystem. The first two are the same class as the reported bug and the fix is incomplete without them.

**The row cap only recognised a leading `SELECT` or `WITH`.** `QueryClassifier.leadingKeyword` takes the leading letter/digit run, so `(SELECT * FROM events) UNION ALL (…)` yielded `""` and was classified a *write*, while `TABLE big_table` and `VALUES (1),(2)` failed the keyword check. All three ran with no cap, no truncation badge and no Fetch All, and never reached the driver with a cap at all. `leadingKeyword` now reaches past opening parens. That can only make a classification equal or stricter: `(DROP TABLE t)` moves from `.write` to `.destructive`, and the `INTO` guard plus the body-wide `filesystemMarkers` scan are untouched.

**Sorting a query result appended `ORDER BY` after the user's `LIMIT`.** `SELECT * FROM orders LIMIT 100` plus a header click produced `… LIMIT 100 ORDER BY "total" ASC`, a syntax error on every engine, quoting SQL the user never typed. And when the query already had an `ORDER BY`, `stripTrailingOrderBy` took the `LIMIT` with it, so the user's own limit was silently replaced by the app's row cap. `SQLLimitDetector` now exposes the offset of the first top-level `LIMIT`/`OFFSET`/`FETCH` over its existing lexer, and `QuerySqlParser.applyingOrderBy` splices the clause ahead of it.

Two traps in that, both caught in review. `OFFSET` is non-reserved in MySQL and MariaDB, so a bare keyword match split `SELECT offset, name FROM events` at the column name; the clause is now recognised only when a row count, a placeholder, `ALL`, or `FIRST`/`NEXT` after `FETCH` follows it. And ANSI `OFFSET n ROWS FETCH NEXT m ROWS ONLY` is only legal alongside an `ORDER BY`, so clearing the sort has to take that tail with it or SQL Server rejects the statement; a standalone `LIMIT` still stays.

**Fetch All replayed the previous run's bound parameters.** `applyPhase1Result` set `baseQueryForMore` without clearing `baseQueryParameterValues`, and it did so on the exact branch that makes Fetch All reachable. Run a parameterized query, then a plain one that truncates, then Fetch All, and PostgreSQL rejects it with `bind message supplies 1 parameters, but prepared statement "" requires 0`. The pair now moves together through `PaginationState.setBaseQueryForMore`.

**A SQLite `UPDATE` or `DELETE` alone in a query tab reported no row count.** Its `executeUserQuery` override took the streaming branch for *every* unparameterized query, including DML, and hardcoded `rowsAffected: 0`. Measured with a compiled probe against a real SQLite file: `execute()` → 5000, `executeUserQuery()` → 0. The override now streams only a row-returning statement and sends everything else through `execute()` with its real `sqlite3_changes()`. Deleting the override outright, which was the first attempt, would have regressed MCP: `MCPConnectionBridge` resolves its own cap and never reaches `executeBoundedQuery`, so a capped MCP read of a large SQLite table would have gone back to materializing every row.

## Verification

| Step | Result |
|---|---|
| `verify.sh build` | PASS |
| `verify.sh test` (7 suites) | PASS, 141 cases |
| classifier suites after the security fix | PASS, 54 cases |
| `verify.sh lint TablePro Plugins TableProTests` | 0 violations |
| `verify.sh docs` | PASS |
| Per-plugin builds | 8 touched plugin schemes build; PostgreSQL builds |
| `verify.sh plugins` (AllPlugins) | fails inside the vendored `oracle-nio` checkout's `@TaskLocal` macro, a known local-toolchain issue unrelated to this change. No error in any file in this diff. CI covers it. |

New tests: `BoundedQueryTests` and `PluginBoundedStreamTests` (early abort proven by counting rows the producer actually produced; exact-cap vs truncated; a non-participating driver stays on the buffered path), plus cases in `QueryExecutorTests`, `QueryClassifierTests`, `SQLLimitDetectorTests` and `PaginationStateTests`.

No screenshots: nothing visual changed. The row cap affordance is the one that already shipped.

No UI automation: the sort-with-`LIMIT` flow needs a live connection and a column-header click, and per the notes in `CLAUDE.md` on this grid, XCUITest reads every row and cell as obscured and needs offset-point clicks. The rewrite is covered by six unit cases over `QuerySqlParser.applyingOrderBy` instead.

Codex was unavailable for the review pass (`Your workspace is out of credits`, not an authentication failure), so no second-lab model has read this diff. It went through `Skill(code-review)` and `Skill(security-review)` instead. That review is what removed the five drivers above, restored the SQLite override, and found both `applyingOrderBy` traps.

## Follow-ups this leaves open

Each is a driver that could bound its fetch once its stream can actually be stopped. None is a regression from this PR; all are pre-existing and also affect export cancellation today.

- **MSSQL** (`MSSQLPlugin.swift:599`) and **Oracle** (`OraclePlugin.swift:425`): hoist the nested `Task {}` so `onTermination` can cancel it, then give the core loop a real cancellation flag. `Task.isCancelled` alone will not do, for the same reason it does not work in libpq's stream: the loop body is synchronous on a `DispatchQueue` and carries no task context.
- **Cassandra** (`CassandraConnection.swift:688`): the paging loop needs a cancellation check between pages.
- **Snowflake**, **Trino**: confirm against a live server whether cancelling the stream task stops the fetch, then opt in.
- **PostgreSQL** (`LibPQPluginConnection.swift:664`): `streamQuery`'s own early abort is inert. `onTermination` does `queue.async` onto the same serial queue the producer occupies, so `cancelAndDrain` runs only after every row is read, and the per-row `Task.isCancelled` sits outside any task and is always false. The bounded read does not go through it, so this PR does not depend on it, but export and compare cancellation do. The fix is to reuse the `PluginQueryCancellationGate` the class already owns.
- **ClickHouse**: derive stream columns from the first `JSONEachRow` line instead of a `LIMIT 0` probe, then opt in.

https://claude.ai/code/session_016E9GaZVKhXdJS3hdrDzctX

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Query results stop fetching at the row limit on MySQL, PostgreSQL, SQLite, Redis, DynamoDB and BigQuery. (#2427)
+- Query results stop fetching at the row limit on MySQL, PostgreSQL, SQLite, MSSQL, Oracle, Cassandra, Redis, DynamoDB, BigQuery, Snowflake, Trino and ClickHouse. (#2427)
+- ClickHouse reads a capped result in one request instead of a separate `LIMIT 0` round trip for the columns.
 
 ### Fixed
 
 - Row limit ignored for a parenthesised `SELECT`, a `TABLE` statement or a `VALUES` list. (#2427)
+- Cancelling an export or a schema compare left the database still sending rows on MySQL, PostgreSQL, MSSQL, Oracle, Cassandra and Trino.
+- An aborted PostgreSQL read charged its leftover rows to the next query on that connection.
 - Syntax error when sorting a query result whose SQL ends in `LIMIT`.
 - The user's own `LIMIT` dropped when sorting replaced an existing `ORDER BY`.
 - Fetch All sending the previous run's parameters after a query that had none.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Query results stop fetching at the row limit on MySQL, PostgreSQL, SQLite, Redis, DynamoDB and BigQuery. (#2427)
+
 ### Fixed
 
+- Row limit ignored for a parenthesised `SELECT`, a `TABLE` statement or a `VALUES` list. (#2427)
+- Syntax error when sorting a query result whose SQL ends in `LIMIT`.
+- The user's own `LIMIT` dropped when sorting replaced an existing `ORDER BY`.
+- Fetch All sending the previous run's parameters after a query that had none.
+- A SQLite `UPDATE` or `DELETE` alone in a query tab reporting success instead of the rows affected.
 - Stall on switching between two loaded table tabs, growing with the rows they hold. (#2424)
 - Scroll position lost when switching away from a table tab and back. (#2424)
 - A table statistics command on every switch between two table tabs. (#2424)
--- a/Packages/TableProCore/Sources/TableProTrinoCore/TrinoStatementClient.swift
+++ b/Packages/TableProCore/Sources/TableProTrinoCore/TrinoStatementClient.swift
@@ -40,10 +40,13 @@ public final class TrinoStatementClient: @unchecked Sendable {
         )
     }
 
+    /// The paging loop runs in an unstructured task, so terminating the stream has to cancel it
+    /// explicitly. Without that the loop keeps fetching pages nobody reads, and its own
+    /// `abortIfCancelled` never fires because nothing ever cancels the task it runs in.
     public func executeStreamed(_ sql: String) -> AsyncThrowingStream<TrinoStreamElement, Error> {
         AsyncThrowingStream { continuation in
             let client = self
-            Task {
+            let statementTask = Task {
                 do {
                     _ = try await client.runStatement(
                         sql,
@@ -55,6 +58,7 @@ public final class TrinoStatementClient: @unchecked Sendable {
                     continuation.finish(throwing: error)
                 }
             }
+            continuation.onTermination = { @Sendable _ in statementTask.cancel() }
         }
     }
 
--- a/Plugins/BigQueryDriverPlugin/BigQueryPluginDriver.swift
+++ b/Plugins/BigQueryDriverPlugin/BigQueryPluginDriver.swift
@@ -651,6 +651,10 @@ internal final class BigQueryPluginDriver: PluginDatabaseDriver, @unchecked Send
 
     // MARK: - Streaming
 
+    func executeBoundedQuery(query: String, rowCap: Int) async throws -> PluginQueryResult? {
+        try await boundedQueryFromStream(query: query, rowCap: rowCap)
+    }
+
     func streamRows(query: String) -> AsyncThrowingStream<PluginStreamElement, Error> {
         AsyncThrowingStream(bufferingPolicy: .unbounded) { continuation in
             let streamTask = Task {
--- a/Plugins/CassandraDriverPlugin/CassandraConnection.swift
+++ b/Plugins/CassandraDriverPlugin/CassandraConnection.swift
@@ -685,13 +685,20 @@ actor CassandraConnectionActor {
         return "Unknown error"
     }
 
+    /// The whole loop is one non-suspending actor job holding the cooperative-pool thread, and
+    /// cass_future_wait blocks it, so nothing outside can interrupt a page. `abort` is the only
+    /// way the consumer's departure reaches this body, and it is read between pages and between
+    /// rows rather than mid-request.
     func streamQuery(
         _ cql: String,
+        abort: PluginStreamAbort,
         continuation: AsyncThrowingStream<PluginStreamElement, Error>.Continuation
     ) throws {
         guard let session else {
             throw CassandraPluginError.notConnected
         }
+
+        guard !abort.isAborted else { return }
 
         let pageSize: Int32 = 5_000
         let statement = cass_statement_new(cql, 0)
@@ -705,7 +712,10 @@ actor CassandraConnectionActor {
 
         defer { cass_statement_free(statement) }
 
+        var aborted = false
+
         while true {
+            if abort.isAborted { break }
             let future = cass_session_execute(session, statement)
             guard let future else {
                 throw CassandraPluginError.queryFailed("Failed to execute query")
@@ -755,7 +765,13 @@ actor CassandraConnectionActor {
             let iterator = cass_iterator_from_result(result)
 
             if let iterator {
+                var rowsThisPage = 0
                 while cass_iterator_next(iterator) == cass_true {
+                    rowsThisPage += 1
+                    if rowsThisPage % 256 == 0, abort.isAborted {
+                        aborted = true
+                        break
+                    }
                     let row = cass_iterator_get_row(iterator)
                     guard let row else { continue }
 
@@ -780,16 +796,16 @@ actor CassandraConnectionActor {
 
             let hasMore = cass_result_has_more_pages(result) == cass_true
 
-            if hasMore {
+            if hasMore, !aborted {
                 cass_statement_set_paging_state(statement, result)
             }
 
             cass_result_free(result)
 
-            if !hasMore { break }
+            if !hasMore || aborted { break }
         }
 
-        if !headerSent {
+        if !headerSent, !aborted {
             continuation.yield(.header(PluginStreamHeader(
                 columns: [],
                 columnTypeNames: [],
--- a/Plugins/CassandraDriverPlugin/CassandraPlugin.swift
+++ b/Plugins/CassandraDriverPlugin/CassandraPlugin.swift
@@ -257,21 +257,36 @@ internal final class CassandraPluginDriver: PluginDatabaseDriver, @unchecked Sen
 
     // MARK: - Streaming
 
+    func executeBoundedQuery(query: String, rowCap: Int) async throws -> PluginQueryResult? {
+        guard let bounded = try await boundedQueryFromStream(query: query, rowCap: rowCap) as PluginQueryResult?
+        else {
+            return nil
+        }
+        /// The buffered path reports a read's row count here, so a bounded read reports the same
+        /// rather than the collector's neutral zero.
+        return PluginQueryResult(
+            columns: bounded.columns,
+            columnTypeNames: bounded.columnTypeNames,
+            rows: bounded.rows,
+            rowsAffected: bounded.rows.count,
+            executionTime: bounded.executionTime,
+            isTruncated: bounded.isTruncated,
+            statusMessage: bounded.statusMessage
+        )
+    }
+
     func streamRows(query: String) -> AsyncThrowingStream<PluginStreamElement, Error> {
         let cql = stripTrailingSemicolon(query)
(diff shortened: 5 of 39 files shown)


datlechin:
## Follow-up commit: row stream producers now stop when their consumer stops

`2bc934b` closes every follow-up this PR originally deferred, plus one the investigation found that was not on the list.

### The defect, measured

Terminating an `AsyncThrowingStream` runs its `onTermination` handler and nothing else. Every affected driver hopped that handler onto the same serial queue its producer already occupied, so the drain sat behind the whole read, and the per-row guard was `Task.isCancelled` inside a non-async `DispatchQueue` closure, where it is always false.

A compiled Swift probe reproducing the exact shape (producer on a serial queue, 40 batches at 20ms, consumer breaking after 2):

| | consumer breaks | producer stops | drain runs |
|---|---|---|---|
| Before | 0.025s | never, 200,000 of 200,000 | **0.938s**, after the entire read |
| After | 0.025s | **10,000 of 200,000** | 0.048s, one batch later |

That probe also settled a trap worth writing down: **`onTermination` fires when the stream value is released, not when the consumer breaks.** Two earlier probe runs "passed" while testing nothing because the harness held the stream in scope. It is why `PluginBoundedStream.collect` is safe (it returns immediately after breaking) and it is now a regression test.

### The mechanism

`PluginStreamAbort` plus `PluginRowStream.make` in PluginKit. The handler is installed before the producer builds, so a producer scheduled *after* termination still sees the flag and declines to send its query at all. `onAbort` lets a driver register task cancellation without clobbering the stream's own handler, which is a footgun the first cut had and which my own ClickHouse code walked straight into.

Still additive: `scripts/check-pluginkit-abi.sh` reports zero removals and zero signature changes across both commits, so no `currentPluginKitVersion` bump.

### Drivers

| Driver | Change |
|---|---|
| PostgreSQL, CockroachDB, Redshift, PGlite | Producer polls the abort and the cancellation gate; owns its own drain; `PQTRANS_INTRANS` guard |
| MySQL, MariaDB | **Not previously on the list.** Identical inert shape. Now polls, and `killQueryOnServer` + drains on abort |
| Cassandra | Abort polled between pages and every 256 rows; opted in |
| MSSQL | Abort threaded into FreeTDS as a closure, since that file is compiled into the iOS app and may not import PluginKit; opted in |
| Oracle | The inner unstructured `Task` is now cancelled on abort; opted in |
| Trino | `TrinoStatementClient.executeStreamed` cancels its paging task, which makes its existing `abortIfCancelled` fire and issue the server-side `DELETE /v1/statement`; opted in |
| ClickHouse | New single-request bounded path; opted in |
| Snowflake | Already correct; opted in with no driver change |

### Two bugs in my own first cut, both caught in review

`PQtransactionStatus` was read **after** `PQsendQuery`, where libpq reports `PQTRANS_ACTIVE` until the result is fully drained, so the transaction guard was dead code. Confirmed with a C probe against a live PostgreSQL 17.11. And `PluginRowStream.make` was overwritable, as above.

### What is not verified

- **Oracle does not compile locally.** The vendored `oracle-nio` `@TaskLocal` macro fails on this toolchain, the same failure that breaks the `AllPlugins` aggregate. CI is the first real check of that change.
- **Nothing here is tested against a live server.** The abort mechanism is measured and unit-tested; the per-driver wiring is compile-verified and reasoned only. ClickHouse's `JSONCompactEachRowWithNamesAndTypes` positional decode is the one I would exercise first, which is why export and compare deliberately stay on the existing path.
- MSSQL, Oracle, Cassandra, Trino and Snowflake are registry-only, so they need their own plugin releases.
- Codex was unavailable throughout (`out of credits`), so no second-lab model has read either commit.

### Left open

`cancelCurrentQuery()` calls `PQgetCancel` off the serial queue and has no transaction guard, and `PQcancel` is a blocking connect with no timeout, so an unreachable server or tunnel stalls that queue. The vendored libpq 17.4 ships the non-blocking `PQcancelCreate`/`PQcancelStart`/`PQcancelPoll`, unused anywhere in the repo.

https://claude.ai/code/session_016E9GaZVKhXdJS3hdrDzctX

PR #2445, opened 2026-08-26
fix(editor): rerank incremental completions for the current prefix

### LINKED ISSUE #2444: Incremental autocomplete keeps stale candidate ordering after an exact match
## What happened?

When the autocomplete popup opens for a short prefix and the user continues typing, the visible list can keep the ranking from the earlier prefix. An exact keyword can remain below a longer function candidate, and the stale first row stays selected for Return or Tab.

For example, after the popup opens for `t`, completing the token to `true` can leave the numeric function `TRUNCATE(n, decimals)` above the exact keyword `TRUE`:

```sql
SELECT * FROM GT_USER
WHERE true
```

The same problem is reproducible with other candidate pairs, including:

- `n` -> `null`: `NULLIF()` can remain above `NULL`
- `i` -> `in`: `INSTR()` can remain above `IN`
- `i` -> `is`: `ISNULL()` can remain above `IS`
- `r` -> `regexp`: `REGEXP_REPLACE()` can remain above `REGEXP`

This is not specific to `TRUE`. It occurs whenever the correct order changes as the prefix grows while both candidates survive filtering.

## Steps to reproduce

1. Open a SQL query editor.
2. Enter a query ending in `WHERE t`.
3. Wait for the autocomplete popup to open for `t`.
4. Continue typing `rue` without closing the popup.
5. Observe that `TRUNCATE(n, decimals)` can remain above `TRUE` and stays selected.
6. Close the popup, remove the last character, and type it again so a fresh popup opens for the complete `true` prefix.
7. Observe that the fresh ranking places `TRUE` above `TRUNCATE`.

The `n` -> `null` and `i` -> `in` sequences provide additional reproductions.

## Expected behavior

The visible completion list should be filtered and ranked for the current full prefix on every incremental update. Exact matches such as `TRUE`, `NULL`, and `IN` should move ahead of longer prefix or fuzzy matches before the first row is selected.

## Confirmed cause

`QueryCompletionAdapter.completionOnCursorMove` immediately returns `synchronousRefilter`, which calls `service.filter` and preserves the previous order. `scheduleRefilter` calculates the correct ranking after a 30 ms debounce, but stores it only in `lastRefilterItems`; it does not update the visible `SuggestionViewModel.items`.

Relevant code:

- `TablePro/Views/Editor/QueryCompletionAdapter.swift`
- `TablePro/Core/Autocomplete/SQLCompletionService.swift`
- `TablePro/Core/Autocomplete/SQLCompletionProvider.swift`
- `LocalPackages/CodeEditSourceEditor/Sources/CodeEditSourceEditor/CodeSuggestion/Model/SuggestionViewModel.swift`

A fresh completion request already performs `getCandidates -> filterByPrefix -> rankResults`. The stale order is specific to the open-popup cursor-update path.

## Acceptance criteria

- Incremental completion updates return candidates ranked for the current prefix.
- `t` -> `true` places `TRUE` before `TRUNCATE()`.
- `n` -> `null` places `NULL` before `NULLIF()`.
- `i` -> `in` places `IN` before `INSTR()`.
- The first selected row matches the newly ranked first item.
- SQL and MongoDB completion behavior remains covered by targeted tests.

## Environment

- TablePro: 0.68.0
- macOS: 26.6.2
- Architecture: Apple Silicon
- Database type: SQL completion path; the ranking code is shared across SQL dialects

devy1540:
## Purpose

Fix the autocomplete popup keeping the candidate order from an earlier prefix, which can leave a longer function selected above an exact keyword match.

Closes #2444

## Changes and rationale

- Distinguish transient `seed` sessions from completed `resolved` sessions.
- Re-rank resolved candidates against the current full prefix on every cursor update so exact matches such as `TRUE`, `NULL`, and `IN` move to the first row.
- Keep seed sessions filter-only because saved keyword favorites are unbounded.
- Reuse the previous result set when the prefix grows, and fall back to the original session candidates after deletion or a non-prefix edit. This preserves exact SQL ranking without repeatedly sorting every MongoDB field candidate.
- Remove the delayed 30 ms ranking task and asynchronous cache that calculated the right order without updating the visible popup.
- Add regressions for `t → true`, `n → null`, and `i → in`, plus 1,000-candidate seed and resolved-session bounds.
- Record the fix under `[Unreleased] > Fixed`.

## Acceptance criteria

- [x] Confirmed that the new `TRUE/TRUNCATE`, `NULL/NULLIF`, and `IN/INSTR` ordering test fails before the production change.
- [x] `QueryCompletionAdapterLifecycleTests` passes after the fix.
- [x] The related `SQLCompletionProviderFuzzyDedupeTests`, `SQLCompletionAdapterFuzzyTests`, `SQLCompletionProviderTests`, `MongoContextAnalyzerTests`, and `QueryCompletionProfileRegistryTests` suites pass.
- [x] `swiftlint lint --strict TablePro/Views/Editor/QueryCompletionAdapter.swift TableProTests/Views/Editor/QueryCompletionAdapterLifecycleTests.swift` passes with zero violations.
- [x] `git diff --check` passes.
- [ ] Manual verification in the built app has not been run.

<details>
<summary>Repository template compatibility</summary>

## 목적

Fix the stale ordering of incrementally filtered autocomplete candidates.

## 내용(의도 포함)

Re-rank bounded resolved candidates while preserving filter-only seed sessions and monotonic narrowing for large candidate sets.

## 성공기준

- [x] Exact-match ordering regressions pass.
- [x] Seed and resolved-session candidate bounds are covered.
- [x] Related autocomplete tests, changed-file SwiftLint, and diff checks pass.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Data Rewind settings in Settings > Data & Results, with an off switch and Clear Saved Changes.
 - Restore Previous Values in the toolbar's Table Actions group.
 
+### Changed
+
+- Editor tabs drawn as a segmented tab picker rather than in Liquid Glass, on every macOS version. (#2439)
+
 ### Fixed
 
 - Autocomplete keeping an earlier prefix's ordering after the typed word becomes an exact match. (#2444)
@@ -29,13 +33,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A save reported as failed on an engine without transactions after some statements had been written.
 - Keep Open for a preview tab, by double-clicking it in the tab strip or from its contextual menu. (#2436)
 - Kafka driver plugin: topics in the grid, consumer group lag, and KafkaQL for seeking and producing. (#2419)
-
-### Changed
-
-- Editor tabs drawn as a segmented tab picker rather than in Liquid Glass, on every macOS version. (#2439)
-
-### Fixed
-
 - Active editor tab indistinguishable from the inactive ones in light appearance. (#2439)
 - Active editor tab drawn darker than its track on macOS 27, and inverting when the window lost focus.
 - Editor tab selection changing with the desktop picture behind the window.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Autocomplete keeping an earlier prefix's ordering after the typed word becomes an exact match. (#2444)
 - Save reporting the number of statements it ran as the number of rows it changed.
 - An edit or a delete on a table with no primary key changing every identical row. (#2107)
 - The same statements committed twice when Cmd+S is pressed again during a slow save.
--- a/TablePro/Core/Autocomplete/CompletionEngine.swift
+++ b/TablePro/Core/Autocomplete/CompletionEngine.swift
@@ -11,6 +11,9 @@ import TableProPluginKit
 /// Completion context returned by the engine
 struct CompletionContext {
     let items: [SQLCompletionItem]
+    /// The wider ranked pool an open popup re-ranks against as the prefix grows. `items` is the
+    /// slice of it the popup shows.
+    let candidates: [SQLCompletionItem]
     let replacementRange: NSRange
     let sqlContext: SQLContext
 }
@@ -95,6 +98,7 @@ final class CompletionEngine {
 
         return CompletionContext(
             items: context.items,
+            candidates: context.candidates,
             replacementRange: mappedRange,
             sqlContext: context.sqlContext
         )
@@ -128,7 +132,7 @@ final class CompletionEngine {
 
         let adjustedCursor = cursorPosition - windowOffset
 
-        let (items, context) = await provider.getCompletions(
+        let (items, candidates, context) = await provider.completionSession(
             text: analysisText,
             cursorPosition: adjustedCursor,
             forcedTableReferences: forcedTableReferences
@@ -164,6 +168,7 @@ final class CompletionEngine {
 
         return CompletionContext(
             items: items,
+            candidates: candidates,
             replacementRange: replacementRange,
             sqlContext: adjustedContext
         )
--- a/TablePro/Core/Autocomplete/Mongo/MongoCompletionService.swift
+++ b/TablePro/Core/Autocomplete/Mongo/MongoCompletionService.swift
@@ -8,6 +8,12 @@ final class MongoCompletionService: QueryCompletionService {
 
     private static let windowRadius = 5_000
 
+    /// How many candidates a session may keep for re-ranking as the prefix grows. Every position's
+    /// vocabulary is well under this except a collection's sampled field paths, which a wide
+    /// document schema can run into the thousands; those sessions prefer what the popup showed,
+    /// and the cap holds either way because filtering an empty opening prefix returns everything.
+    private static let sessionPoolLimit = 400
+
     init(schemaProvider: SQLSchemaProvider?, databaseType: DatabaseType?) {
         self.schemaProvider = schemaProvider
         _ = databaseType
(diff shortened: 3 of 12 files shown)


datlechin:
Please use english

devy1540:
@datlechin Sorry, I change english

datlechin:
Thanks for this. The diagnosis was right and I have pushed a commit on top rather than asking you to rework it, since a few things only showed up once I traced the ranking against a real dialect. Your commit stays as the first one.

## What I kept

Deleting the 30 ms `scheduleRefilter` task and its cache, and ranking on the keystroke instead. That was the bug, and your read of it was exact: the right order was computed and then thrown away.

Your tests too. Driving a real `TextViewController` through `EditorControllerFixture` is better than the seams I had written, so I dropped mine and moved everything onto yours.

## What changed, and why

**Seeded sessions rank as well.** Keeping them filter-only leaves the same bug in the window before the analyzed request lands: the seed list is in declaration order, and `DESCRIBE` sits ahead of `DESC` in the keyword table, so typing `desc` there picks the wrong one. The cost you were guarding against is real but small: `filterByPrefix` already fuzzy-matches every candidate on that path, and ranking only sorts the survivors. Measured on this Mac in Debug, `filterAndRank` is 36 us at 40 candidates and 507 us at 600. Instead of skipping the rank I bounded the pool, which is what the unbounded favorites actually needed.

**Ranking alone does not fix the reported case on PostgreSQL.** `getCompletions` truncated candidates to `maxSuggestions` (40 in a `WHERE` clause) before the adapter ever stored them. PostgreSQL declares 41 `T`-prefixed functions and the shared list adds more, and a function scores `300 - 500 + length` against `TRUE` at `400 - 500 + 4`, so every one of them outranks it. `TRUE` was cut at prefix `t` and typing `rue` re-ranked a set it was never in, which is your own before/after still failing. Snowflake is affected too; MySQL is not, which is why the tests pass without it.

A session now keeps ten times what the popup shows and re-ranks that, so a longer prefix can reach a candidate the opening one ranked out of view. The popup still shows `maxSuggestions`. The pool is bounded rather than whole because the cost is linear: 4 ms at 5,000 candidates and 16 ms at 20,000 would be a dropped frame on a wide schema.

**Each update re-filters the session's own candidates** instead of narrowing from the previous keystroke's. It is a superset, so nothing is lost, and deleting a character widens the list again, which the narrowing path could not do.

**`rankResults` scores once per candidate** rather than twice per comparison, and breaks ties on the candidate's generated position. `sorted(by:)` is documented as not stable, and the generator's order is meaningful.

**`filter` came off the `QueryCompletionService` protocol.** Leaving an order-preserving alternative next to `rank` is what let the wrong one be called; `rank` is now the only way to update an open session.

**MongoDB gained a tier for the completed token.** Its comparator went anchored match, then kind priority, so a shell method (300) outranked a keyword (400) the user had finished typing. Same defect class, five lines. Its session pool is capped too, because an empty opening prefix filters nothing away and a wide document schema samples thousands of field paths.

## Verification

Build passes. 224 completion cases pass across `QueryCompletionAdapterLifecycleTests`, `QueryCompletionRankingTests`, `SQLCompletionProviderConcurrencyTests`, `SQLCompletionProviderFuzzyDedupeTests`, `SQLCompletionProviderTests`, `SQLCompletionAdapterFuzzyTests`, `CompletionEngineTests`, `MongoContextAnalyzerTests` and `QueryCompletionProfileRegistryTests`. SwiftLint reports zero violations on the changed paths.

Your success criteria listed manual verification as not run, so I added `EditorAutocompleteFocusUITests.testTypingToAnExactKeywordCommitsThatKeyword`: it types `where t`, waits for the popup, types `rue` without closing it, presses Return and asserts the inserted text. It asserts the text rather than the panel's rows, because the panel is borderless and its rows are not reliably queryable. Passed four runs out of four.

I also folded the two `### Fixed` headings `[Unreleased]` had grown into one and put the sections in the order `CHANGELOG.md` asks for. Your entry is unchanged.

## Found while in here, not fixed

Three defects in the same subsystem, none of which this PR depends on:

- A session seeded before a request that returns nil is never cleared, and the popup then stops appearing entirely. Type `SELECT * FROM users WHERE ` with the trailing space, wait a second, then type `n`: no popup, and none for `nam` or `name` either, because `RENAME` keeps matching the seeded keywords. It recovers on a space, a non-matching prefix, or Escape. `QueryCompletionAdapter.openSession` returns nil without clearing the seed.
- `SQLSchemaProvider.allColumnsFromCachedTables` iterates a Dictionary, so which columns survive the 40-item cap differs on every launch. Typing `SELECT ` with no FROM offers a different set each time.
- `SQLSuggestionEntry.deprecated` tests every label against `MongoVocabulary.deprecatedCollectionMethods`, so a SQL column named `count`, `update`, `insert` or `remove` renders greyed out with a warning triangle, while the flag can never fire on a MongoDB suggestion, which is what it was written for.

Issue #2279, opened 2026-08-20
Add keyboard navigation between SQL statements in the editor


datlechin:
## What

Two commands to move the cursor between SQL statements in the editor: previous statement and next statement.

## Why

The result bar already has keyboard switching between result sets. The editor has no equivalent for statements, so working through several queries in one tab still means scrolling and clicking.

## Scope

- Two items in the Query menu (`QueryMenuBuilder`), both bindable under Settings > Keyboard.
- Suggested defaults: Ctrl+Cmd+Up and Ctrl+Cmd+Down. Check for conflicts before picking.
- Optional third command: run the statement under the cursor, then move to the next one.

## Not this

Jupyter's modal editing, where Esc enters a command mode and Enter goes back to editing. It is not a macOS pattern and it collides with the two-step Escape behaviour the editor already has.

Issue #2244, opened 2026-08-19
JSON result mode offers filters and row editing whose effects it never shows


datlechin:
Found while reworking the result-mode command routing in #2222. Both halves predate that PR; #2222 deliberately preserved the existing behaviour rather than changing it unannounced, so they are filed here instead.

JSON mode offers two sets of controls whose effect it never shows.

**The Filters toggle toggles a panel that does not exist there.** `ResultsViewMode.showsRowFilters` is `.data || .json`, so the funnel button appears in the status bar in JSON mode. But `FilterPanelView` renders only inside the `case .data:` arm of `MainEditorContentView.resultsSection` (line 661 on `main`). Clicking the toggle in JSON mode flips `filterState.isVisible` and draws nothing. The same is true of `Cmd+Option+F`.

**Row editing acts on rows JSON mode does not redraw.** Add Row, Duplicate Row, Delete and Paste are enabled in JSON mode, because `GridSelectionOwner.resolve(.table, .json)` is `.dataGrid` and the JSON view genuinely shows the data grid's rows and its selection. But `ResultsJsonView` renders `resolvedTableRows(for: tab)`, the loaded rows, with no pending-change overlay. So Add Row inserts a pending row that does not appear, Paste pastes rows that do not appear, and Delete marks rows that keep rendering unchanged. The Save button lights up and the user cannot see why.

Delete is the interesting one: selecting rows in Data mode, switching to JSON to check the payload, then pressing `Cmd+Delete` is a real flow, and it works. It just gives no feedback until you switch back.

**Two ways to settle it**, and it is a product call rather than a code one:

1. Make the effects visible. JSON mode reads the pending-change overlay the grid already has, and gains the filter panel. Costs the most, keeps every command meaningful.
2. Make JSON mode read-only. The toggle and the four row commands turn off there, and it becomes an inspection view. Costs the Data-then-JSON-then-Delete flow, and needs a CHANGELOG line because it removes something that works today.

**Verified on `main` at `edae33c`:** `showsRowFilters` at `QueryTab.swift:23`, `FilterPanelView` at `MainEditorContentView.swift:661` inside `case .data:` (line 623), `ResultsJsonView` construction at line 588 with no change-manager input.

PR #2453, opened 2026-08-26
feat(settings): make Find a rebindable shortcut so Cmd+F can open the filter bar

datlechin:
A user asked to use `Cmd+F` for the filter drawer instead of Find. This makes that possible without changing anyone's defaults.

## Root cause

Not a wrong default, a missing capability. `Find…` was the only command wired outside the shortcut-customization system:

- `EditMenuBuilder.swift:118-124` hardcoded `keyEquivalent: "f", modifiers: .command` and passed no `shortcut:` / `keyboard:`, unlike its siblings Find Next and Find Previous.
- `Cmd+F` sat in `ShortcutAction.reservedAppShortcuts`, and `reservedConflict(for:context:)` returns that for **every** context, so `ShortcutConflictResolver` refused to bind `Cmd+F` to anything at all.

So the request was blocked by a hard gate, not declined as a preference.

## Why the defaults do not move

Apple's HIG lists `Command-F / Open a Find window` as a standard shortcut and says "In general, don't repurpose standard keyboard shortcuts for custom actions". Measured on macOS, Mail and Xcode both give **filter** a non-F key (`Cmd+L`, `Cmd+Option+J`).

TablePlus binds `Cmd+F` to Open Row Filter and ships no grid find at all, which is where the request comes from. But TablePlus users are themselves asking for the opposite (TablePlus #2279, still open), Postico 2 ships `Cmd+Option+F` for the filter bar exactly as we do, and DataGrip, the only surveyed tool that has both a grid find and a filter, gives `Cmd+F` to Find.

This shortcut has already flip-flopped three times (`CHANGELOG.md:2164` → `:1538` → `:658`/`:213`), and the current arrangement is what the grid find bar with Search All Rows and the #2244 JSON-mode work were built for. A preference split between two camps is settled by rebinding, not by a fourth reversal.

## The change

- `ShortcutAction.find`, context `.global`, default `Cmd+F`. `.global` is explicit and commented, because only a global context overlaps `.dataGrid`, and that overlap is the entire reason the yield works.
- `Cmd+F` removed from `reservedAppShortcuts` and from `editorBuiltIns`.
- `Find…` now routes through `KeyboardSettings` like every other customizable command.
- The vendored `CodeEditSourceEditor` local key monitor drops its `case (commandKey, "f")`.

That last one is required, not cosmetic. The monitor runs before NSMenu key-equivalent matching and ignores `KeyboardSettings`, so without it Settings would show Find on another key while the editor still opened Find on `Cmd+F`. It is safe because `TextViewController+FindActions.swift` already declares `performFind` on the same controller, so the nil-targeted menu item reaches the identical `showFindPanel()` through the responder chain. Commit `f5afbe221` already added `Cmd+Shift+D` and `Cmd+Shift+K` to that same switch, so editing it is established practice here.

**One edit for the user.** `shortcut(for:)` already yields an action's default when another action's override claims that key in an overlapping context, so binding Toggle Filters to `Cmd+F` stands Find down on its own. Two menu items can never both hold `Cmd+F`, which is the AppKit-blanks-the-loser trap in `CLAUDE.md`.

## Fixed while here

Review caught that the recorder's **Reassign** button called `clearShortcut` unconditionally, persisting the cleared sentinel even when the losing action was still on its default. `KeyboardShortcutModels.swift:390-392` says explicitly that storing the stand-down is wrong: it makes it permanent, shows the action as customized, and its Reset arrow appears to work until the next launch re-applies the clear. In this feature's headline flow that meant a user who later moved the filter bar back to `Cmd+Option+F` would find Find permanently unbound. Reassign now clears only a loser that has an explicit override; one on its default is left to yield and recovers on its own.

## Verification

| Step | Result |
|---|---|
| `verify.sh build` | PASS |
| `verify.sh test` (14 menu + keyboard suites) | PASS, 92 executed, 92 passed |
| `verify.sh lint TablePro TableProTests LocalPackages/CodeEditSourceEditor` | 0 violations |
| `verify.sh docs` | PASS |

The `category` switch is exhaustive with no `default:`, so the new case had to be handled at every site to compile at all. `MainMenuShortcutCoverageTests.everyShortcutActionIsReachable` pins that `.find` reaches exactly one menu item.

Tests added: defaults still pin `.find` == `Cmd+F` and `.toggleFilters` == `Cmd+Option+F` (a guard against reversal four); `Cmd+F` is no longer reserved in any context; Find yields and exactly one menu item ends up holding `Cmd+F`; `Find…` carries no hardcoded key equivalent; and Reassign leaves Find recoverable.

No UI automation: the flow is Settings > Keyboard recorder input, and `ShortcutConflictResolver` consults `SystemHotkeyChecker`, which reads live System Settings through `CopySymbolicHotKeys`. Asserting on it would make the test depend on the machine, so the coverage sits on the deterministic app-owned layer instead.

## Notes for the reviewer

- `docs/features/keyboard-shortcuts.mdx:266` pictures the Settings > Keyboard list, which now has a Find row. The screenshot needs re-capturing; I did not fabricate one.
- Keyboard settings sync (`SyncCoordinator.swift:880`). A user running this build and an older one with sync on, who rebinds `Cmd+F`, gets a duplicate key equivalent on the **old** build, which still hardcodes `Cmd+F` for Find. `sanitized()` only drops bare keys, so it does not filter this. Not fixable from this side and it heals when the other build updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Restore Previous Values in the Edit menu, taking back a save that already committed. Starter license. (#2107)
 - Data Rewind settings in Settings > Data & Results, with an off switch and Clear Saved Changes.
 - Restore Previous Values in the toolbar's Table Actions group.
+- Rebindable Find shortcut in Settings > Keyboard, for giving `Cmd+F` to the filter bar instead.
 
 ### Changed
 
--- a/LocalPackages/CodeEditSourceEditor/Sources/CodeEditSourceEditor/Controller/TextViewController+Lifecycle.swift
+++ b/LocalPackages/CodeEditSourceEditor/Sources/CodeEditSourceEditor/Controller/TextViewController+Lifecycle.swift
@@ -259,10 +259,6 @@ extension TextViewController {
         case (commandKey, "]"):
             handleIndent()
             return nil
-        case (commandKey, "f"):
-            _ = self.textView.resignFirstResponder()
-            self.findViewController?.showFindPanel()
-            return nil
         case ([commandKey, .shift], "D"):
             duplicateLine()
             return nil
--- a/TablePro/Core/Menu/EditMenuBuilder.swift
+++ b/TablePro/Core/Menu/EditMenuBuilder.swift
@@ -119,8 +119,8 @@ enum EditMenuBuilder {
             MenuItemFactory.item(
                 String(localized: "Find…"),
                 action: #selector(MainSplitViewController.performFind(_:)),
-                keyEquivalent: "f",
-                modifiers: .command
+                shortcut: .find,
+                keyboard: keyboard
             ),
             MenuItemFactory.item(
                 String(localized: "Find Next"),
--- a/TablePro/Models/UI/KeyboardShortcutModels.swift
+++ b/TablePro/Models/UI/KeyboardShortcutModels.swift
@@ -82,6 +82,7 @@ enum ShortcutAction: String, Codable, CaseIterable, Identifiable {
     case nextStatement
     case runStatementAndAdvance
     case previewSQL
+    case find
     case findNext
     case findPrevious
     case aiExplainQuery
@@ -149,7 +150,7 @@ enum ShortcutAction: String, Codable, CaseIterable, Identifiable {
              .executeQueryWithoutLimit, .cancelQuery, .explainQuery, .formatQuery,
              .foldAll, .unfoldAll, .toggleFold,
              .previousStatement, .nextStatement, .runStatementAndAdvance,
-             .previewSQL, .findNext, .findPrevious, .aiExplainQuery, .aiOptimizeQuery:
+             .previewSQL, .find, .findNext, .findPrevious, .aiExplainQuery, .aiOptimizeQuery:
             return .editor
         case .undo, .redo, .cut, .copy, .copyRowsExplicit, .copyWithHeaders, .copyAsJson,
              .paste, .delete, .selectAll, .clearSelection, .addRow, .duplicateRow,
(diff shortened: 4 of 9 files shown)


datlechin:
Follow-up after re-reading the Codex job records properly (the earlier run's findings were in the job result, not stdout, and I had missed them).

**Codex adversarial-review returned `No-ship` on the first commit**, and it was right:

> a supported Settings sequence can still create two Cmd+F menu claimants

Reproduction: assign `Cmd+F` to Find Next (accepting the conflict against Find), then assign `Cmd+F` to Toggle Filters. The second assignment was accepted, because `findNext` was classified `.editor` and `toggleFilters` is `.dataGrid`, so `findConflict` treated them as non-overlapping. Both are nil-targeted menu key equivalents, so AppKit would blank one.

That context split was wrong on its own terms: `MainSplitViewController.findNext` and `findPrevious` branch on `hasActiveGridFind` and route to the grid **or** the editor, exactly like `performFind`. Fixed in 62612a650 by classifying all three find commands `.global`, which is the same argument the first commit used for `.find`.

This also closes a pre-existing hole the review surfaced: binding a data-grid action to `Cmd+G` previously raised no conflict against Find Next and silently cost the user a shortcut. There is now a regression test for it.

Tests: 96 executed, 96 passed. Build and lint clean.

Two review points I did not act on, with reasons:

- **Package Example loses `Cmd+F`** (P2). Confirmed real: `LocalPackages/CodeEditSourceEditor/Example` has no menu Find and no `performFind` responder, so the deleted monitor case was its only route. Keeping the case is not an option, because it matches the literal `"f"` character rather than the user's configured key, so it fires before the menu whatever the user binds and defeats the whole feature. The Example still has its Show/Hide Find button. Worth raising upstream separately.
- **Comment style** (P3). Trimmed to three lines stating the invariant. Doc comments of this kind are the established convention in this file (see `ShortcutContext.overlaps` and `KeyboardSettings.shortcut(for:)`); the rule targets behavior-describing inline comments.

PR #2547, opened 2026-08-27
fix(connections): synchronize background connection pane state

### LINKED ISSUE #2545: Background connections keep showing ‘Preparing the session’ after they have connected
## Summary

When TablePro reopens multiple connections from the previous session, a connection that finishes connecting in the background can keep showing the “Preparing the session” screen after it is already connected.

The workspace rail and toolbar report that the connection is connected, including the server version, but the detail pane still shows:

- “Preparing the session”
- “This is taking longer than usual.”
- A Cancel button

Clicking Cancel and then Connect makes the content appear immediately because the database session was already connected. The second Connect does not need to establish a new session.

## Environment

- TablePro: 0.68.1 (build 124)
- Platform: macOS
- Observed with: MySQL 8.0.x through Cloud SQL Auth Proxy
- Also present in the relevant connection and pane code on the current `upstream/main`

The affected state-management path is not specific to MySQL or Cloud SQL. Those were the connection types used during reproduction.

## Steps to reproduce

1. Open multiple saved database connections in one TablePro window.
2. Configure TablePro to reopen the last session at launch.
3. Leave one connection selected.
4. Quit and reopen TablePro.
5. Wait until the workspace rail reports that the restored connections are connected.
6. Select a connection that completed while another connection was selected.

## Actual behavior

The selected connection has contradictory UI state:

- The workspace rail reports `Connected`.
- The toolbar displays the connection and database server version.
- The sidebar remains visible, which indicates that the logical window pane is content.
- The detail pane still displays `Preparing the session`.
- The detail pane still has a Cancel button.
- The stale preparation screen remains indefinitely.

Clicking Cancel and then Connect displays the connected workspace immediately.

## Expected behavior

When a background connection finishes connecting, its cached panes should represent the final connected phase.

Selecting that connection later should display its content immediately, without showing a stale connection progress screen or a Cancel button.

## Root cause

Each connection owns persistent `NSHostingController` instances through `WorkspacePanes`.

During connection reconciliation, `MainSplitViewController.reconcileStatus(of:)` currently performs the connected transition in this order:

```swift
adoptSession(session, into: workspace)
refreshPanes(of: workspace)
transition(to: nextPhase, for: connectionId)
```

Relevant code:

- `TablePro/Core/Services/Infrastructure/MainSplitViewController.swift`, `reconcileStatus(of:)`
- `TablePro/Core/Services/Infrastructure/MainSplitViewController.swift`, `transition(to:for:)`
- `TablePro/Core/Services/Infrastructure/MainSplitViewController.swift`, `refreshPanes(of:)`
- `TablePro/Core/Services/Infrastructure/MainSplitViewController.swift`, `applySelectedWorkspace()`
- `TablePro/Core/Services/Infrastructure/WorkspacePaneHost.swift`

For a background workspace completing its initial connection:

1. `nextPhase` is resolved as `.connected`.
2. `adoptSession` installs the connected session and coordinator.
3. `refreshPanes` runs while `workspace.phase` is still `.connecting`.
4. `buildDetailView(for:)` therefore assigns `ConnectingStateView` to the workspace’s detail hosting controller.
5. `transition(to: .connected)` updates the workspace phase.
6. Because the workspace is not selected, `transition(to:for:)` does not call `applyPhase()` and does not rebuild the cached pane.
7. Selecting the workspace later only swaps in the existing hosting controller through `showSelectedPanes()`. It does not refresh the controller’s root view.

The resulting state is:

```text
session.status = connected
workspace.phase = connected
logical pane = content
cached detail root = ConnectingStateView
```

A later status event may not repair the pane because `reconcileStatus(of:)` returns early when the session contents and phase are already current.

## Why Cancel and Connect appear to fix it

Cancel changes the selected workspace phase and immediately rebuilds its pane.

The following Connect call finds an existing driver and returns early from `ensureConnected()`. The selected workspace then runs `applyPhase()` and rebuilds the detail pane from the connected state.

This repairs the stale UI without establishing a new database connection.

## Additional affected transitions

The same cached-pane invariant can be violated by other background phase transitions:

| Background transition | Possible cached pane after transition | Required pane |
|---|---|---|
| Initial connection succeeds | Connecting | Content |
| Connection attempt fails | Connecting | Unavailable |
| Connected session disappears | Empty or stale content | Unavailable |
| Automatic reconnect starts | Content | Connecting |
| Automatic reconnect succeeds | Connecting | Content |

The fix should cover the shared transition invariant instead of only moving one `refreshPanes` call in the initial-success branch.

## Suggested fix direction

Ensure that a background workspace’s cached panes are rendered from the final phase.

The state update order should be:

1. Adopt or release the session state.
2. Update `workspace.phase`.
3. Let the selected workspace use the existing `applyPhase()` path.
4. Refresh cached panes for a background workspace after its phase has been updated.

Refreshing every workspace unconditionally during selection would hide the symptom, but it may defeat the design of preserving one long-lived pane tree per connection. The root fix should keep the cached pane synchronized when the phase changes.

## Acceptance criteria

- [ ] A connection that completes in the background displays content immediately when selected.
- [ ] The detail pane never shows `ConnectingStateView` when the live session is connected and the logical pane is content.
- [ ] A background connection failure displays the unavailable state when selected.
- [ ] A background session loss displays the disconnected state when selected.
- [ ] A background reconnect displays the connecting state and later returns to content.
- [ ] Switching connections continues to preserve editor, grid, scroll, selection, and undo state.
- [ ] Repeated connected-status events do not leave or recreate a stale progress pane.
- [ ] Regression tests cover the final phase and the actual cached detail render state together.

## Test coverage gap

Existing tests cover these pieces separately:

- `ConnectionWindowPhaseMachine`
- `ConnectionWorkspaceRegistry`
- `ConnectionWindowPaneResolver`
- `WorkspacePanes` teardown

They do not cover this sequence as one test:

```text
Workspace A selected
Workspace B connecting in the background
Workspace B receives a connected session
Workspace B is reconciled
Workspace B is selected later
Workspace B’s cached detail root is inspected
```

A regression test should assert both:

```text
workspace.phase == connected
cached detail render state == content
```

Checking only the phase or pane resolver output will not detect this bug.

## Regression source

The ordering was introduced with the per-connection pane ownership refactor:

- `486458a40c04b539738df84266911fed358f36a1`
- `refactor(connections): give each connection its own panes instead of rebuilding on switch (#2116)`

devy1540:
## Purpose

Closes #2545

Fix the case where a workspace that finishes connecting in the background remains stuck on the `Preparing the session` screen when selected later.

## Changes and intent

- Refresh the persistent panes owned by a background workspace after updating it to the final connection phase.
- Keep the existing `applyPhase()` path for the selected workspace.
- Preserve the normal workspace-switching behavior, which swaps existing hosting controllers without rebuilding editor, grid, scroll, selection, or undo state.
- Add an AppKit and SwiftUI regression test that mounts two workspaces in a real `NSWindow`, completes the background connection, selects it, and verifies that `MainContentView.onAppear` activates the coordinator and installs command actions.
- Record the fix in the Unreleased Fixed section of `CHANGELOG.md`.

## Success criteria

- [x] Confirmed that the regression test failed before the production change at `coordinator.isActivated` and `commandActions`.
- [x] Confirmed that the same regression test passes after the fix.
- [x] Passed 52 adjacent tests across six suites covering connection phases, pane resolution, workspace registries and containers, and editor teardown.
- [x] Passed all 13,126 `TableProTests` in 1,517 suites with `-testLanguage en -testRegion US`.
- [x] Passed the macOS Debug build.
- [x] Passed SwiftLint with zero violations on the changed Swift files.
- [x] Passed `audit-refactor-health` and `git diff --check`.
- [x] Completed independent quality, security, UI contract, challenge, and synthesis reviews with no blocking findings.

A full run under the local Korean system locale exposed 153 existing assertions that compare localized output directly with English strings. The same full test target passed when the test language and region were pinned to English. The focused regression test and adjacent suites also passed independently.


datlechin:
Thanks for the report and the diagnosis, both were exact. I pushed a commit on top that keeps your fix's behaviour and moves it to the ownership boundary, plus a second bug the investigation turned up in the same place.

## Why more than the one line

`refreshPanes(of:)` in the unselected branch does fix all five transitions. It works by repainting a second time after the first paint drew the wrong phase, so `reconcileStatus` still paints before it writes the phase. Two things fall out of leaving that in place.

A workspace can reach the window with no phase change to follow at all. `adoptWorkspace` builds one, resolves `.connected` when the session already has a driver, and inserts it; `startActivationConnectIfNeeded` returns early because `allowsActivationConnect(.connected)` is false, so no transition ever fires and nothing assigns a `rootView`. The panes keep the empty view from `WorkspacePanes.init`, and `reconcileStatus` returns at its `alreadyRendered` guard on every later status event, so it never heals. Opening a second connection into a window that is already on screen shows a blank sidebar, grid and inspector. That is reachable through `WindowManager.openTab`, so through the MCP server, a deeplink, or the connection switcher.

And a session can be replaced under an unchanged phase. A database switch on a reconnect-based engine hits `reconcileStatus` with `nextPhase == workspace.phase == .connected`, which `transition`'s `phase != next` guard skips.

## What the commit does

Nothing paints before the phase is final. `reconcileStatus` adopts or releases and stops there, `releaseSession` no longer paints, and `transition(to:for:)` is the single sink: it writes the phase, then syncs the panes of the workspace it names, selected or not, and on an unchanged phase too. Only the window's own chrome stays gated on the phase actually moving.

`WorkspacePanes` records a `WorkspacePaneRenderKey` of what its four root views were last built from: the resolved pane, the connection record, and a session revision that `ConnectionWorkspace` bumps whenever its session is replaced. It holds a revision rather than the session so a released driver is not kept alive by the record. `refreshPanes` is the only writer.

That key is what lets `applySelectedWorkspace` sync on every switch, which is what fixes the blank-pane case. Without it that call has to be unconditional, and then every connection switch re-evaluates `MainContentView` and `SidebarView`, which is the work the per-connection panes exist to avoid.

## Measured, because the fix depends on it

Probes built with `swiftc` against the macOS 14 target, since none of this is documented:

- A `rootView` write on a detached hosting controller is deferred, not lost. The last value written is what mounts when the pane is re-parented, and `@State`, `NSTableView` scroll offset and selection, and `NSTextView` undo all survive it.
- Every pane builder erases one `_ConditionalContent`, so the type inside the `AnyView` is stable and only the branch changes. 30 identical-branch rewrites on a mounted pane cost 0 view creations and 0 teardowns. 8 background workspaces cost 0.0141ms.
- Two writes in one run-loop turn coalesce to the last one. Split across turns, the intermediate mounts and tears down. So `layoutSubtreeIfNeeded` must stay out of the repaint path; `WorkspacePanes.teardown()` needs it only because its panes are never parented again.

`CLAUDE.md` said the opposite of the first and third points, and `WorkspacePanes` blamed `AnyView` for the old cost. Both are corrected in the commit, and the workspace-phase invariant now records this rule.

## Tests

Your test is kept and the suite is extended to six cases, each asserting the final phase and the cached render state together: background connect succeeds, connect fails, session lost, reconnect, repeated connected events, and a workspace adopted with a live session.

Reverting only the sync in `transition` fails 5 of the 6; with it, 6 of 6 pass. 82 cases pass across the nine neighbouring suites (`ConnectionWorkspaceRegistry`, `ConnectionWorkspaceContainers`, `ConnectionWindowPaneResolver`, `ConnectionWindowPhaseMachine`, `MainSplitViewControllerDetailWidth`, `SplitPaneHoldingPriority`, `WindowTitleResolver`, `EditorLifecycleTeardown`, and the pane synchronization suite). macOS Debug build passes, SwiftLint reports 0 violations on the changed files.

An adversarial review caught two regressions in my own first draft, both fixed here:

- Narrowing the adoption guard to content equivalence alone dropped the phase condition, so a recovered tunnel would have left the workspace holding the driver the recovery had just disconnected, and that driver's cached password. `isContentViewEquivalent` answers whether the views draw the same, not whether it is the same session. The reconnect test now asserts the replacement driver is adopted.
- Routing the connection-record change through the key skipped the repaint when a connection is deleted while its session is still open, which purges the per-connection registries without changing the record. That path keeps its unconditional repaint, and it is the one place that skips the key.

No screenshots: the bug needs two connections and a background connect completing while another is selected, which is not something a still frame shows. The regression suite is the evidence instead.

Issue #2393, opened 2026-08-24
Font in detail setion of the right sidebar should refers to the editor font settings


tyrival:
### Problem

The font in the detail section of the right sidebar of the data panel, as well as in the input box and the multi-line input box, currently only the JSON format refers to the editor font settings; all should refer to the font settings and size.

### Related database type

None

PR #2580, opened 2026-08-29
fix(datagrid): value-filter popover value font and row alignment

shreeve:
## Problem

The column value-filter popover renders its distinct values in the system font instead of the Data Grid Font, and its value rows sit at a different left edge than the Select All row above them, so the two checkbox columns do not line up.

## Root cause

The value `Text` carried no font modifier, so it fell back to the system face. That is the #2393 defect class: it looks right only while the user leaves the Data Grid Font at the default, so it ships invisibly. The alignment gap comes from two owners of the left edge: the Select All row is padded `.horizontal, 14` while the value rows keep the plain `List` style's default row insets.

## Fix

- The value label resolves `ThemeEngine.shared.valueFontSwiftUI`, the same way every other value-rendering popover does. The occurrence count keeps `.callout.monospacedDigit()`; it is metadata, not a stored value.
- One private `horizontalInset` constant now owns the left edge, consumed by the header, the controls row and the value rows' `.listRowInsets`, so the columns cannot drift apart again. Same pattern `ColumnVisibilityPopover` uses for its rows.

## Tests

`ColumnValueFilterPopover.swift` added to `ValueFontTests.standaloneValueViewsResolveTheValueFont`, the source-scanning guard that fails when a value-rendering view stops resolving the value font. Suite passes (7 tests). The alignment itself is a shared-constant layout fact with no deterministic UI-test surface; the popover needs a live connection and a loaded column to open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - MongoDB `.sort()` and `.projection()` silently ignored when written with unquoted keys.
 - Compare & Sync unable to drop an overloaded PostgreSQL routine, or any trigger.
 - PostgreSQL sequence DDL naming the schema it was read from, in SQL export and the structure editor.
+- Value-filter popover listing values in the system font, with rows indented differently than Select All.
 
 ## [0.69.0] - 2026-08-27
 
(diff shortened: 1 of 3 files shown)


datlechin:
Could you take the screenshot of screen before and after changes?

shreeve:
Thanks for your review. I stopped using TablePro and I built a new DuckDB viewer in Rust with GPUI called DuckTable at:

https://github.com/shreeve/ducktable

PR #2600, opened 2026-09-01
fix(csv): keep document opening on the main actor

salmonumbrella:
## What changed

- Keep CSV and TSV document opening on AppKit's main actor by inheriting `NSDocument`'s default serial-read behavior.
- Add a regression test covering both advertised CSV and TSV document types.
- Record the macOS 27 crash fix in the changelog.

## Why

Opening a CSV or TSV on macOS 27 can terminate TablePro with `EXC_BREAKPOINT` / `SIGTRAP`, with `_dispatch_assert_queue_fail` and `_swift_task_checkIsolatedSwift` reached from the `NSDocumentController` opening queue. Under Swift 6, `NSDocument` is main-actor isolated, but `CSVDocument` opted into concurrent reads, allowing AppKit to open it on a background queue. Removing that override restores the framework default and keeps document construction and reading on the actor AppKit requires. This matches the underlying AppKit concurrency issue tracked in [Apple Feedback #38](https://github.com/1024jp/AppleFeedback/issues/38).

## Usage

No usage change.

## Verification

- `git diff --check origin/main..HEAD`: pass.
- Roborev reviewed the final commit; the correctness fix and regression coverage were accepted.
- Xcode build, test execution, and manual macOS 27 verification were not run because this work was completed on Linux. CI or a macOS checkout still needs to open a sanitized CSV and TSV before merge.

Serial document opening can make existing up-front row indexing visible on the main actor for very large files. Moving that indexing behind an async boundary is separate work; this PR restores actor correctness without widening scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A failed whole-schema trigger read counting as a schema with no triggers.
 - Compare & Sync publishing one pair's results after the pickers moved to another.
 - Clicks and hovers in the scrolled tab strip's edge padding landing on a tab clipped off the edge.
+- Crash when opening CSV and TSV files on macOS 27.
 
 ## [0.70.0] - 2026-09-01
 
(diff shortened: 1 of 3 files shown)


datlechin:
Cherry-picked onto `main` as fb2cea074 (authorship preserved) and shipped in v0.71.0. The only conflict was the CHANGELOG entry, which moved into the 0.71.0 `Fixed` section as `(#2600)`.

Verified on macOS since the PR noted it was authored on Linux: project regenerated, `CSVDocumentConcurrencyTests` passes for both `public.comma-separated-values-text` and `public.tab-separated-values-text`.

Closing as merged manually. Thanks for the fix.

https://claude.ai/code/session_01J6xU4Zx4DRJ5JaxMP437uT

PR #2636, opened 2026-09-04
feat(editor): chart each EXPLAIN tree node against the largest step in the plan (#2633)

### LINKED ISSUE #2633: Add a simple bar chart on EXPLAIN visualization, tree view
### Problem

A simple bar chart on EXPLAIN visualization, tree view would be very helpful for a quick comparison between nodes.

Thank you.

### Proposed solution

Current tree view on TablePro:

Image

Implementation on Tabularis:

Image

### Alternatives considered

_No response_

### Related database type

_No response_

datlechin:
Adds the bar chart #2633 asked for: a sortable magnitude column in the EXPLAIN tree, with a **Metric** menu above it. Each row's bar is drawn against the largest step in the plan, so the expensive node stands out without reading eight-digit numbers down a column.

Three defects had to be fixed first, because the bar is drawn from the numbers they get wrong.

## Before / After

The tree before this change is the one in the issue: `Operation`, `Cost`, `Rows` and `Actual Time`, all plain right-aligned text, with nothing on a row proportional to any value.

After, against a real PostgreSQL 17 `EXPLAIN (ANALYZE, FORMAT JSON)`:

EXPLAIN tree with the Self Time bar column

The `Seq Scan` fills the track and reads red at 21,513 ms of a 29,292 ms query. `Gather Merge` and `Hash Join` are visibly short. `Limit` at 0,019 ms still draws a hairline rather than nothing, so "tiny" stays distinguishable from "not reported".

Both `docs/images/explain-tree.png` and its dark twin are re-captured, since they pictured the old column set. They are 1512x861 because the display attached here is 1x; the originals were 3026x1720, so they are accurate but half resolution and worth re-shooting on a Retina display.

## The bar

Length is `value / largest node`, linear. Colour comes from `QueryPlanSeverity`, from the share of the plan for an additive metric and from the largest node for a row count, because rows are re-reported by every ancestor and a "share" of them means nothing. A non-zero value never rounds away to an empty track.

The metric menu offers only what the plan reports, and the column is titled by the metric it charts:

| Metric | What it is |
|---|---|
| Self Cost | `exclusiveCost` |
| Self Time | measured time, corrected for loops and workers |
| Estimated Rows / Actual Rows | rows out, with actuals corrected for loops |

Measured plans default to Self Time, estimated ones to Self Cost. The stored choice is never overwritten by a plan that cannot offer it, so opening a SQLite plan does not lose the metric picked for PostgreSQL.

`Canvas` draws it, which is what the plan diagram beside it already uses. Not a progress control: the HIG calls progress indicators transient and says a stationary one reads as a stalled app. Not `Gauge`: `.linear` is `@available(macOS, unavailable)`. Not Swift Charts: the vectorized `BarPlot` path is macOS 15. `NSLevelIndicator.relevancy` is the style the HIG names for comparing items, but a compiled probe against the real framework showed it draws no fill at any value and ignores `fillColor`, so it has no channel for severity.

## The three fixes underneath

**`costFraction` could exceed 1, and the badge inverted.** It divided a node's exclusive cost by the *root's* total, and PostgreSQL prices a `Limit` far below its child on purpose. Replayed against a live PostgreSQL 17: `ORDER BY id LIMIT 2` gave the Gather Merge **5.626**, a merge join gave an Index Scan **11.828**, against a doc comment promising 0.0-1.0 and a `QueryPlanSeverity` with no arm above `.critical`. The denominator is now the sum of every node's exclusive cost, which is what a share means, always lands in 0...1 and is identical wherever the old value was already in range. Node by node across five real plans, the LIMIT plan changes 2 of its 4 badges (a `Seq Scan` doing 11% of the work stops claiming to be critical) and the other four change 0 of 18.

**A plan that reports nothing claimed every node was cheap.** `costFraction` defaulted to 0 and 0 is `.low`, so SQLite, ClickHouse, DuckDB and Dameng plans painted a green "Low cost" dot on every row and said so to VoiceOver. `severity` is now optional: no cost, no badge, no claim. The glyph's footprint is reserved rather than removed, because a plan can price some nodes and not others and the operation text has to keep one left edge.

**Columns the plan cannot fill are hidden.** A SQLite plan used to show three permanently empty columns; it now lists its steps and nothing else, and a CockroachDB plan keeps `Rows` while dropping `Cost` and `Actual Time`. Driven by `isHidden` and written in both directions on every update: measured, `removeTableColumn` then `addTableColumn` strands the column at the end of the header and resets the width the user chose, while hiding preserves both.

The outline's `autosaveName` is bumped to `.v2`. Measured: with a saved four-column layout, AppKit restores what it knows in the saved order and appends anything it has never seen, so the bar column would have arrived stranded past `Actual Time` for everyone who had ever resized a plan column. The cost is a one-time reset of plan-outline column widths.

## Loops and workers

PostgreSQL reports actual time and rows as per-execution averages. Measured here: an inner Index Scan at `Actual Loops: 500` reports `Actual Total Time: 0.000`, and under a Gather with 2 workers launched a Seq Scan reports `Actual Loops: 3` against an execution time of 83 ms. So Self Time multiplies by loops and divides by the worker count, and the divisor belongs to a Gather's **descendants**, not the Gather itself, which runs once in the leader. Applying it to the Gather too reported 40.23 ms for a node measured at 120.697 ms. With it right, the self times of a plan sum to its execution time (120.701 against 120.728), which is the invariant that makes Self Time additive and its share meaningful.

**The `Actual Time` column is untouched and still shows the per-loop figure.** That is documented behaviour, pinned by `MySQLPlanParserTests`, and what psql, pgAdmin and DBeaver show. Self Time is a separate, differently named metric.

## Also fixed, found while investigating

- **`Workers Launched: 0` was hidden.** `visibleProperties` dropped any `"0"`, so a Gather reporting `Workers Planned: 2, Workers Launched: 0` showed the 2 and swallowed the 0, which is the answer to why the query was slow. Confirmed against a server with `max_parallel_workers = 0`. A small set of keys whose zero is the answer is now exempt.
- **MySQL priced the wrapper block above every table in it.** A table got only `cost_info.read_cost`, so the `Query Block` node absorbed every table's `eval_cost` as its own exclusive cost and came out the reddest node in every MySQL plan. It now takes `read_cost + eval_cost`, and only when `eval_cost` is present, so a plan without it reads exactly as before. Validated on MariaDB 12.3, where the per-table costs sum exactly to the block cost (0.0923548 + 0.9166336 = 1.0089884); **not** validated against a real MySQL server, which was not available here and for which MySQL publishes no `cost_info` example, so that arm follows its documented cost model rather than a measurement.

## Verification

- `build` PASS, `test` **116 executed, 116 passed** across 14 suites, `swiftlint --strict` clean on the app and test paths, `docs` PASS.
- New: `QueryPlanMetricIndexTests`, `QueryPlanLoopCorrectionTests`, whose fixtures are verbatim PostgreSQL 17 output rather than invented numbers.
- Updated with the API change in the same commit: `QueryPlanCostTests`, `QueryPlanSeverityTests`, `QueryPlanNodeSummaryTests`, `QueryPlanOutlineSortTests`, `PostgreSQLPlanParserTests`, `MySQLPlanParserTests`.
- UI automation covers the deterministic branch only. The sample database is SQLite, which reports no metric at all, so `testAMetriclessPlanShowsNoEmptyColumnsAndNoMetricChooser` pins that the chooser is absent and the empty columns are gone. **The populated bar cannot be UI-tested without a server**, so it was verified by driving a Debug build against a local PostgreSQL 17 instead, which is where the screenshots come from.
- Codex was unavailable for the second-model review: it reported `You've hit your usage limit ... try again at Sep 7th`. `Skill(code-review)` ran in its place.

Fixes #2633

https://claude.ai/code/session_015jGG9vt2GwiTxNk1FHKBCJ


datlechin:
Screenshots of the new column, captured by driving a Debug build against a local PostgreSQL 17. The docs images in the diff are the same two shots.

They are 1512x861 rather than the 3026x1720 the other `docs/images` files carry, because the display attached here is 1x. Accurate, but worth re-shooting on a Retina display before they ship.

PR #2650, opened 2026-09-06
feat(datagrid): image and SVG preview for cell values

### LINKED ISSUE #2535: SVG viewer for binary and text cells
### Problem

The cell viewers cover JSON, PHP serialized data and binary, but not SVG. A column holding SVG markup, as text or as bytes, shows the source and nothing else, so there is no way to see what the image is without copying it into another app.

### Proposed solution

- Render SVG in the cell viewer when the value parses as SVG, with a source and a rendered tab.
- Detect from the content, not the column type, since SVG turns up in both text and blob columns.
- Render in a sandbox with no network access and no scripting. An SVG is an untrusted document and must not be able to fetch anything.
- The same treatment for the other image formats already reachable as bytes, so a binary cell holding a PNG shows the PNG.

### Alternatives considered

Copy the value into a file and open it.

### Related database type

N/A / General

datlechin:
A cell holding SVG markup showed its source and nothing else, and a BLOB holding a PNG showed a hex dump. Both now open on the picture, with the stored form one segment away.

## What changed

`CellValueContentDetector` gains an `.image(CellImageFormat)` case and a `PluginCellValue` entry point, so the data grid and the row inspector resolve the same answer for the same value. Raster formats are identified by `CGImageSourceGetType` against an explicit UTI allowlist, from a 1 KB header read; SVG by scanning its XML prologue for a root `svg` element. `NSImage(data:) != nil` is deliberately not the test, because `NSImage.imageTypes` also covers PDF, DICOM, Photoshop and every camera raw format.

Three hosts share one `CellImageViewer`: the blob popover (`Image | Hex`), a new popover for SVG in a text column (`Image | Source`, source still editable where the cell is), and the row inspector. `Copy Image` writes through `ClipboardService`, and the pop-out reuses `ValueViewerWindowController`.

Raster decodes in process through ImageIO: the stored dimensions come from the header without decoding a pixel, and the preview is a bounded thumbnail. Measured, a 20000x20000 PNG reports its size in a millisecond with nothing allocated, and thumbnails to 1024² in 0.94s.

## Why SVG renders in a separate process

`NSImage(data:)` is the only SVG renderer macOS exposes, and it cannot be trusted with a value from a database.

- **A 147-byte SVG segfaults the process.** `<use href="#a">` inside `<g id="a">` recurses 120,654 frames through `SVGReader::resolveDefinitions` / `SVGNode::findDefinitionNode` and blows the 8 MB main-thread stack. The crash report is `EXC_BAD_ACCESS (SIGSEGV)`, `KERN_PROTECTION_FAILURE`, "Could not determine thread index for stack guard region", inside `-[_NSSVGImageRep initWithData:]` → `CGSVGDocumentCreateFromData`. Five of twelve recursion shapes crash: `use` self, mutual, `xlink:href`, inside `<defs>`, and `<symbol>`. Self-referencing `pattern`, `mask`, `clipPath`, `marker`, `filter` and `linearGradient` are all handled safely.
- **A cycle check is not enough.** A cycle-free 1,149-byte document (`<g id="gN"><use href="#gN-1"/><use href="#gN-1"/></g>` chained 20 deep) takes 5.5 seconds to decode. At 40 levels, 2,189 bytes, it never finishes. No input cap and no reference-graph validator makes it safe.
- It is a stack overflow, so no `try`, no larger thread stack and no cooperative timeout recovers from it.

So `tablepro-imagerender`, a `type: tool` target embedded at `Contents/MacOS/` exactly like the existing `mcp-server`, reads the document on stdin and writes a PNG on stdout. The app kills it on a three-second deadline; the tool installs a `sigaltstack` guard so a stack overflow exits cleanly instead of writing a crash report. Measured end to end: a benign document renders in 66 ms warm and 530 ms on the first spawn, every bomb above is contained, the app always survives, and no crash reports are produced.

The sandbox the issue asks for comes free with it. Measured against a local HTTP listener while rasterizing, `NSImage`'s SVG path issues **zero** network requests for an external `<image href>`, `xlink:href`, a CSS `@import`, an XXE external entity, an external `<use>`, `<script>`/`onload`, and a `<foreignObject>` HTML ``. `file://` and relative references are not loaded either: they draw the red broken-image placeholder. Inline `data:` URIs do render.

QuickLook was the alternative and was rejected: it also contains the crash, but the generator that answers is resolved by UTI across every registered third-party QuickLook extension, so the no-network property would not be ours to guarantee; it needs untrusted bytes on disk with a `.svg` extension; and an abandoned request keeps working inside the shared `quicklookd`.

## The allowlist has to be authoritative, not advisory

A prologue scan reads an XML comment, and DICOM carries a 128-byte free-form preamble, so a whole DICOM fits inside a comment ahead of an `<svg>` root and reads as SVG. `NSImage(data:)` then declines it as a vector, falls through to `NSBitmapImageRep`, and runs ImageIO's DICOM decoder over fully attacker-controlled bytes: exactly the decoder the raster allowlist exists to exclude. Built and measured a 4,494-byte polyglot that rasterizes at the 64x64 its own DICOM tags declare.

`SvgDocumentGate` is the answer, compiled into both the app and the renderer so they cannot disagree: a real vector document is invisible to ImageIO, so anything ImageIO claims is refused. Measured, plain SVG and SVG behind a byte order mark, an XML declaration, a doctype and a comment all pass, while the polyglot and a plain DICOM are refused at both the sniffer and the renderer. `CellImageSnifferTests` builds the polyglot and pins it.

## Scope

Nothing renders in the grid cell itself: a cell keeps its compact hex, so the drawn-cell, column-window and accessibility invariants are untouched. No new `ValueDisplayFormat` case either, since it is a persisted per-column override whose storage drops the whole per-table dictionary on an unknown rawValue, and detection here is per value. `Display As > Raw Value` still turns detection off.

## Verification

- `verify.sh build` PASS
- `verify.sh test` PASS over every suite that owns a type this touches
- `swiftlint lint --strict` clean over `TablePro`, `TableProTests`, `TableProUITests`
- `docs/scripts/check-writing-style.sh` and `check-docs-against-source.py` both clean
- The probes behind every measurement above were compiled and run against this machine's toolchain

## Review

Codex was out of credits for the whole of this branch (`You've hit your usage limit ... try again at Sep 7th`), so the second-model pass did not happen. `Skill(security-review)` and `Skill(simplify)` ran instead, and between them found the DICOM polyglot above, a dead view left by the hex split, an SVG-stored-as-bytes case that showed hex in the popover and markup in the inspector, a per-field ImageIO call on every text value in the row inspector, and an undrained stderr pipe on the helper. All are fixed here.

Fixes #2535

https://claude.ai/code/session_01EkZFosuu6fvVgdVmNaUFt9


datlechin:
Two gaps this branch could not close on the machine it was built on, both environmental rather than a property of the change.

**Screenshots.** `docs/images/cell-image-viewer{,-dark}.png` are committed as placeholders at the house 1560x960, so the page renders and no link is broken, but they are flat colour, not captures. The docs `<Frame>` needs real light and dark shots of the popover before this page ships. Capturing means driving a Debug build with `osascript`, and the developer's own `/Applications/TablePro.app` has been running throughout; addressing the wrong instance would send keystrokes into an app holding live database connections.

**`CellImageViewerUITests` is written but never executed.** XCUITest cannot start while another TablePro is running, and one was, for the whole session. Its typing sequence deliberately never types the closing `'`, because `BracketPairs.allValues` auto-closes quotes (the same workaround `QueryFormatUnterminatedLiteralUITests` documents), but that path is unverified. Treat the first CI run of that suite as the real check.

Everything else was verified locally: build PASS, 151/151 across every suite owning a changed type plus 69/69 on the final pass, `swiftlint --strict` clean, and both docs scripts clean.

PR #2656, opened 2026-09-06
refactor(datagrid)!: split the trailing pane into an inspector and an assistant, each with its own command

datlechin:
The trailing pane had no single subject. One `NSSplitViewItem` hosted a SwiftUI view that multiplexed six surfaces (row details, row-as-JSON, AI chat, table info, expanded JSON, expanded PHP) behind a three-way segmented picker, so nothing could title it, nothing owned its keyboard, and all of its state was bundled into one non-`Equatable` struct rebuilt wholesale on a 50ms debounce.

## What changed

**Two surfaces, two commands.** The assistant leaves the inspector's tab strip and becomes its own surface with `View > Show Assistant` (`Cmd+Option+A`) beside the existing `View > Show Inspector` (`Cmd+Option+I`, unchanged). Both share one `NSSplitViewItem`: toggling the surface on screen collapses the pane, toggling the other swaps and reveals it. The inspector keeps a segmented control, now for **Fields** and **JSON**, which are two renderings of the same row.

**Adaptive field rows.** Short scalars get a compact label-leading / value-trailing row; long text, JSON, PHP, binary and images keep a full-width editor and can expand in place. Roughly three times as many fields fit the pane.

**The affordances exist.** The NULL / DEFAULT / EMPTY / SQL-function menu is drawn on every field instead of only inside a hover overlay. `Tab` and `Shift+Tab` move between fields, `Ctrl+Option+N` and `Ctrl+Option+D` set NULL and DEFAULT. Field search and an edited-only filter are always present.

**The two full-panel takeovers are gone**, along with their hand-rolled `chevron.left / Fields` back button. A structured value grows in place with the rest of the row still around it, and the existing pop-out windows take anything larger.

## Defects fixed

Seven, all inside the replaced code, each verified with a reproduction:

- **Pickers snapped back to NULL.** `BooleanPickerView` and `EnumPickerView` computed `isNullValue = originalValue == nil`, and `originalValue` never changes on edit. On a NULL column, picking a value recorded the edit and showed the modified dot while the control still read NULL. Every field of a freshly inserted row was in that state.
- **The same expression made a multi-row selection read NULL** when none of the rows was NULL. `hasMultipleValues` was tracked and `placeholderText` handled it, but the pickers consulted neither.
- **Two fonts in the structure inspector.** `inheritedValueFont` and `editorMinHeight` were `default`-armed switches that disagreed with the exhaustive editor switch, and `.typePicker` took the picker branch, so Name and Type rendered in different fonts (the #2393 shape).
- **Primary key, foreign key and modified markers were silent to VoiceOver** (bare `Image` and `Circle` with no label).
- **Table Info showed the last table's statistics** on any tab with no table, including after that tab was closed.
- **Auto-show inspector on row select did nothing on query tabs**, contradicting both docs pages describing it.
- **The inspector ignored per-column display formats.** Its values came from a raw pass that decoded bytes with `isoLatin1`; the dead `selectedRowDataForSidebar` beside it computed formatted values nothing displayed.

Plus: **AI conversation history was read off disk on every connection window.** `MainContentView.onAppear` forced the lazily-built `AIChatViewModel`, whose `init` calls `loadConversations()`, regardless of the AI setting and whether the assistant was ever opened.

## Architecture

`InspectorContext` splits into `RowInspectorContext` and `AssistantContext`, both `Equatable`. `RightPanelState` splits into `RowInspectorState` and `AssistantState` under a `TrailingPaneState` owner. The untyped `[(column:value:type:)]` tuple path is deleted; `MultiRowEditState` stays the single row model, keeping its name because it is still genuinely multi-row. `RightPanelState.onSave` is removed: it was set and called only by `MainContentCommandActions` and no view ever read it. The assistant becomes a fourth `NSHostingController` in `WorkspacePanes`, which is where `sizingOptions = []` is applied and nowhere else (#1872).

Two new pure, tested types carry the decisions that had no home: `FieldValueState` (what a field is showing, resolved once instead of by each editor) and `InspectorFieldLayout` (inline against stacked).

## Measurements behind the design

- `List` is lazy and **stays lazy with variable row heights**: 336 subviews and ~135ms to first display at 100, 500 and 2000 rows, with the real row (editors, menu, badges). `Form(.formStyle(.grouped))` is eager, four subviews per row, 525ms at 2000. The field list is a `List`; `Form` is used only for the fixed-size Table Info pane.
- Swapping the hosted child of an inspector split item **preserves the pane width exactly** (380 to 380).
- Per-surface `minimumThickness` was measured and rejected: raising it force-grows the pane past the user's width and starves the content pane to its own minimum. One 270pt floor for both.
- The key view loop inside a SwiftUI `List` has **exactly one stop**, so Tab moved between fields not at all. `@FocusState` plus `.onKeyPress(.tab)` (macOS 14.0+) is what delivers it.

## Migration

The per-connection key `com.TablePro.rightPanel.activeTab.<uuid>` conflated the surface with the view mode. It migrates once: `AI Chat` seeds the assistant, `JSON` seeds the inspector on JSON, `Details` seeds the inspector on fields.

**A note for existing users:** AppKit never adds a new top-level item to an already-autosaved toolbar, which this repo has hit twice (`subitemAddRow`, `subitemRestorePreviousValues`). The Assistant toolbar button therefore ships in the default set for new users, and everyone else reaches the assistant from **View > Show Assistant** and its shortcut, or adds the button from Customize Toolbar. The docs say so.

## Verified

- `build` PASS, `lint` 0 violations across `TablePro`, `TableProTests`, `TableProUITests`
- Unit tests PASS: 167 cases across the new suites plus every suite owning a type this touches
- `docs` PASS on both checks, including keyboard-shortcut and menu-path claims against source

New suites: `FieldValueStateTests` (the picker regressions), `InspectorFieldLayoutTests`, `InspectorSubjectTests`, `TrailingPaneStateTests` (including migration), `TrailingPaneSurfaceTests`, `ShortcutUniquenessTests` (no such guard existed, and this adds a shortcut).

`InspectorFieldAffordanceUITests` covers the always-visible field menu, the field search and the header with no mouse events over a field. **It was not run locally**: XCUITest cannot start while another TablePro is running and the developer's own copy was open, so CI runs it.

## Before / after

Screenshots below show the pane's empty state, which is where the information-architecture change reads: the `Details | JSON | AI Chat` picker and its untitled pane become a titled inspector with the assistant on its own toolbar button.

A populated field list could not be captured: the data grid draws its cells with CoreText and takes no synthetic pointer events, so scripted input cannot select a row. `docs/images/cell-inspector.png` and `row-details-inspector.png` are still the pre-existing "Screenshot coming soon" placeholder cards and need real captures.

https://claude.ai/code/session_01PTe76cZbDyb13MQGjUj8r1


datlechin:
## Before / After

The pane's empty state, at the same 362x861 crop. Look at the top of the pane and the trailing toolbar.

**Before:** a `Details | JSON | AI Chat` segmented picker, an untitled pane, "No Selection", one toolbar button.

**After:** the pane is titled after what it inspects (`Track` / `Table`), the picker is gone, and the toolbar carries an Assistant button beside the inspector toggle.

Screenshot files are at the paths below on the authoring machine; drag them in to attach:

- `inspector-before.png`
- `inspector-after.png`

A populated field list could not be captured: the data grid draws its cells with CoreText and takes no synthetic pointer events, so scripted input cannot select a row. `InspectorFieldAffordanceUITests` covers the affordances instead, and CI runs it.

datlechin:
## Codex review: 14 findings, all addressed

A Codex working-tree review ran to completion (2h11m) and returned 14 findings, 4 at P1. Every one is fixed in `2583630c0`. Two were breaking CI in ways my own test selection had missed.

### P1

- **Structured editors were disabled for nearly every column.** Feeding the display format into `FieldEditorResolver` forwarded `.raw`, which the resolver reads as "skip JSON, PHP and image detection" — and `.raw` is also what the resolver returns for *no format at all*. Absence now stays `nil`; `.raw` is reserved for a deliberate Raw choice.
- **`ValueFontTests` scanned deleted paths.** It reads source files by string, so grepping for a type name never found it. Retargeted to `RowInspector/InspectorFieldRow.swift` and the moved editors; `CLAUDE.md:157` updated to name the new owner.
- **`MainWindowToolbarInspectorPlacementTests` required exactly `[.flexibleSpace]`** between the tracking separator and the Inspector toggle. `MainWindowToolbarLayoutTests.swift` holds five suite types and `-only-testing` matches the *type*, so my earlier run passed a different suite. The contract now asserts what actually matters: one flexible space immediately after the separator, and nothing but pane toggles after it.
- **The new UI test used a fixed 60pt offset.** On the 1024x768 runner the object browser overlaps the grid, so the click missed the row. Now uses the shared `gridPoint(in:of:dy:)` helper, and selects the row before revealing the inspector.

### P2

- The field list is mounted and hidden on a JSON switch rather than destroyed, so its search term, filter, expansion, focus and scroll position survive (the old pane did this deliberately).
- Expand is offered only where the editor actually resizes, and never behind a pending NULL/DEFAULT pill.
- Expand no longer *shrinks* an editor dragged taller than 320pt.
- The edited-only filter is reachable for structure rows, whose edits are committed rather than pending.
- The assistant's context is seeded when it is first activated, not left nil until an unrelated event.
- The Assistant toolbar item follows the connection's liveness, as its View-menu command already did.
- Copy Value on a field marked NULL copies the stored value rather than an empty string.

### Verified after the fixes

`build` PASS · `lint` 0 violations, 292 doc references clean · **211 unit cases, 211 passed**, including `ValueFontTests` and `MainWindowToolbarInspectorPlacementTests`.

Three findings I judged as design notes rather than defects and did not change: window-scoped `ChromePaneLayout` surface capture, re-mounting the trailing host on an `AppEvents.aiSettingsChanged` subscription, and deriving JSON edit normalization from the effective override. Each is a real edge case; none is reachable without a specific multi-workspace or settings-flip sequence, and each is noted here so a reviewer can push back.

PR #2659, opened 2026-09-06
fix(datagrid): restore the inspector's edit guards and keep each connection's trailing surface its own

datlechin:
Follow-up to #2656. Two independent adversarial reviews ran against that branch after it merged: a Codex `adversarial-review` (verdict: **do not ship**, three high findings, two of them paths to unintended database edits) and a six-lens agent audit that adjudicated the four findings I had deferred on judgement rather than evidence. They converged on the same defects. Every one is fixed here, and the two I deferred that turned out to be real are among them.

## Unintended database edits

**A hidden inspector field stayed keyboard-live.** Switching to JSON hid the field list with opacity, hit-testing and accessibility, but not `.disabled`. The merge base had that fourth modifier and the rewrite dropped it. Neither opacity nor `allowsHitTesting(false)` blocks the keyboard, so a field still holding first responder went on accepting text and the list's `Ctrl+Option+N` / `Ctrl+Option+D` behind the JSON view. Each of those stages a real cell change that a later Save commits, unseen.

**PHP-serialized fields became editable.** The merge base passed `isReadOnly: !isEditable || isPhpField`; the rewrite passed `isReadOnly: !isEditable`. TablePro cannot round-trip a PHP-serialized payload without PHP, so the field is shown and never mutated. Losing the guard exposed Set NULL, Set DEFAULT, Set EMPTY and the SQL functions on it, each of which overwrites the payload. The policy is now one function, `isFieldEditable(_:kind:rowIsEditable:)`, read by the editor, the value menu and the keyboard shortcuts alike, because the shortcut path checked only `isServerOwned`.

## Per-connection surface ownership

**Chrome restore wrote one connection's surface onto another, and persisted it.** `userPaneLayout` is window-scoped, but the reveal wrote its captured surface into `trailingPaneState`, which resolves to whichever workspace is selected *by then*, and `applySelectedWorkspace()` mounts the incoming workspace before applying chrome. Reachable with a single connection: open the Assistant, disconnect (which nils the state, so the capture falls back to Inspector), reconnect (which correctly reloads Assistant from defaults), and the restore immediately overwrites it. `TrailingPaneState.surface.didSet` then persists the foreign value under that connection's id. `trailingSurface` is removed from `ChromePaneLayout` entirely: the window records geometry, the workspace owns the surface, and the reveal mounts what the selected workspace already has.

**Adopting a session never re-parented the trailing pane.** `showSelectedTrailingPane()` had three callers and none was on the adoption path, so a connection restoring a remembered Assistant kept the Inspector that `viewDidLoad` mounted before its state existed, while every command reported the Assistant as active. `transition(to:for:)` now mounts the surface for the selected workspace once the phase is final.

**Disabling AI left the Assistant parented and interactive.** Every visibility question re-resolves the stored surface against the setting on each read, but nothing subscribed to `AppEvents.aiSettingsChanged` (one subscriber app-wide, in `SQLEditorCoordinator`). The panel this replaced normalized its own tab on that event. The controller subscribes now and reconciles the hosted child and toolbar validation; the stored choice is left alone, so turning AI back on returns the Assistant.

## Regressions from the first PR's own fixes

- **Auto-show inspector on row select evicted the Assistant.** Widening that setting to every grid selection (a fix in #2656) meant clicking a row swapped the Assistant away mid-question and persisted Inspector as the surface. It now reveals the inspector only when the Assistant is not the visible surface.
- **Tab was consumed unconditionally**, so focus could never leave the field list for the search field, the filter or the view-mode control. It is handled while another field is reachable and ignored at either end.
- **The type badge was focus-gated on inline rows**, and a read-only result can never take focus, so the column type was never shown at all there. Always drawn now.
- **Explain with AI and Optimize with AI validated enabled and did nothing** with AI off. Both gated on the setting.

## Verified

`build` PASS · `swiftlint --strict` 0 violations · **222 unit cases, 222 passed** · both docs checks PASS.

New suite `InspectorEditPolicyTests` pins the two edit guards, which is what neither had. `ChromePaneLayoutTests` drops the surface it no longer carries.

Two UI tests are also fixed here. `JSONRowInspectorUITests` failed on #2656's CI: its `openRowAsJSON` used a fixed 60pt grid offset, which lands on the object browser on the 1024x768 runner, and matched `window.menus.firstMatch`, which is no longer reliably the contextual menu now that every inspector field draws its own value menu. Both now use the shared `gridPoint` helper and match the item across the window's menus.

## A note on how #2656 shipped

Its macOS Tests run was cancelled by the next push, so unit and UI tests never completed for it. The PR's own run finished afterwards: unit tests passed, and UI shards 0/3 and 1/3 failed on the two suites fixed above plus my two new ones. This branch is what makes that run green.

https://claude.ai/code/session_01PTe76cZbDyb13MQGjUj8r1


datlechin:
## Added: three shipped docs claims the code does not deliver

A completeness pass over the merged diff found three sentences on the live docs site that describe behaviour the code does not have. Two of them became wrong *because* of a fix in this branch's predecessor, which narrowed the code without narrowing the page.

- **`json-viewer.mdx`** said the expand button works on "a JSON, PHP, text or **binary** field". `FieldEditorContent.canExpand` returns false for `.blobHex` and `.image`, so binary and image fields have no expand button at all. The page now says so.
- **`data-grid.mdx`** said long text, JSON, PHP, binary and images "take the full width **and can be dragged taller**". Only `MultiLineEditorView` and `JsonEditorView` use `ResizableEditorContainer`; PHP is expand-only, `BlobHexEditorView` is `maxHeight: 120` and `ImageFieldView` is a fixed `height: 220`. The page now names which two resize.
- **`data-grid.mdx`** said search filters "by column name or value". The predicate matches `field.originalValue`, so a field the user has just edited does not match its new text. Now "stored value".

`FieldEditorContent.canExpand` and `FieldPickerSentinel` were both switches that decided a visible control and that no suite owned. New `FieldExpansionPolicyTests` covers the expansion policy, including that a field showing a NULL or DEFAULT pill has no editor to grow while a *stored* NULL still does.

Both docs checks pass, `swiftlint --strict` is clean, and the new suite is 22/22.

### On how this was found

The six-lens audit I ran against this work was itself partly broken: its prompt told each lens to review `git diff origin/main...HEAD`, but by the time they ran, #2656 had been squash-merged and the worktree sat on `main`, so that command returned nothing. The lenses that ran it literally reviewed an empty diff. The findings that did land came from agents reading the files directly, and the completeness critic — which audited `f0ac68fb6` explicitly — is what caught both the broken premise and these three doc claims.

PR #2668, opened 2026-09-07
refactor(toolbar): rebuild the main window toolbar around native grouped items

datlechin:
## The problem, measured

The toolbar carried 17 controls plus a hosted SwiftUI status item, and drew the same facts twice.
Measured on a Debug build (macOS 26.2, PostgreSQL) by resizing the window and reading the
accessibility tree:

| Window width | What survived |
|---|---|
| 1512pt | everything, inspector flush to the trailing edge |
| 1200pt | Back/Forward, the window title, the centred item, `»`, Assistant, Inspector. **Every command was already in the overflow menu.** |
| 1000pt | the sidebar control, the window title, `»`, Assistant, Inspector. The centred item was dropped too, despite `visibilityPriority = .high`. |

Drawn twice at the same instant: `NSWindow.title` read `TablePro` while a capsule 200pt away also
read `TablePro`; `NSWindow.subtitle` read `tablepro_license · public` while a chip beside it read
`tablepro_license › public`; and a cylinder button and a cylinder chip opened the same chooser
through two entry points with two different enablement gates.

## Root cause

The toolbar was used as a status surface as well as a command surface, and the status half was
two `NSHostingController`-backed items. A view-backed item is opaque to AppKit:

- `NSToolbarItem.h` on `validate()`: "Since items with custom views don't always have meaningful
  target/actions, they do nothing." It is never validated.
- `NSToolbarItemGroup.h`: "Properties that get set on the parent toolbar item, such as label or
  view, apply to the entire item." The connection group's own subitems were inert, so neither
  reached the overflow menu, the customization palette or validation.
- `NSToolbar.h`: centred items are placed "assuming space allows", and `visibilityPriority` only
  ranks items already eligible to overflow. `.high` is a hint the header words as "suggest".

So AppKit could not shrink the centre and could only drop it whole, and the width it held while
still visible is what pushed seven real commands into the overflow menu at 1200pt.

## What changed

**The centre is a control again, and native.** A `NSToolbarItemGroup` of two titled items: the
connection (engine glyph plus name) and the container (cylinder plus database), each opening its
own chooser. That is the shape Apple documents for `principal` placement, "the location field for a
web browser is a principal item", and the shape Xcode ships, measured through its accessibility
tree as a list of role `path` holding Active Scheme and Active Run Destination.

Measured on a compiled AppKit harness: a **native** centred group is still fully visible at 700pt
where the hosted one was dropped at 1000pt, because AppKit can compress a group it draws itself.
Being centred was never the problem.

**Icon-only, and grouped by job.** Five capsules instead of a run of singletons: sidebar,
Back/Forward, the centred pair, Table Actions plus Editor, and the pane toggles. Nine controls in
the default set, against Finder's measured 8 and Xcode's 13.

**Commands outrank the centre.** The two names in the centre have no length limit, so every command
still went to overflow at 1200pt until the commands were raised to `.high`. That ordering is what
the HIG describes: trailing items "remain visible at all window sizes", centre items "automatically
collapse into the system-managed overflow menu".

| Window width | Now |
|---|---|
| 1500pt | sidebar, Back/Forward, centred pair, Table Actions, Editor, Safe Mode, Assistant, Inspector |
| 1200pt | every command, plus Assistant and Inspector. **The centre yields first**, into the overflow menu |
| 1000pt | sidebar, Back/Forward, Table Actions, Assistant, Inspector, `»` |

**Nothing is drawn twice.** The window title names the tab; the centre names the connection and the
container. The subtitle is gone, which is what `principal` documentation means by "takes precedent
over a title".

**Relocations**, each to a surface that already existed:

- Query duration, the running spinner and Stop moved to the results status bar, beside the rows
  they describe. The editor tab strip gained a per-tab spinner, so a query running in a background
  tab is still visible, which the old window-wide indicator could not tell you.
- Safe Mode became a native `NSMenuToolbarItem` whose glyph follows the level, plus
  **Database > Safe Mode**.
- Snowflake's warehouse and role became **Database > Session Context**. They previously existed
  only inside the hosted connection group, with no overflow entry, no menu command and no shortcut,
  so a window narrow enough to clip that group could not change either.
- Sidebar Tables/Favorites gained **View > Show Tables** and **View > Show Favorites**; the
  segmented control had been the only route.
- **Database > Schema** gained Open Schema Switcher…, which is the only route to the searchable
  chooser for the inner scope.

**Deleted**: the hosted item builders and their hosting-controller dictionary, `ToolbarSubject`,
`ConnectionStatusView`, `TableProToolbarView`, `SafeModeBadgeView`, the four SwiftUI subject-button
wrappers, `ConnectionScope` and its resolver, and seven `ConnectionToolbarState` fields with no
readers. Net 1,086 insertions against 1,605 deletions.

## Defects fixed on the way

- The container chooser opened over a session the health monitor had given up on: the centred chip
  was gated only on the engine's capability while the button beside it and the menu command were
  both correctly disabled. There is now one entry point with one gate.
- Customize Toolbar drew the Connection and Status tiles from an already-deallocated hosting
  controller, so both were blank. Both items are gone.
- Stop advertised `⌘.` as a literal, so it kept advertising it after Cancel Query was rebound.
- A duration is now tagged with the tab that produced it. Moving the readout under a tab's own rows
  would otherwise have reported a background tab's query as the selected tab's.
- The Safe Mode item wrote a bare tooltip after the level-aware one, and the item only rewrites the
  tooltip when the level changes, so the level never came back.
- **View > Show Tables** ran through the segmented control's toggle, which closes the sidebar when
  you pick the list already showing. The menu commands reveal instead.

## Autosave

`toolbarIdentifier` moves to `.v3`. A stored `.v2` list names identifiers the delegate no longer
vends, and replaying it would leave existing users the crowded toolbar this change exists to fix.
Saved arrangements reset once.

## Verified

- `verify.sh build`: PASS
- `verify.sh test` over the 13 affected suites: 98 executed, 98 passed
- `verify.sh lint TablePro TableProTests`: 0 violations
- `docs/scripts/check-writing-style.sh` and `check-docs-against-source.py`: both clean
- AppKit behaviour settled by compiled probes rather than assumed: native centred group survival
  at 700pt, `NSToolbar.items` never containing group subitems, an icon-only toolbar still drawing
  an item's `title`, and an autosaved display mode being restored over an init-time write.

No UI automation was added. The behaviour that regressed is toolbar overflow at specific window
widths, which depends on the real titlebar geometry; the accessibility-tree measurements above are
how it was checked, and they are not deterministic enough on a CI runner to pin as a test. The
structural rules are pinned instead: no item is view-backed, the default set stays inside a
titlebar, commands outrank the centre, and every relocated command has a menu-bar route.

https://claude.ai/code/session_014qSA9xxbAUskqMd55gTCkG


datlechin:
## Before / After

Screenshots could not be attached from the CLI (`gh` has no image upload), so the same comparison is written out. Both taken on a Debug build at a 1500pt window, macOS 26.2, PostgreSQL.

**Before**, left to right: sidebar segmented control, Back/Forward, a globe button, a cylinder button, then a centred hosted view carrying tag badges, the engine glyph, the connection name in a coloured capsule, a divider, a database chip, a chevron, a schema chip, a padlock, a query spinner with Stop, an elapsed-time readout and a schema-refresh spinner. Then Refresh, Save, Add Row, Restore, Open Quickly, New Tab, Preview, Assistant, Inspector. The window title read `TablePro` while the capsule read `TablePro`; the subtitle read `tablepro_license · public` while the chip read `tablepro_license › public`.

**After**, left to right, icon-only, five capsules:

`[list | star]`  `[‹ ›]`  `[engine-glyph TablePro | cylinder tablepro_license]`  `git_hub_releases`  `[↻ ✓ + ⟲]`  `[⊞ 🔍]`  `[lock ⌄]`  `[✨]`  `[▤]`

The centred pair is one capsule with a divider between its two halves, each half a control that opens its own chooser. The window title beside it names the tab, which is the one fact the centre does not carry.

datlechin:
## Review fixes (32ff55832)

A second review pass on the committed diff found seven real defects. All fixed, with regression tests.

**Correctness**

- `PaginationCoordinator.swift` — Fetch All tagged the duration with `selectedTabId` instead of the `tabId` it fetched for, so switching tabs mid-fetch attributed the time to the wrong tab. `tabId` was already in scope six lines above.
- `SchemaMenuDelegate` clears the menu on every open, so the statically added **Open Schema Switcher…** was destroyed the first time the submenu was used. The command now comes from the delegate. My earlier test passed because it inspected the freshly built menu; `schemaSubmenuSurvivesItsDelegate` now runs the delegate first.
- The status bar read `isExecuting`, which counts claims only. Fetch All registers *unclaimed* work, so it ran with no spinner and no Stop button anywhere. Now `isBusy`, which is what the readout it replaced effectively watched.
- `SafeModeToolbarItem` carries no action, so AppKit never sent `validateToolbarItem(_:)` for it and the new validation arm was dead code: the padlock stayed live over a session that had gone. It resolves its own enablement now.
- One timing slot meant any tab finishing a query erased the duration another tab was showing. Keyed by tab, and pruned when a tab closes.
- `dismissScopeSwitcher()` called the presenter's `dismiss()` immediately before `present()`, which defeated toggle-to-close and destroyed typed search text. The presenter now tracks which chooser is up, so the same command closes it and the other replaces it.
- `EditorTabStrip.executionOwner` held the coordinator strongly, the reference shape `MainWindowToolbar` documents as unsafe. Now weak.

**Cleanup**

- The schema-refresh indicator lost its only surface when the centred item went; re-homed to the status bar.
- `NSImage(named:)` returns the shared cached instance, so writing `isTemplate`/`accessibilityDescription` on the engine icon rewrote it for every other consumer. Copied first.
- `databaseVersion` and `databaseGroupingStrategy` became write-only; removed with their writers.
- Dropped a stale `currentSchema` observation that forced a full validation pass on every schema switch, and three now-unused `import SwiftUI`.

126 tests pass, build clean, lint clean.

datlechin:
## CI failures fixed (0647fef54)

Three red checks, investigated with compiled AppKit probes and CI artifacts. Two were mine; one was not.

### Unit tests — mine

`WindowBusyStateGuardTests.executionIndicatorReadsTheRegistry` scanned for the literal `isExecuting: coordinator?.tabExecution.isAnyExecuting`, which lived in the deleted hosted toolbar view. Re-spelling the scan would only have moved the problem, since it would then anchor on a local binding.

Fixed at the source instead: `ExecutionReadout.isExecuting` is no longer a stored `Bool` any caller can hand any value, it is computed from a stored `TabExecutionRegistry`. The invariant now lives in the type rather than in a text search, and the guard is behavioural — it claims, settles, and exercises unclaimed work (Fetch All) against the readout. Reintroducing the stored flag changes the memberwise initializer and the test file stops compiling, which is what the suite header promises.

### UI tests 1/3 — mine

`QueryInsightsTabUITests` failed with `Multiple matching elements found` for the **Query Insights** menu item.

Measured with a compiled harness: AppKit titles its auto-managed Window-menu entry `Title (Subtitle)` when a window has a subtitle, and exactly `Title` when it does not. This branch drops the subtitle, so a window showing the Query Insights tab now contributes a menu-bar item titled exactly `Query Insights`, colliding with the Database command of the same name. CI's own failure dump names both matches.

The collision is inherent — the command and the tab are the same feature — so the fix is to scope the query, not rename anything:

```swift
app.menuBars.menuBarItems["Database"].menuItems["Query Insights"]
```

Precedented at `OpenQuicklyCommandUITests.swift:35`. **I did not restore the subtitle**: that would undo the deliberate de-duplication this PR exists for.

New `MenuQueryScopeGuardTests` stops the next one being written unscoped. It fails on any UI test resolving `menuBars.menuItems[…]` by a title `WindowTitleResolver` can produce, and a second case fails if the resolver learns a title the guard does not cover. Server Dashboard, Users & Roles, ER Diagram and Create Table were all latent instances of the same class.

### UI tests 2/3 — not mine

`CopyObjectsUITests.testAFilterCarryingASecondStatementBlocksContinue` fails identically on **main**: runs 34104392446 and 34028478633, same assertion, same message. At the merge base (0ff1ccfea) this exact case reports First Run Failed / Retry 1 Passed — the green baseline was a retry-rescued green. The branch touches no file on the Copy To path.

I have left it alone. A proposed test-only fix (re-rooting the poll at the sheet) was refuted under verification: `exists` re-resolves the chain from the app element each poll, so re-rooting does not avoid the 10,026-cell walk, and the starvation theory did not survive measurement. Making the assertion pass without understanding why the catalog read stalls would hide a real app defect. Worth its own issue.

### Also fixed

`Database > Safe Mode` held a child also titled **Safe Mode**, so the menu bar had two items with one title, which System Settings binds App Shortcuts by. The container this PR introduced is now **Safe Mode Level**; the level keeps its shipped name. A stale comment claiming the schema is readable from the window subtitle is corrected.

102 unit tests pass, lint clean.

PR #2669, opened 2026-09-07
feat(linux): add result grid context menu with Copy as and Export Results dialog

Seha16:
## Summary

- Right-click menu on every result grid, including query results in the SQL editor, which had no menu before: Copy, Copy as (Rows, With Headers, JSON, CSV, CSV with Headers, Markdown, IN Clause, INSERT Statement), Copy column name, Show Row as JSON, Set Value (Empty, NULL), Export Results..., Insert row, Duplicate, Delete. Editable-only items are hidden on read-only grids.
- Export Results dialog (`adw::Dialog`) with a format picker (CSV, JSON) and CSV options that match the macOS app: convert NULL to empty, convert line breaks to spaces, field names in first row, sanitize formula-like values, delimiter, quote mode, line break, decimal separator. Reset to Defaults. Options persist in `preferences.json`.
- Pure renderers live in `tablepro_core::export` with unit tests. The old `render_csv` in the app went through the grid's display formatter, so it truncated text at 10k chars and wrote the `NULL` sentinel into files; the new path renders full values.
- `serde_json` gets `preserve_order` so JSON output keeps column order.

Deferred to follow-up PRs: Paste rows, Set Value > NOW() / CURRENT_TIMESTAMP (needs a raw SQL expression marker in the change tracker), SQL / Markdown / HTML / XML / XLSX file export.

## Test plan

- [x] `./scripts/ci-local.sh` (fmt, clippy `-D warnings`, build, 338 unit tests; 21 new for the renderers)
- [x] SQLite smoke DB with a comma, a line break, a `"` and a `=SUM(A1)` value. Right-clicking on the editor result grid shows the read-only menu; on a table tab, it shows the full menu.
- [x] Copy as Markdown / JSON / IN Clause / CSV with Headers and plain Copy read back from the Wayland clipboard with the expected text.
- [x] Export dialog: switches and dropdowns write `preferences.json` on change, Reset to Defaults restores defaults, Export writes the file and shows the toast. Output with semicolon + line breaks to spaces: `2;has,comma;'-2.25;"line1 line2";false` and `3;'=SUM(A1);10;"say ""hi""";false`.
- [x] Set Value > Empty marks the cell as modified and shows the unsaved-changes bar.
- [x] Show Row as JSON opens a read-only dialog with column order preserved.
- [x] Paginator export button opens the same dialog.
- [x] Dark theme

04-context-menu
05-copy-as-submenu
07-row-as-json
08-export-dialog

https://github.com/user-attachments/assets/1a616125-f26d-4ac3-bf60-399cf0e133e0

04-context-menu
05-copy-as-submenu
07-row-as-json
08-export-dialog

https://github.com/user-attachments/assets/3460f7f1-35c2-4d06-88ca-22ce4e1f6398

### DIFF
--- a/linux/Cargo.lock
+++ b/linux/Cargo.lock
@@ -4151,6 +4151,7 @@ version = "1.0.149"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "83fc039473c5595ace860d8c4fafa220ff474b3fc6bfdb4293327f1a37e94d86"
 dependencies = [
+ "indexmap 2.14.0",
  "itoa",
  "memchr",
  "serde",
--- a/linux/Cargo.toml
+++ b/linux/Cargo.toml
@@ -23,7 +23,7 @@ async-trait = "0.1"
 anyhow = "1"
 thiserror = "2"
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+serde_json = { version = "1", features = ["preserve_order"] }
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "net", "time", "sync"] }
 tokio-util = { version = "0.7", default-features = false }
 russh = "0.55"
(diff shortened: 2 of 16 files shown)


datlechin (approved):
Thank you

PR #2675, opened 2026-09-08
feat(toolbar): put the transport rate beside the connection in the centre

datlechin:
Moves the live throughput figure into the centred connection group, next to the connection name and the database. The connection switcher keeps the totals and the transport name; the toolbar gets the rate.

## Why it could not be done with a title

The obvious shape, appending the rate to the connection item's `title`, does not work, and the reason is not the one I expected. Measured on a real 1200pt `NSToolbar` with a centred group:

| Step | group.x | group.w |
|---|---|---|
| initial title `prod-db` | 552.0 | **96.0** |
| set title `prod-db ↓145 kB/s` | 552.0 | **96.0** |
| `validateVisibleItems()` | 552.0 | **96.0** |
| window resize | 552.0 | **96.0** |
| `displayMode` round trip | 519.5 | **161.5** |
| remove + re-insert the item | 519.5 | **161.5** |

**AppKit sizes a view-less item's title once, when the item is inserted.** A longer title written afterwards is clipped to the old width, and neither `validateVisibleItems()` (the channel `StatefulToolbarItem.applyTitle()` uses) nor a window resize re-measures it. The only things that do rebuild the toolbar visibly.

And when it *does* re-measure, the geometry bites: a centred group is laid out around its midpoint, so between `↓0 kB/s` (group.x 488.5, w 223.0) and `↓145 kB/s` (group.x 481.5, w 237.0) the leading edge moves 7.0pt one way and the database item 7.0pt the other. Every time the figure changes bucket.

Padding with figure spaces does not rescue it either. Measured at 13pt system font: `↓145 kB/s` is 61.93pt but `␣↓64 kB/s` is 64.33pt, so trading a digit for a figure space makes the string *wider*. U+2007 is not a digit width in SF, and the decimal separator is 5pt narrower than a digit.

## What this does instead

A fixed-width `NSTextField` in a toolbar item, added as a third subitem of the centred group. Measured, with the width pinned nothing moves at all: group and field frames were byte-identical across `↓0 kB/s`, `↓145 kB/s`, `↑1.2 MB/s` and `↓88 MB/s` (group.x 512.5, group.w 175.0, field.x 607.5, field.w 82.0). Adding or removing the subitem *does* re-measure (89.0 ↔ 175.0), which is what it is for: that happens when a connection is adopted, never under a running tunnel.

This is the one view-backed item in the toolbar, and `MainWindowToolbarNativeContractTests` now says so explicitly rather than passing by accident. It is not the shape #2668 removed: that was a hosted SwiftUI status item with no width of its own to give back, so AppKit could only drop it whole. This one is pinned to a width measured from the widest figure it can ever draw, so it never needs compressing.

Two details that follow from the measurements:

- `.byteCount` cannot produce the figure. `allowedUnits: [.kb]` floors nothing (512 still comes back `512 bytes`) and zero comes back `Zero kB`. Both are four characters wider than the figures around them. `TransportRateLabel` assembles a localized number and an unlocalized unit instead, floored at kB.
- The field uses `monospacedDigitSystemFont`, so the text does not shimmer inside its fixed frame either.

## Behaviour

- Only for a connection whose bytes the app carries, SSH tunnel or SOCKS proxy. Every other transport, and a direct connection, has no readout and no reserved space.
- Reads `0 kB/s` while nothing moves, which is most of the time. Polling TablePlus's equivalent gave `0 B/s` in 19 of 20 idle samples and 37 of 45 under load; the figure being quiet is normal, not broken.
- The ticker samples once a second and only for a measurable transport, and writes into the field rather than calling `validateVisibleItems()`, so a new figure is a redraw inside one view and never a layout pass.
- The readout publishes no action, so AppKit never validates it and it has no menu-bar command. It does get a disabled overflow entry naming the direction in words, because the centred group is the first thing AppKit sheds when the window narrows.

## Verification

- `build` PASS
- `test TransportRateLabelTests MainWindowToolbarNativeContractTests ConnectionTransportActivityTests TransportRateSamplerTests TransportActivityRegistryTests MainWindowToolbarValidationTests`: 74 cases, 74 passed. The new tests pin the width budget against the widest candidate, that monospaced digits do not vary, and that a connection with no measurable transport carries no readout.
- `uitest ConnectionActivityFooterUITests` PASS.
- `lint TablePro TableProTests`: no findings in any file this branch touches.
- `docs` PASS.

## Not done

No UI test for the readout itself: the sample database is a local file, so a UI test can only assert the readout is absent, which `unmeasuredConnectionsCarryNoReadout` already does at the unit level. Asserting a live figure needs an SSH server with a query moving data through it.

No screenshot, for the same reason.

Unrelated, pre-existing: `verify.sh lint` reports `legacy_swiftui_aspect_ratio` in `ImportFromAppSourcePicker.swift` and `SupportView.swift`, and a stale `NSAccessibilitySortDirectionAttribute` reference at `CLAUDE.md:214` from #2670. None are files this branch touches.

https://claude.ai/code/session_014qSA9xxbAUskqMd55gTCkG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inspector shows the selected row as Fields or JSON, with AI Chat moved out of its tab strip.
 - Structured inspector values expand in place instead of replacing the whole pane.
 - Object tree context menu regrouped into four groups by intent, with the connection-wide commands off object rows.
+- Back and Forward absent from the toolbar while a tab has no history to walk, in place of two dimmed arrows.
 - View Options as a control in the sidebar's filter row, in place of an entry on every context menu.
 - View ER Diagram and New View on the object tree's empty-area menu only.
 - Keyboard shortcuts shown on sidebar context menu items that have a menu bar equivalent.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inspector shows the selected row as Fields or JSON, with AI Chat moved out of its tab strip.
 - Structured inspector values expand in place instead of replacing the whole pane.
 - Object tree context menu regrouped into four groups by intent, with the connection-wide commands off object rows.
-- Back and Forward absent from the toolbar while a tab has no history to walk, in place of two dimmed arrows.
+- Back available with unsaved edits, asking to discard them, in place of standing down until they were saved.
 - View Options as a control in the sidebar's filter row, in place of an entry on every context menu.
 - View ER Diagram and New View on the object tree's empty-area menu only.
 - Keyboard shortcuts shown on sidebar context menu items that have a menu bar equivalent.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Show Tables and Show Favorites in the View menu.
 - Running indicator on the editor tab whose query is executing.
 - Sort direction setting for the data grid, applied to the default row sort and to the first click on a column header. (#2665)
-- Transport activity under the connection switcher: bytes carried, and live throughput on an SSH tunnel or SOCKS proxy.
+- Transport activity for an SSH tunnel or SOCKS proxy: live throughput in the toolbar, bytes carried in the connection switcher.
 
 ### Changed
 
--- a/TablePro/Core/Services/Infrastructure/MainWindowToolbar+Delegate.swift
+++ b/TablePro/Core/Services/Infrastructure/MainWindowToolbar+Delegate.swift
@@ -47,11 +47,16 @@ extension MainWindowToolbar {
             /// Not navigational, unlike back and forward. `isNavigational` asks AppKit to lift an
             /// item to the leading edge of the content area, which is the opposite of what
             /// `centeredItemIdentifiers` asks for, and this group is the centred one.
-            return makeNativeGroup(
+            ///
+            /// The throughput readout joins it here rather than standing on its own, so it is
+            /// centred with the pair it describes and shares their capsule.
+            let group = makeNativeGroup(
                 id: itemIdentifier,
                 label: String(localized: "Connection"),
-                subitems: [subitemConnection(), subitemDatabase()]
+                subitems: connectionGroupSubitems()
             )
+            adopt(connectionGroup: group)
+            return group
         case Self.safeMode:
             return subitemSafeMode()
         case Self.editorGroup:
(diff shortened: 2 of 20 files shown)


datlechin:
## Back and forward: present and dimmed

Reverts the hiding introduced earlier on this branch, and lets Back ask before it discards.

### Root cause

Availability was modelled as **presence**. `subitems` is a composition property, not a visibility one, and the state driving it is `@ObservationIgnored` (`MainContentCoordinator.swift:323`). Reading `canNavigateBack` inside `withObservationTracking` registered nothing for the history, and none of the four properties `observeItemState` tracked changes when the user drills a foreign key. So once hidden, the pair **stayed** hidden until a tab switch.

Dimming needs no observation at all: `menuOnlyItem` sets `autovalidates = true`, and AppKit sends `validateToolbarItem:` to each subitem with `isEnabled` following each direction independently (measured).

### What Apple does

HIG, quoted: *"If a menu bar item isn't actionable, disable the action instead of hiding it from the menu."* Toolbars: leading-edge navigation items *"aren't customizable"*, to ensure they are always available.

Measured through the accessibility API on three running apps:

| App | With nowhere to go |
|---|---|
| Xcode | group present w=75, both segments `DISABLED` |
| Finder, column view | both `DISABLED` at w=75; after one navigation back enabled, forward `DISABLED`, same width |
| System Settings | Go Back enabled, Go Forward `DISABLED`, w=75 |

None hides it.

### Changes

- Deleted `syncNavigationVisibility()`, `navigationSubitems` and the stored `navigationGroup`; the delegate builds the group fresh with both subitems, as every sibling group does.
- Removed the two dead observation reads. `MainWindowToolbar+Validation.swift:72-75` is untouched: it was always correct.
- `navigableTabId` no longer refuses on unsaved cell edits. `step(back:)` routes through `confirmDiscardChangesIfNeeded(action: .navigation)`, the alert refresh, sort, pagination and filter already use. Staged **structure** edits still refuse outright, because that alert clears `changeManager` and nothing else.
- Fixed `MainWindowToolbarCustomizationTests.identifierIsStable`, which pinned the autosave name to `v3` while this branch bumped it to `v4`. **The branch was red on it.** It lives in a different `@Suite` type inside `MainWindowToolbarLayoutTests.swift`, so a `-only-testing` run by file name never reached it.

### Collateral, each verified with an in-process probe

Three findings, all fixed by the revert: the capsule vanishing the moment a cell edit was staged; hidden subitems escaping `refreshShortcutHints()` so a keyboard rebind never reached them; and both contract tests for the hiding being vacuous. The replacements assert on the **vended** item, because the four existing predicate tests stayed green through the whole defect.

### From the Codex review

Two P2 findings, both real and both mine:

- **A discarded primary-key edit could misdirect the row anchor.** `clearChanges()` drops the change records but leaves the edited value in `TableRows`, so an anchor captured after the discard keyed on a value the user had just thrown away. Forward then found nothing, or a different row carrying that key. Fixed twice over: the departing entry is captured *before* the guard, and the anchor rule moved into a pure `NavigationRowAnchor` that refuses any key column with a staged edit.
- **The alert said "Going back" when going forward.** Forward reaches the same guard. Reworded for both directions.

### Verification

`build` PASS. 119 cases across 14 suites, 119 passed, enumerated by suite **type** name. `docs` PASS. `lint` clean in every file this branch touches; the two `legacy_swiftui_aspect_ratio` findings are in `ImportFromAppSourcePicker.swift` and `SupportView.swift`, which it does not touch.

No UI automation for the dimmed state: `NavigationHistoryUITests` drives Back through the View menu by design, and asserting the toolbar button's enablement needs a connection with browse history that the sandboxed UI-test launch does not set up. Noted rather than skipped silently.

PR #2697, opened 2026-09-09
feat(plugins): change a SQLite column's type, nullability and default through the structure editor

datlechin:
Completes SQLite structure editing. #2694 and #2695 shipped foreign key editing through a reviewed table rebuild and left three things refused by name: a column's type, nullability or default; a rename or drop alongside a foreign key change; and a primary key change. The first two land here. The third does not, and the last section says why.

## Column type, nullability and default

SQLite has no `ALTER TABLE` for any of them, so they travel in the rebuild's new definition. The rebuild's design principle is that every column the save does not touch keeps its stored text byte for byte, which is what carries a `CHECK`, a `COLLATE`, a `GENERATED ALWAYS AS` and a `DEFAULT` containing a comma across; a column the save *does* touch has to be rewritten the same way, replacing only the span that changed.

`SQLiteColumnDeclaration` reads the declaration as a grammar rather than searching it for keywords, and that is the whole correctness argument. Measured on 3.54, searching is wrong in both directions and silently so:

| Declaration | What a keyword search does |
|---|---|
| `b TEXT CHECK (b IS NOT NULL)` | matches `NOT NULL` inside the check, so "make it NOT NULL" reads as already done and changes nothing |
| `a INTEGER REFERENCES p(id) ON DELETE SET NULL` | matches `NULL` inside the action, so "make it nullable" rewrites the action into a syntax error |

Both are covered by tests. The grammar itself also refutes the published railroad diagrams three times, each measured: a type name is a repeatable run of words (`UNSIGNED BIG INT`), a bare `NULL` is a constraint rather than a type, and `GENERATED` ends a type only as the exact run `GENERATED ALWAYS AS`.

Three edits are refused rather than guessed at:

- **Retyping an `INTEGER PRIMARY KEY`.** Measured: after the rebuild the key stops generating values and then stores NULL on every insert that omits it, twice over, with no error.
- **A column carrying the same constraint twice.** `a INT NOT NULL NOT NULL` is legal; rewriting one of a pair leaves the other in force and the edit reports success having changed nothing.
- **Nullability or default on a generated column.** Those belong to the expression, and SQLite rejects a `DEFAULT` on one outright.

A type change also re-reads every value through the new affinity: TEXT `'007'` copied into an `INTEGER` column is stored as `7`, with no undo. The review sheet says so before the script runs.

## Rename and drop alongside a foreign key change

Previously refused, for a good reason: a rebuild replays its dependent SQL verbatim and rewrites only the column's own declaration, so a trigger, a view or an incoming foreign key would be left naming a column that is gone.

The answer is not to rewrite dependent SQL by hand. `ALTER TABLE RENAME COLUMN` already does it, correctly, including views, which measured on 3.54:

```
CREATE VIEW v_named AS SELECT b FROM t     ->  CREATE VIEW v_named AS SELECT c FROM t
CREATE TRIGGER tr AFTER UPDATE OF b ON t   ->  CREATE TRIGGER tr AFTER UPDATE OF c ON t
CREATE INDEX ix ON t(b)                    ->  CREATE INDEX ix ON t(c)
```

So the rename and the drop run as their own `ALTER TABLE`, appended after the rebuild in the same transaction, and SQLite carries the change into every dependent object itself. The rebuild builds the table under the columns' current names and anything the save expressed in final names is mapped back.

## A latent bug in the shipped rebuild

`PRAGMA legacy_alter_table` decides whether the rebuild's own `RENAME TO` survives a dependent view, and the shipped planner never sets it. Measured:

| Setting | `ALTER TABLE t_rebuild RENAME TO t` with a view over `t` | A later `DROP COLUMN` that breaks a trigger |
|---|---|---|
| `1` | succeeds | succeeds silently, leaving the trigger broken |
| `0` | fails: `error in view v: no such table: main.t` | refuses with the reason |

It works today only because Apple's libsqlite3 defaults it to `1`; upstream defaults it to `0`, so a driver bundling its own SQLite would fail every rebuild on a schema with a view. `SQLiteColumnReorderPlanner`'s doc comment claimed no pragma was needed and was wrong as a general statement.

The plan now sets it `on` for the table rename and `off` for the column `ALTER`s, and restores the connection's original in the epilogue, because the setting survives the commit.

## What `ALTER TABLE` does not catch

`DROP COLUMN` at `legacy_alter_table=0` refuses a column an index needs, and a primary key column. It does not notice two cases, both measured, both of which committed silently:

- a view declared with an explicit column list, `CREATE VIEW v(a,b,c) AS SELECT * FROM t`, which breaks on the count
- a trigger on **another** table that writes a positional row into this one

Neither is visible to `PRAGMA foreign_key_check`, which is scoped to one table's keys. The plan now ends by `EXPLAIN`-ing one statement per view and per trigger subject table, which compiles each object's body against the table as it now stands. Measured: it reports `expected 3 columns for 'v' but got 2` and `table t has 2 columns but 3 values were supplied`, and executes nothing.

## Verification

| Step | Result |
|---|---|
| `build` | PASS |
| `test` (12 suites) | PASS, 169 cases |
| `build SQLiteDriver`, `CloudflareD1DriverPlugin`, `LibSQLDriverPlugin` | PASS |
| `docs` | PASS |
| `lint` | PASS for every path this branch touches |
| `abi` vs merge base | additive: no public symbol removed |
| `scripts/check-sqlite-rebuild-assumptions.sh` | 21 of 21 hold |

PluginKit 24 to 25: a new `PluginColumnAlteration` type, `alteredColumns` on `PluginTableRespecification` through a new init overload with the old one `@_disfavoredOverload`, and two new `Context` fields. `minimumCompatiblePluginKitVersion` stays at 19, so every shipped plugin keeps loading and no bulk re-release is needed.

`verify.sh plugins` still fails on this machine before reaching any of my code, on the vendored oracle-nio `@TaskLocal` macro. The three SQLite-family targets were built individually instead.

The assumption checker grew from 12 facts to 21, adding the `legacy_alter_table` pair, the `EXPLAIN` revalidation, the column-declaration grammar and the `INTEGER PRIMARY KEY` retype. A future macOS upgrade re-checks all of them rather than trusting this description.

## Screenshots

Not captured. Driving a Debug build needs Screen Recording granted to this session, which it does not have. The change is to the existing review sheet's script and caveats, which `docs/features/table-structure.mdx` already pictures.

## Primary key editing, and why it is not here

`SchemaChange.modifyPrimaryKey` has exactly one producer in the whole app: `applyPrimaryKeyChangeUndo`, an undo handler. Nothing stages it going forward, on any engine, so primary key editing is unreachable in the Structure editor everywhere and not just on SQLite. Building the SQLite half would be wiring behind a control that does not exist.

It also carries risks the foreign key work did not, both measured:

- Moving the key onto an `INTEGER` column repoints every row's identity. The `INSERT`'s explicit `rowid` is discarded when the new table declares an alias, so a row holding NULL in the new key silently becomes a fabricated one.
- It is the first rebuild whose damage lands outside the table being rebuilt, and every verification the plan has is scoped to this table.

Worth doing, as its own change, starting with how a primary key change gets staged at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Column type, nullability and default changes for SQLite, libSQL and Cloudflare D1.
+- Column rename and drop alongside a foreign key change in one save.
 - Foreign key add, remove and edit for SQLite, libSQL and Cloudflare D1, applied as a reviewed table rebuild.
 - Real constraint names for SQLite foreign keys, in place of a positional placeholder.
 - Menu of the engine's own default values on the Structure tab's Default cell, with No default, NULL, Empty string and a Custom editor. (#2688)
@@ -74,6 +76,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Rebuild failing on a table with a dependent view where SQLite defaults `legacy_alter_table` off.
+- A dropped column silently leaving a broken trigger or view behind.
 - "Unsupported schema operation" when adding a foreign key to a SQLite, libSQL or Cloudflare D1 table.
 - Add and Remove offered on the Foreign Keys tab for engines that cannot edit foreign keys.
 - Incomplete foreign keys, indexes and columns reaching the database on Save.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Column type, nullability and default changes for SQLite, libSQL and Cloudflare D1.
+- Column type, nullability and default changes for SQLite, libSQL and Cloudflare D1, checked against every dependent view and trigger before the change commits.
 - Column rename and drop alongside a foreign key change in one save.
 - Foreign key add, remove and edit for SQLite, libSQL and Cloudflare D1, applied as a reviewed table rebuild.
 - Real constraint names for SQLite foreign keys, in place of a positional placeholder.
@@ -76,8 +76,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Rebuild failing on a table with a dependent view where SQLite defaults `legacy_alter_table` off.
-- A dropped column silently leaving a broken trigger or view behind.
 - "Unsupported schema operation" when adding a foreign key to a SQLite, libSQL or Cloudflare D1 table.
 - Add and Remove offered on the Foreign Keys tab for engines that cannot edit foreign keys.
 - Incomplete foreign keys, indexes and columns reaching the database on Save.
--- a/Plugins/BeancountDriverPlugin/Info.plist
+++ b/Plugins/BeancountDriverPlugin/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>TableProPluginKitVersion</key>
-	<integer>24</integer>
+	<integer>25</integer>
 	<key>TableProProvidesDatabaseTypeIds</key>
 	<array>
 		<string>Beancount</string>
--- a/Plugins/BigQueryDriverPlugin/Info.plist
+++ b/Plugins/BigQueryDriverPlugin/Info.plist
@@ -5,6 +5,6 @@
 	<key>TableProMinAppVersion</key>
 	<string>0.42.0</string>
 	<key>TableProPluginKitVersion</key>
-	<integer>24</integer>
+	<integer>25</integer>
 </dict>
 </plist>
--- a/Plugins/CSVExportPlugin/Info.plist
+++ b/Plugins/CSVExportPlugin/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>TableProPluginKitVersion</key>
-	<integer>24</integer>
+	<integer>25</integer>
 	<key>TableProProvidesExportFormatIds</key>
 	<array>
 		<string>csv</string>
--- a/Plugins/CSVImportPlugin/Info.plist
+++ b/Plugins/CSVImportPlugin/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>TableProPluginKitVersion</key>
-	<integer>24</integer>
+	<integer>25</integer>
 	<key>TableProProvidesImportFormatIds</key>
 	<array>
 		<string>csv</string>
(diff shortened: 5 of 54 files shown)


datlechin:
Codex reviewed the branch and found sixteen issues, seven of them P1. Its verdict was "not safe to merge as written", and it was right. All of the P1s and the cheap P2s are fixed in 7203f810; the rest are refused rather than mishandled.

**Correctness bugs it caught, each now fixed and tested:**

- **Every drop-without-rename save failed.** `namedAsBuilt` deferred drops only when a rename existed, so with no rename the temporary `CREATE TABLE` already omitted the column and the trailing `ALTER TABLE ... DROP COLUMN` raised `no such column`, rolling back the whole save.
- **A name swap sent an alteration to the wrong column.** `alteredColumns` is recorded against the current name, but `namedAsBuilt` was mapping it as a final name. Renaming `A`→`C` and `B`→`A` applied `A`'s new type, and its value coercion, to `B`. Alterations are no longer remapped.
- **`0.5` refused every edit on its column.** The tokenizer splits on `.`, so a `DEFAULT 0.5` arrived as three tokens and the remainder read as an unknown constraint. Numeric, exponent and `X'…'` blob literals are now consumed whole.
- **A table-level `PRIMARY KEY(id)` evaded the rowid-alias guard**, so `CREATE TABLE t(id INTEGER, PRIMARY KEY(id))` could be retyped to TEXT and start storing NULLs. The alias is now determined from the whole table definition.
- **Setting Primary Key = YES silently rebuilt the table `NOT NULL` and left the key unset.** A `.modifyColumn` folds every field the user changed, and the rebuild carries three of them; the rest were dropped and reported as applied. Changes to primary key, auto-increment or generation are now refused by name.
- **The trigger probes were wrong in both directions.** A view carrying only an `INSTEAD OF DELETE` trigger failed the unconditional `INSERT` probe, blocking every drop on that schema, and an `AFTER UPDATE` trigger was never compiled at all. Each trigger is now probed through the event it fires on.
- **`Context` gained defaulted parameters instead of an overload**, replacing a published mangled symbol. The exact five-argument initializer is restored, delegating.
- **A retype could break an existing foreign key** while enforcement is off: TEXT `'007'` copied into an INTEGER column becomes `7` and no longer matches a parent's `'007'`. The key check now runs after a retype as well as after a key change.
- **Replacing a default deleted its `CONSTRAINT` name.** Replacing now writes inside the clause; only removal takes the name.
- **Drops now run before renames**, so renaming `a` to `b` while dropping the existing `b` no longer trips SQLite's duplicate-name check.

**Refused rather than half-handled**, each valid as two saves: a save that frees a column name and takes it in the same breath, in either direction (rename-into-a-dropped-name, or drop-and-readd). The message says which name and what to do.

Two of its P2s I did not act on, and why: routing a staged index or check-constraint *deletion* into the rebuild before the copy would help a narrow case (retyping to a type that collapses two unique values while dropping that unique index), and mapping a renamed column through an added generated column's expression. Both currently fail the transaction and roll back with SQLite's own message rather than committing anything wrong, so they are a message-quality gap, not a correctness one. Worth doing; not worth widening this change for.

Re-verified after the fixes: build PASS, 150 test cases PASS, three SQLite-family plugins PASS, docs PASS, lint clean, and all 21 assumptions in `scripts/check-sqlite-rebuild-assumptions.sh` still hold.

PR #2719, opened 2026-09-11
fix(sidebar): keep the object list for a database with no tables, and list routines in schema-grouped sidebars

datlechin:
From user feedback: "When viewing a database, if there are no tables it hides the whole UI, but a database can have no tables and still have procedures, functions and so on."

## Root cause

Three things, one class of bug: the sidebar decided "empty" from half the data.

- **Flat sidebar (MySQL, MariaDB, SQL Server, PostgreSQL by default).** v0.73.0 swapped the whole outline for a full-pane "No Tables / This database has no tables yet." view whenever the table list came back empty. #2705 (unreleased) added a routines check to that decision, but `SchemaService.runLoad` commits the tables before routines, triggers and types arrive, so a zero-table database still flashed the empty pane on every load, and kept it for good when the routine fetch failed, because `fetchObjectsSafely` turned the failure into "nothing". The full-pane view also removed the Tables section, whose context menu is where **New Table…** lives.
- **Schema-grouped sidebar (Oracle, Snowflake, BigQuery, Dameng, Trino).** Each schema row listed its tables and nothing else. Routines were fetched once for the session schema into connection-wide storage the schema rows never read, so a schema holding only procedures read as empty.
- **No state to tell the cases apart.** Routines, triggers and types were bare arrays. "Still loading", "the fetch failed" and "there are none" were the same empty array.

## Fix

- `SchemaService` keeps routines, triggers and types as `MetadataLoadState` (the type the tree already uses), connection-wide and per schema. Every commit follows the rule in CLAUDE.md: loading only with nothing to show, a failed refresh keeps good rows, a failed first load reports itself, a cancelled fetch never leaves a spinner. A failed tables fetch now still settles the other kinds. `MetadataFetchOutcome` plus `enteringLoad`/`settled(by:discardingValue:)` hold those rules in one pure place.
- **Flat shape:** `SidebarObjectListPresentation` loses `.empty`. A loaded database always shows its outline, and a section with no rows says why with the status row the tree already draws: **Loading…**, the error, or **No items**. A search that matches nothing keeps the existing no-match view.
- **Schema-grouped shape:** a per-schema load now fetches that schema's routines, triggers (where the engine browses them) and types alongside its tables, with a generation guard so a superseded load cannot commit. The sidebar's per-schema loads (expand, search, a schema's **Refresh**) now go through `DatabaseManager.withMetadataDriver` with the schema's resolved scope, as CLAUDE.md requires of every metadata read. They used the raw session driver before, which a tab bound to another Snowflake database could have answered for. Schema rows resolve their children through `DatabaseTreeObjectGroupResolver` and the same container status rule the database tree uses, so a schema with only procedures shows a Procedures group. Search keeps a schema whose procedure, function or trigger matches, and keeps a schema whose side fetch failed rather than dropping it unjudged. Refresh reloads the side kinds in place.
- **Group placeholders, both tree shapes:** an empty kind group used to say **No items** while its fetch was still running, and a failed fetch showed the same false **No items** with the error moved up to the container. A group's placeholder now follows its own kind's phase (`DatabaseTreeSidePhases`), and the container shows a failure only for a kind it lists no group for.
- The schema-grouped empty state said "No Datasets / This project has no datasets yet." on every engine. It now uses the engine's own schema noun. BigQuery's schema noun is corrected to "Dataset" in its curated metadata and plugin, which also fixes its rename menu title.
- `loadSchemaTables`, `reloadSchemaTables` and `refreshLoadedSchemaTables` are renamed to `…SchemaObjects`, since they now load more than tables. Every caller and test moved in this commit.

## Tests

- `SchemaServiceSideObjectsTests` (new): a zero-table database settles its routines as loaded; a failed first routine fetch is reported, not read as none; a failed tables fetch still settles routines; reload after failure recovers; a failed reload keeps rows; a schema load fetches that schema's routines and triggers; a schema's routine failure is reported and leaves its tables; reload keeps routines on failure; invalidate clears per-schema side objects.
- `MetadataLoadStateTests`: `enteringLoad` and every `settled` arm.
- `SchemaServiceSideObjectsTests` also pins that a schema whose routine fetch failed is not settled for a search.
- `DatabaseTreeNodeTests`: empty-section and empty-container status, a group placeholder follows its own kind's phase, a failure goes on the container only when no group of its kind is listed, a procedures-only schema gets a Procedures group.
- `DatabaseTreeFilterTests`: a matching procedure or trigger keeps its schema; name-matched and unmatched schema buckets.
- `SidebarObjectListPresentationTests`: updated to the new spec (a loaded, empty database is a list).

No UI automation: the flow needs a server database with no tables and at least one procedure. The only deterministic database the UI test harness can open is the bundled SQLite sample, which has tables and no procedures, and SQLite has no procedures at all.

## Review

- **Codex `review`** read the diff and raised three findings, all fixed here: per-schema loads bypassed the metadata route (P1); a failed side fetch counted as settled, so search dropped the schema (P2); empty kind groups said **No items** while loading or after a failure (P2).
- **Codex `adversarial-review`** did not run: the workspace hit its usage limit. `code-review` at high read the diff instead and raised two findings, both fixed: a cancelled load left the side kinds it had marked loading with nothing to settle them, and routing each schema through its own scope opened a pooled connection per schema during a search. Per-schema loads now take the database's scope, since every fetch names its schema.

## Verification

- `verify.sh test` on 15 suites owning the changed types: 242 passed, 0 failed.
- `verify.sh lint` on every touched file: 0 violations. (The step's `agent docs:` line flags a stale `CLAUDE.md` symbol that is already on main.)
- `verify.sh docs`: pass.
- `BigQueryDriverPlugin` scheme builds.

## Before / After

Not captured. Driving a sandboxed Debug build against a local MariaDB takes synthetic input on a machine another agent was building on in parallel. The before state is v0.73.0's full-pane "No Tables" view; the after state is the outline with **Tables** reading **No items** and **Procedures** listing the procedure.

https://claude.ai/code/session_01V8tJZrNmbUg8JY4X9d7yNn


datlechin:
## Before / After

The same empty SQLite file (no tables, no views) opened in a sandboxed app.

Before (0.73.0): the whole object list is replaced by "No Tables", including the Tables section that holds **New Table…**. This is the view the user reported.

Sidebar in 0.73.0 replaced by a full-pane No Tables message on an empty database

After: the outline stays. The Tables section says "No items" and keeps its context menu, and the Triggers section is still listed.

Sidebar outline on an empty database with the Tables section showing No items and a Triggers section below

SQLite has no procedures, so this shows the empty-list half of the fix. The zero-table-with-procedures case needs a MySQL, SQL Server or Oracle server and is covered by the unit tests described in the PR body.

https://claude.ai/code/session_01V8tJZrNmbUg8JY4X9d7yNn

PR #2720, opened 2026-09-11
fix(connections): name why a driver is unavailable and offer the fix where the error appears

datlechin:
## Problem

A user imported connections from TablePlus. Every connect to the SQL Server one failed with:

> MicrosoftSQLServer driver plugin not loaded. The plugin may be disabled or missing from the PlugIns directory.

They installed the MSSQL plugin because the message said to. Nothing changed. The real cause was the connection's database type (`MicrosoftSQLServer`, which no plugin provides). Editing the connection and choosing the type fixed it. They also asked that the error take them straight to the Plugins window.

## Root cause

`DatabaseDriverFactory.createDriverFromPlugin` could only tell "not installed" apart from everything else. Three different causes, each with a different fix, all ended up in one unlocalized message that blamed a disabled or missing plugin:

- an unrecognized database type
- a plugin the user switched off
- a plugin that failed to load

The failure vocabulary had one plugin case (`pluginMissing`, shown as **Install Plugin…**). No window, alert or sheet could offer any other fix.

Three related defects sat on the same path:

- `prepareForConnecting` logged a failed on-demand install and threw the error away, so the user saw "not installed", clicked Install, and hit the same failure before the real reason appeared.
- Connecting called `installMissingPlugin`, which silently switched a registry plugin back on that the user had turned off. A bundled plugin that was turned off got the generic message instead.
- Test Connection's install prompt then ran `.openConnection` with the throwaway test id. The test never ran again and nothing was reported.

## Fix

- `PluginManager.driverUnavailability(for:)` names the cause: unknown type, disabled, failed to load (with the recorded reason), outdated, or not installed. A bundle that fails during the eager load is now recorded in `rejectedPlugins` too, so its reason is known.
- The factory throws a typed `PluginError` for each cause, each with a localized description and recovery suggestion. The raw string is gone, and "not installed" names the type's display name.
- `ConnectionRecoveryAction` has four cases: **Install Plugin…**, **Enable Plugin**, **Open Plugin Settings**, **Edit Connection…**. `pluginMissing` became `actionRequired(info, action)` in both `ConnectionAttemptOutcome` and `ConnectionUnavailableReason`, and every switch and fixture was updated in this commit.
- Enable Plugin switches the plugin on and reconnects. If the load gate withdraws the plugin, the reconnect comes back as "failed to load".
- Open Plugin Settings opens **Settings > Plugins > Installed** with that plugin selected (`PluginsSettingsNavigation`). The pane also offers **Try Again**, because nothing reports back from Settings.
- Edit Connection… opens the connection form. After the edit, the window resets to **Connect**.
- `ConnectionRecoveryPerformer` runs the action for every surface: the inline pane, the welcome window's failure alert, the Test Connection sheet, and the launch router's alert. Failures stay inline wherever a window exists.
- `prepareForConnecting` never switches a disabled plugin back on, and it throws the install error with its own reason.
- Test Connection runs the test again after installing the plugin it asked for.
- A type is only called unrecognized when no plugin, curated entry or registry entry provides it, so a type a newer registry plugin provides still gets **Install Plugin…**.
- A linked or team-library connection is not in the local store and cannot be edited here, so it is never offered **Edit Connection…**.
- The edit state resets only when the database type actually changes, so a rename, a colour change, a group move or a sync batch leaves **Edit Connection…** in place.
- A failure from `openTable` or `openQuery` launch intents, which open the window before connecting, is handed to that window's pane instead of being dropped.
- An eager-load rejection takes its database types from the registry sidecar when the bundle declares none. A failed install names its reason in the message itself, so fallback alerts that show only the description still carry it.
- `prepareForConnecting` classifies first and installs only a driver that is not installed. A plugin that failed to load is no longer uninstalled, with its settings, before the registry has been asked for a replacement.
- A disabled lazy plugin's code is no longer loaded to answer a driver lookup: `activateLazyBundle` returns before the load gate for a disabled entry.
- A changed database type retires every fix offered for the old type, not only Edit Connection.
- A failure handed to a window from a launch intent passes an attempt fence (`acceptsExternalFailure`), so it cannot undo a Cancel or overwrite a newer attempt.
- Every rejection path (discovery, withdrawal, eager load) takes its database types through one helper that falls back to the registry entry.
- A type unknown locally fetches the registry manifest before it is classified, so the result no longer depends on whether the launch-time fetch has finished.
- `PluginManager.shared` now defaults to `AppStorageEnvironment.shared.defaults`, which is `.standard` in a shipped build. That is what lets a UI test switch a plugin off without writing to the developer's real defaults domain.

One semantic change: `.actionRequired` still blocks the automatic connect on window activation, as `pluginMissing` did, but it now allows a manual **Reconnect**. A fix made in Settings or the connection form happens away from the window, and the old rule left no way back to connect afterwards.

Out of scope: already stored connections with an unrecognized type are not migrated. They now fail with **Edit Connection…**, which is the repair. The importer mapping that produced `MicrosoftSQLServer` is fixed in a separate PR.

## Tests

- `DriverUnavailabilityTests` (new): the availability matrix; a disabled plugin stays disabled through `prepareForConnecting` and the registry is never contacted; a failed on-demand install surfaces its reason, using an injected unreachable registry.
- `ConnectionFailureClassifierTests`: each error maps to its action, an unrecognized type never offers Install, and a failed install stays a plain failure.
- `ConnectionWindowPhaseMachineTests`: actions round-trip, editing the record resets only the Edit Connection state, and the Reconnect rule.
- `ConnectionRecoveryPresentationTests` (new): the pane's button titles, the retry rule, and how navigation requests are consumed.
- `pluginMissing` fixtures were updated in `ConnectionWindowPaneResolverTests`, `MainMenuBuilderTests` and `WindowTitleResolverWindowTests`.
- `PluginDisabledRecoveryUITests` (new): switch off the bundled SQLite plugin, disconnect, reconnect, check that **Enable Plugin** appears and reconnects.

Verification:

- `verify.sh build`: PASS.
- `verify.sh test` on the 20 owning suites: PASS, 184/184.
- `verify.sh lint` on every touched file: 0 violations.
- `verify.sh docs`: PASS.
- `verify.sh uitest PluginDisabledRecoveryUITests`: INCONCLUSIVE locally, 0 cases. The runner stopped at "Timed out while enabling automation mode", which is an unanswered macOS automation prompt on this machine. CI runs it.

The unknown-type flow has no UI test. A connection with an unrecognized type cannot be made through the UI, and seeding one means hand-writing `connections.json` in the `StoredConnection` schema, which has dozens of required keys. The unit tests cover that path.

## Review

- **Codex `review`**: 8 findings, all fixed. Six are listed in Fix. The other two: a reveal request for a withdrawn plugin now shows the rejected-plugins banner instead of doing nothing, and the new explanatory comments were removed per CLAUDE.md.
- **Codex `adversarial-review`**: 7 findings. Five are fixed (listed in Fix). Two predate this change and are left for their own PRs:
  - Reopening a recently closed tab consumes its entry, and any overflow SQL, before the connect succeeds, so a failed connect loses the tab.
  - A session state built for one database type is reused after the connection's type is edited to another.
- Codex hit its usage limit after those two passes, so the last round of fixes was read by the `code-review` skill instead.

## Screenshots

Not captured. The same automation prompt that stopped the UI test also blocks driving the app from a script on this machine. The states to capture are the inline pane for **Enable Plugin**, **Open Plugin Settings** with **Try Again**, and **Edit Connection…**.

https://claude.ai/code/session_01V8tJZrNmbUg8JY4X9d7yNn


datlechin:
## Before / After

Opening an empty `.sqlite` file with the SQLite plugin switched off. The connect runs in `TabRouter.openDatabaseFile`, not in the window, so the window only learns the result when the session goes away.

**Before:** "Disconnected from pr3-empty" for a connection that never connected, and **Reconnect** as the only fix. The same failure also came up a second time as a "Connection Failed" alert on top of the pane.

**After:** "Could not connect to pr3-empty" with **Enable Plugin**, and no alert over it.

What changed: `DatabaseManager` now records a `ConnectionEndReason` in place of a bare `ConnectionFailureInfo`. It is either `connectFailed(info, action)`, classified where the failure is recorded, or `sessionLost(info)`. `onSessionChanged` turns an attempt the window did not own into `.actionRequired` or `.failed`, and only a lost session into `.disconnected`. A connect that fails after `TabRouter` has opened a window now carries that window's connection id, so `LaunchIntentRouter` leaves the failure to the pane. The `adoptRecoverableConnectFailure` path is gone.

Captured with no automation permission: a sandboxed Debug build (`TABLEPRO_UI_TEST_SANDBOX`), `com.TablePro.disabledPlugins` set in the `com.TablePro.uitest` suite and removed afterwards, and `screencapture -l` on the window.

Before: Disconnected from pr3-empty with Reconnect as the primary button

After: Could not connect to pr3-empty with Enable Plugin as the primary button

Issue #1986, opened 2026-07-29
Database request: Neo4j


lipengyu:
### Problem

Currently, our project relies on Neo4j for data storage. While this works for basic CRUD operations, we face significant performance bottlenecks when handling complex, deeply connected data (e.g., multi-hop queries, knowledge graphs, or real-time recommendation engines). Traditional SQL JOINs are inefficient, and existing NoSQL solutions lack native graph traversal capabilities.

### Proposed solution

I propose implementing a unified graph database abstraction layer​ that supports Neo4j as the primary backend, while ensuring compatibility with high-performance alternatives such as FalkorDB​ and NornicDB.

Key objectives:

1. Unified Adapter Layer:​ Develop a common interface to abstract the underlying Cypher/GQL dialects. This allows developers to switch between Neo4j, FalkorDB, and NornicDB with minimal configuration changes.

2. Performance Optimization:​ Leverage FalkorDB's low-latency, memory-efficient architecture for real-time caching scenarios, while utilizing Neo4j's robustness for large-scale persistence.

3. Documentation & Tooling:​ Provide clear setup guides for each supported database, including Docker Compose templates for local development.

### Alternatives considered

1. Vendor-Specific Implementation:​ Building exclusively for Neo4j creates a hard dependency and prevents us from leveraging specialized features (like FalkorDB’s Redis-based speed).

2. Manual Query Management:​ Writing separate query sets for each database type is unsustainable and increases maintenance burden.

3.Sticking with Current Stack:​ Forcing relational databases to handle graph logic results in exponential query latency as data complexity grows.

Supporting Neo4j, FalkorDB, and NornicDB​ concurrently through an abstraction layer ensures the system remains performant, flexible, and future-proof.

### Related database type

None

PR #2742, opened 2026-09-11
feat(plugin-weaviate): add Weaviate REST driver

### LINKED ISSUE #1724: Database request: Weaviate

J2TeamNNL:
Fixes #1724

## Summary

Adds Weaviate as a registry-only REST plugin. Collections are tables, objects are rows, `uuid` is the primary key. Wire, auth and codec live in `TableProWeaviateCore` under Packages/TableProCore. `Plugins/WeaviateDriverPlugin` is a thin adapter, the same shape as Spanner after #2712.

v1:

- Connect: host, port 8080, http/https, API key as `Authorization: Bearer` or anonymous
- `GET /v1/schema` for collections and properties
- `GET/POST/PATCH/DELETE /v1/objects` for browse and row edit
- GraphQL `/v1/graphql` in the editor for vector and hybrid search
- Vectors display as a JSON number array. They are not a SQL BLOB and the inline editor does not write them back
- No FKs, routines, triggers, SSH, or SQL dialect. `tableEntityName` is Collections
- Auth field ids are Weaviate-prefixed (`wvAuthMethod`, `wvApiKey`) so they do not collide with Elasticsearch in TableProTests
- PluginKit 25. No bump. Slug `weaviate`. No `plugin-weaviate-v*` tag

## Verification

This environment has no Docker and no Swift toolchain. Live Weaviate was not measured. Fake-HTTP tests cover schema, objects, 401/500 errors, GraphQL Get flattening of `_additional.id`, GraphQL error bodies, and uuid PATCH/DELETE/POST. Registry, field-id, QueryClassifier, and engine-count tests are in TableProTests.

Docs house style and `check-docs-against-source.py` pass (35 engines). `swiftlint --strict` and `xcodebuild` could not run here.

## Not in this PR

- gRPC
- Weaviate modules beyond browse and GraphQL
- FalkorDB / NornicDB (#1986)
- PluginKit bump
- Tagging `plugin-weaviate-v*`

OceanBase (#1748) is on a separate branch and is not stacked here.

### DIFF
--- a/.github/plugin-registry.json
+++ b/.github/plugin-registry.json
@@ -415,6 +415,20 @@
       "category": "database-driver",
       "homepage": "https://docs.tablepro.app/databases/typesense"
     },
+    "weaviate": {
+      "target": "WeaviateDriverPlugin",
+      "bundleName": "WeaviateDriverPlugin",
+      "bundleId": "com.TablePro.WeaviateDriverPlugin",
+      "bundled": false,
+      "displayName": "Weaviate Driver",
+      "summary": "Weaviate driver over the REST API with a GraphQL console",
+      "databaseTypeIds": [
+        "Weaviate"
+      ],
+      "icon": "weaviate-icon",
+      "category": "database-driver",
+      "homepage": "https://docs.tablepro.app/databases/weaviate"
+    },
     "xlsx": {
       "target": "XLSXExport",
       "bundleName": "XLSXExport",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Google Cloud Spanner as a registry plugin over the REST API. (#1226, #2480)
+- Weaviate as a registry REST plugin, collections as tables. (#1724)
 - TiDB and Databend connection types on the MySQL driver. (#1066, #2514)
 - Empty state in the inspector and the assistant for a connection that is not up.
 - **Check connections** in Settings > General, including Only when I use the connection. (#2700)
--- a/Packages/TableProCore/Package.swift
+++ b/Packages/TableProCore/Package.swift
@@ -23,6 +23,7 @@ let package = Package(
         .library(name: "TableProTrinoCore", targets: ["TableProTrinoCore"]),
         .library(name: "TableProGoogleCloud", targets: ["TableProGoogleCloud"]),
         .library(name: "TableProSpannerCore", targets: ["TableProSpannerCore"]),
+        .library(name: "TableProWeaviateCore", targets: ["TableProWeaviateCore"]),
         .library(name: "TableProNumberFormatting", targets: ["TableProNumberFormatting"]),
         .library(name: "TableProR2SQLCore", targets: ["TableProR2SQLCore"])
     ],
@@ -103,6 +104,11 @@ let package = Package(
             path: "Sources/TableProSpannerCore"
         ),
         .target(
+            name: "TableProWeaviateCore",
+            dependencies: [],
+            path: "Sources/TableProWeaviateCore"
+        ),
+        .target(
             name: "TableProR2SQLCore",
             dependencies: [],
             path: "Sources/TableProR2SQLCore"
@@ -161,6 +167,11 @@ let package = Package(
             name: "TableProSpannerCoreTests",
             dependencies: ["TableProSpannerCore", "TableProGoogleCloud"],
             path: "Tests/TableProSpannerCoreTests"
+        ),
+        .testTarget(
+            name: "TableProWeaviateCoreTests",
+            dependencies: ["TableProWeaviateCore"],
+            path: "Tests/TableProWeaviateCoreTests"
         ),
         .testTarget(
             name: "TableProR2SQLCoreTests",
--- a/Packages/TableProCore/Sources/TableProCoreTypes/DatabaseType.swift
+++ b/Packages/TableProCore/Sources/TableProCoreTypes/DatabaseType.swift
@@ -40,12 +40,13 @@ public struct DatabaseType: Hashable, Codable, Sendable, RawRepresentable {
     public static let trino = DatabaseType(rawValue: "Trino")
     public static let kafka = DatabaseType(rawValue: "Kafka")
     public static let cloudflareR2SQL = DatabaseType(rawValue: "Cloudflare R2 SQL")
+    public static let weaviate = DatabaseType(rawValue: "Weaviate")
 
     public static let allKnownTypes: [DatabaseType] = [
         .mysql, .mariadb, .tidb, .databend, .postgresql, .sqlite, .redis, .mongodb,
         .clickhouse, .mssql, .oracle, .dameng, .duckdb, .cassandra, .redshift,
         .etcd, .cloudflareD1, .dynamodb, .bigquery, .spanner, .snowflake, .libsql, .beancount,
-        .surrealdb, .teradata, .trino, .kafka, .cloudflareR2SQL
+        .surrealdb, .teradata, .trino, .kafka, .cloudflareR2SQL, .weaviate
     ]
 
     /// Icon name for this database type — asset catalog name (e.g. "mysql-icon") or SF Symbol fallback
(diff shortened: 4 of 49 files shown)


datlechin:
Fixes the code-review findings on this PR. Every claim below was measured against Weaviate 1.39 running in Docker, not read off a documentation page.

## The root causes

**The filter builder answered "I cannot" with `nil`, and the caller read `nil` as "no filter".** An unsupported operator, an unparsable number and an unmapped data type all ended as a browse with no `where` clause, which returns the whole collection and reports success. `WeaviateFilterBuilder` now throws, the way `TypesenseFilterBuilder` does, and the errors are localized and name the column.

**Filter types were resolved when the query was built, from a cache that can be empty.** They now resolve in `executeSearch`, which already has the collection in hand, so an `int` column can no longer be filtered as text.

**A structured property was emitted bare into the GraphQL selection set.** A collection with a `geoCoordinates`, `phoneNumber`, `object` or cross-reference property answered `Field "place" ... must have a sub selection` and the whole filtered browse failed. Each shape now gets the sub-selection Weaviate wants.

## The operator map, measured

| TablePro | Weaviate |
|---|---|
| `=` `!=` | `Equal`, `NotEqual` |
| `>` `>=` `<` `<=` | `GreaterThan` family, refused on text and boolean, which return zero rows silently |
| `CONTAINS` `NOT CONTAINS` `STARTS WITH` `ENDS WITH` | `Like` with `*`, refused on int, number, date and uuid, where the server's regex compile fails |
| `IN` `NOT IN` | `ContainsAny`, `ContainsNone` |
| `BETWEEN` | `And` of `GreaterThanEqual` and `LessThanEqual`, from `secondValue` |
| `IS NULL` `IS NOT NULL` | `IsNull` |
| `IS EMPTY` `IS NOT EMPTY` | `len(prop)` |
| `REGEX` | refused with a localized message |

Value fields follow the property's element type, since Weaviate has no array value field: `int[]` filters with `valueInt`, `date` with `valueDate`, `uuid` with `valueText`.

## Everything else

- A stored API key is no longer sent when Auth Method is None. The test that asserted the key was sent is inverted. The API Key row itself stays visible under both methods: `hidesUsername` keys off field visibility, and the dropdown rule hides a built-in row only when the value differs from the default, which is `none` here, so hiding the row would put a Username field back on a connection that has no username.
- A console path keeps its query string: `GET /v1/objects?class=Article` was resolving to `/v1/objects%3Fclass=Article`.
- `Dictionary(uniqueKeysWithValues:)` over server-supplied names is gone from both sites; a duplicate property name no longer traps the app.
- A vector component at exactly 2^63 no longer traps `Int(value)`.
- `_additional` keeps `distance`, `score` and `certainty`, and a console result shows the fields the query asked for rather than every column the collection has.
- A uuid sort reaches GraphQL, which sorts by the object id. A property type Weaviate cannot sort is passed through so the server explains, rather than painting a sort chevron over unsorted rows.
- A GraphQL operation is classified from its body, whether it is typed bare, wrapped in a `{"query": ...}` envelope, or sent under a `POST /v1/graphql` console line. The driver forwards the envelope verbatim, so all three had to be read the same way.
- Every user-facing string in the module and the plugin is localized.
- The write generator reports what it skipped, and the driver logs it, so a save that wrote nothing is no longer silent.
- A response body is parsed once instead of two or three times per browse.

## Two review findings did not survive verification

- The `sqlIsNotConsole` finding was stale: commit 33eb92e4d already requires a leading `/`, so `DELETE FROM Article` is rejected and the test passes.
- The engine-count finding was wrong: `origin/main` asserts 34 and has no OceanBase, so 35 is correct here. OceanBase (#2741) is still open, and whichever of the two merges second takes the 36 bump.

## Found while working here, not fixed

Each is real and verified, and none of them is made worse by this change. Say the word and they can follow.

- `WeaviateJSON.parsedValue` JSON-parses the value of any property whose declared type is not int, number or boolean, so a `text` property edited to `{"a":1}` is written as an object and Weaviate rejects the update while the grid reports a save.
- `properties(from:types:)` drops any property whose name starts with `_`, so a column called `_source` is silently not written.
- An insert posts an explicit `null` for every column left blank, which is not the same as leaving a property unset for a cross-reference or an `object`.
- `ping()` calls `/v1/.well-known/ready`, which Weaviate serves without authentication, so a revoked API key leaves the health monitor reporting a healthy session while every query answers 401.
- `WeaviateConsoleParser` drops anything after the path on the same line, so `POST /v1/objects {"class":"Article"}` is sent with no body.
- Both browse paths always fetch the vector, even when the grid is not showing that column.
- The `/v1` prefix is added for four path roots only, so `GET /nodes` and `POST /batch/objects` 404.
- `offset + limit` above `QUERY_MAXIMUM_RESULTS` (10,000 by default) is a hard error on both transports, so paging past row 10,000 fails.

## Verification

- 63 tests in `TableProWeaviateCoreTests`, including the new filter, path, selection-set and write-skip suites.
- The generated GraphQL for 16 filter shapes, a uuid sort, and a collection carrying `geoCoordinates`, `phoneNumber`, `object` and cross-reference properties, each run against Weaviate 1.39 in Docker and returning the right rows.
- `swiftlint --strict` clean over the app, the plugin, the package and the tests.
- Both docs checks pass.

No UI automation: the change is a driver, and every path is reachable from a unit test.

datlechin:
The eight defects listed as "found while working here, not fixed" are now fixed too, in `f6e1c9b9e`.

**A `text` cell is written as text.** `WeaviateJSON.parsedValue` ran `JSONSerialization` over every value whose type was not int, number or boolean, so a `text` property edited to `{"a":1}` went to the server as an object and Weaviate rejected the update while the grid reported a save. Only a type the grid actually renders as JSON is parsed back now: arrays, `object`, `geoCoordinates`, `phoneNumber` and cross-references.

**A property named with a leading underscore is written.** The write path dropped any column starting with `_`, so a property called `_source` was silently never saved. The guard now names the one synthetic column it exists for, `_additional.`.

**An insert no longer posts a null for every blank column.** A new row in a collection with ten properties was sending `{"title":"x","author":null,"refs":null,…}`. For a cross-reference or an `object` property, an explicit null is not the same as leaving the property unset. Only the columns the user filled are sent.

**A ping asks something that checks the key.** `ping()` called `/v1/.well-known/ready`, which is the readiness probe. Measured against a Weaviate started with `AUTHENTICATION_APIKEY_ENABLED`: `ready` answers 200 with no key at all, while `/v1/meta` answers 401 without one and 401 with a wrong one. A revoked or rotated key left the health monitor reporting a session that every query then failed on. `ping()` reads `/v1/meta`.

**A console body typed on the request line is kept.** `POST /v1/objects {"class":"Article"}` was parsed into method and path with the payload dropped, then sent with no body. The parser takes the rest of the line as the body when there is nothing below it, and `QueryClassifier` reads the same two places so the read-only gate cannot disagree with what the driver sends.

**Any console path resolves under `/v1`.** The prefix was added for four path roots by name, so `GET /nodes` and `POST /batch/objects` reached the base URL and answered 404. Every path that is not already `/v1` gets it.

**A browse fetches the vector only when the grid is showing it.** Both paths asked for it unconditionally: `GET /v1/objects` with `include=vector`, and the GraphQL `Get` with `_additional { id vector }`. On a 1536-dimension collection that is megabytes of floats per page, decoded and stringified, to be thrown away.

**The 10,000-row paging cap is documented.** Weaviate refuses an offset and limit that add up past `QUERY_MAXIMUM_RESULTS`. The server's own message names the setting and the numbers, so it is left to report it; the limit is now in the page's Limitations list rather than being a surprise. Cursor paging with `after` would be a feature, not a fix.

## Verification

73 tests in `TableProWeaviateCoreTests`, 30 in `TableProTests`, `build` passes, and the Weaviate plugin compiles and packages in the `AllPlugins` aggregate. `swiftlint --strict` and both docs checks are clean.

One gap worth naming: the insert change lives in `WeaviatePluginDriver+Metadata.mappedChange`, and the test target does not link the plugin, so it has no unit test. The core behaviour it feeds is covered.

PR #2784, opened 2026-09-12
fix(datagrid): clip the row-number strip to the grid's scroll view

datlechin:
## Before / After

Before and after: the find field with the row number painted into it, then intact

The find field reads `1 nd in loaded rows` before: the strip's row number sits on the magnifier glyph and its divider crosses the field. Nothing of the grid is above the grid after. The faint two-line ghost of the rows inside the column header band is separate and unchanged, and is not this fix (see below).

## What was wrong

The data grid pins its row-number strip with `NSScrollView.addFloatingSubview(_:for: .horizontal)`, and the strip is **document-tall** (24,000pt on a 1,000-row page) so the numbers stay level with their rows at every scroll offset.

Nothing clipped it. Measured on macOS 27, AppKit reparents a floating subview into a `_NSScrollViewFloatingSubviewsContainerView` whose `clipsToBounds` is `false`, and `NSScrollView`'s is `false` too, which is `NSView`'s documented default for apps built against the macOS 14 SDK. The chain from the strip to `NSThemeFrame` therefore had no clip in it at all.

So the strip painted its row numbers and its column separator straight out of the grid and onto whatever the window stacked above it. Everything above the grid is an earlier sibling in the same `VStack`, so it is drawn first and the strip lands on top of all of it: the find bar, the filter panel, the key-pattern search bar, the result tab bar and the banners.

## The fix

`scrollView.clipsToBounds = true`, set where the strip is installed.

The scroll view is the view this code owns, which is why the clip belongs there. Clipping the floating container instead would reach into a view AppKit owns and would stop working the day AppKit reparents somewhere else.

## Measured, in the running app

Attached to a Debug build showing the Chinook sample's `Track` table with the find bar open and the grid scrolled down:

| | result |
| --- | --- |
| Hiding the strip | changes exactly the band at pt 52.0..88.5, above the header |
| `clipsToBounds = true` on the floating container | changes exactly that same band |
| `clipsToBounds = true` on the scroll view | changes exactly that same band |
| The strip inside the viewport, before and after | identical, to the pixel |

So the clip removes the escaped painting and nothing else.

## Tests

Two in `DataGridRowNumberRenderingTests`, both of which fail with the fix reverted (checked):

- `stripPaintsNothingAboveTheGrid` hosts the grid under a band, the shape the window has, scrolls down, rasterises the band and requires its leading 120pt to match the band beside it. It holds the band to its own colour rather than to a literal, because the cached drawing lands in an HDR colour space that shifts the fill.
- `scrollViewClipsToBounds` guards the flag itself.

`makeGrid` grew a `stripAbove:` parameter that also routes the install through `DataGridView.installRowGutter`, so the test exercises the production path instead of a copy of it. At the default of `0` every existing case builds exactly as before.

Verified: `DataGridRowNumberRenderingTests`, `DataGridRowGutterTests` and `SortableHeaderRenderingTests` pass (33 cases), `swiftlint --strict` reports 0 violations over `TablePro` and `TableProTests`.

## Also in here

`TableProTests/Views/Results/DataGridPendingChangeMarkTests.swift` passed `nil` for `DataGridCellPalette.booleanTrueText` and `booleanFalseText`, which `3d69ab597` made non-optional without updating this caller. It broke the whole test target's compile on `main`, so nothing in `TableProTests` could run until it was fixed. Two lines, `.labelColor` each, matching what `DataGridCellPalette.standard` uses.

## Not fixed here

The faint ghost of the rows inside the column header band, visible in both screenshots. Measured, that is AppKit's own header material sampling the content scrolled under it: a stock `NSTableHeaderView` on macOS 27 leaks pure magenta content at alpha 0.139, and the grid's opaque `windowBackgroundColor` fill already cuts it to 0.043 in the running app. Suppressing the rest means not calling `super.draw` from `SortableHeaderView.draw` and drawing the header cells directly, which is a separate change to header rendering.


datlechin:
Before / After, same window, same table, same scroll offset.

PR #2786, opened 2026-09-12
fix(hig): show the whole statement in the query confirmation dialog

### LINKED ISSUE #2759: Write confirmation modal too small
### What happened?

This query execution confirmation dialog is useless when all I see is a single word. I have TablePro queries executing via MCP. So I don't have the UI open and I cannot see the whole SQL in TablePro.

### Steps to reproduce

Turn on write confirmations and execute a multi-line query.

### Expected behavior

Should show the entire SQL query in a scrollable box so that I can actually confirm the query I'm approving.

### Database type

_No response_

### TablePro version

0.71.0

### macOS version & chip

macOS 26.6.2 & M5 Pro

### Screenshots / Logs

Image

datlechin:
Fixes #2759.

## The bug

The write confirmation put the statement into `NSAlert.informativeText`, a proportional label with no scrolling and no selection, after cutting it to 200 characters. Measured on this SDK: an alert with no accessory is 260x223pt, and the label does grow for long text (520x1613pt for 40 lines of SQL) without ever scrolling. The cut and the container are one defect, not two.

The reporter's screenshot shows a second half of the problem. `confirm(sql:operationDescription:connectionId:isDestructive:)` was handed neither the caller nor the connection, so the dialog could not say who was asking or where. Their "BEGIN" was not truncation: `TransactionControlTool` passes `sql: action.uppercased()`, and its own label ("transaction begin") plus the connection name reached only the MCP elicitation path. What they saw was "MCP query execution" over one word, for *Claude beginning a transaction on their database*.

This path is shared. Around 25 call sites reach it: the query editor, structure and schema apply, create table, rename, row import, users and roles, maintenance, rewind, AppleScript, and MCP. Every one of them was hiding the statement it asked about.

## The fix

The app already had exactly one surface for "review these statements, then Execute or Cancel": `SQLReviewSheet`, which is the Users & Roles apply confirmation and the table-rebuild review. It shows the full statement, syntax highlighted, scrollable, selectable, with Copy All. The confirmation now uses it.

It is hosted through a new `AlertHelper.runStatementConfirmation`, built on the `runPairingApproval` pattern that already handles a decision with no window open: a critical sheet on the window the user was working in, `NSApp.runModal` when there is none. That second path is the ordinary case for MCP, not a fallback, because the request arrives while TablePro may have no window at all.

- `PairingApprovalGate` becomes `ModalDecisionGate<Value>`, so pairing and confirmation share one resume-once gate instead of two copies.
- `SQLReviewSheet` drops `@Environment(\.dismiss)`, which is inert once hosted, in favour of the `isPresented` binding that was declared and never read. One dismissal channel.
- `OperationConfirming.confirm` takes an `OperationConfirmationRequest` carrying the caller and a resolved connection name. `DefaultExecutionGate` gains a `connectionNameResolver` closure beside its two existing resolvers, so the provider stays the composition root.
- A rename passes `sql: nil` and has no statement to show. It keeps a plain alert rather than rendering an empty box, and no longer trails a blank line under "Are you sure you want to execute this query?".

## Before / After

Before is the reporter's own screenshot from #2759, on 0.71.0. After is this branch, driven against the bundled sample database with Safe Mode set to Alert.

| Before (#2759) | After |
| --- | --- |
| Confirmation dialog reading MCP query execution, Are you sure you want to execute this query, BEGIN | Confirmation dialog showing a fourteen line UPDATE in full |

The same pair is now on `docs/features/safe-mode.mdx`, which had no picture of this dialog.

## Two more fixes in the same subsystem

**The MCP elicitation channel had the same bug.** `MCPStatementGate.preview` flattened every newline and cut at 400 characters, and that text is the only thing an elicitation-capable client (Claude Desktop) shows beside its Approve control, after which TablePro's own dialog never runs. The cut landed before the `WHERE` clause of any statement worth confirming. It now keeps its line breaks and caps at 2,000, sized for a form field in someone else's interface.

**A token without Full Access could approve its own write.** `MCPAuthPolicy.checkSafeModeDialog(principal:)` strips `.confirmationPreCleared` for a non-admin or anonymous principal, and two doc comments in that file assert the stripping as an invariant. It had zero callers: `MCPStatementGate` called the overload that strips nothing, so any `tools:write` token declaring `elicitation` could answer its own prompt and TablePro's dialog never appeared. `MCPStatementGate` now passes its principal. `MCPStatementConsentGuardTests` covers the capability logic and scans the source for the call, because the defect was a missing argument and a behavioural test cannot see one.

## What I built and tested

- `verify.sh build`: PASS
- `verify.sh test` over the ten suites that own the changed types: PASS
- `verify.sh lint TablePro TableProTests`: PASS, 0 violations
- `verify.sh docs`: PASS
- Driven against a running Debug build on the bundled sample database: the dialog shows all 14 lines of a multi-line `UPDATE`, Cancel denies the statement, Execute runs it, and closing the window cancels.

New tests: `OperationConfirmationPromptTests` (17 cases) and `MCPStatementConsentGuardTests` (4). `SQLReviewSheetTests`, `ExecutionGateTests` and `MCPStatementGateTests` updated in the same commit.

No `TableProUITests` case. Raising this dialog needs a live connection at a confirming Safe Mode level, which is not deterministic on CI.

## Review findings addressed

A Codex defect review, a Codex adversarial review and a security review each read the diff. Six findings, all fixed:

1. **Execute took the Return key for ordinary writes.** `AlertHelper.addConfirmAndCancel` strips Return from the confirming button on purpose, and `SQLReviewSheet` gives Execute `.defaultAction` whenever the statement is not classified destructive. Since `isDestructive` covers only `DROP`, `TRUNCATE` and a `DELETE` with no `WHERE`, every ordinary MCP `UPDATE` would have landed in a focus-stealing dialog where a stray Return meant Execute. `PrimaryAction.takesDefaultAction` is false for every externally raised confirmation.
2. **Execute was live while the statement was still rendering.** Now disabled until `prepared` exists.
3. **The review sheet rewrote MongoDB `$oid` to `ObjectId()` and appended `;`**, so the approved text differed from the text that runs. `showsStatementsVerbatim`.
4. **A confirmation still truncated at 20,000 characters**, so a large literal before the `WHERE` clause hid the predicate; `execute_query` accepts 102,400 units. A confirmation never truncates, and the large-text path uses `HighlightedSQLTextView` rather than a SwiftUI `Text`, which lays out for seconds at that size.
5. **The windowless Execute deadlocked.** `runHostedDecision` holds the main actor inside `NSApp.runModal`, which returns only once the button resolves the gate, but the button deferred the answer to a `Task` that inherits the main actor and cannot start until `runModal` returns. `PrimaryAction.Work.immediate` answers inline; applying a plan keeps the asynchronous form.
6. **A source-scanning guard test resolved the wrong path** and would have failed on every checkout.

## Known limitations

Escape does not dismiss the hosted sheet under synthetic input, and I could not establish whether a real keypress does. A local `.keyDown` monitor installed over the sheet logs Tab and every letter and never logs key code 53, so AppKit turns Escape into `cancelOperation(_:)` before any monitor sees it, and the harness cannot deliver it either way. `ModalDecisionWindow` overrides `cancelOperation(_:)`, which is the end of that responder chain, and Cancel, the close button and the window delegate all cancel correctly.

MCP request cancellation does not tear down a hosted decision: neither branch observes task cancellation, so a prompt outlives the request that raised it. This is not a regression, the `NSAlert` path it replaced called `alert.runModal()` on the same branch, and it is reported rather than fixed here.

https://claude.ai/code/session_01CsJzH4VK7mTEBBye3f9LhL


datlechin:
After, driven against the bundled sample database with Safe Mode set to Alert.

PR #2788, opened 2026-09-12
style(datagrid): bring the theme slot switch and the grid coordinator back under the lint limits

datlechin:
`swiftlint lint --strict` fails on `main` with two errors from #2781:

```
TablePro/Theme/ThemeDefinition.swift:212:1: error: Vertical Whitespace before Closing Braces
TablePro/Views/Results/DataGridCoordinator.swift:32:7: error: Type Body Length: Class body should
    span 1100 lines or less excluding comments and whitespace: currently spans 1101 lines
```

PR CI never runs that. The `lint` job lives in `build.yml`, which only fires on a `v*` tag, so this is
waiting at the next release rather than on any pull request.

The blank line goes. For the coordinator, `visualState(for:)` moves next to `visualState(of:atDisplayRow:)`,
which already lives in `DataGridView+RowIdentity.swift`: the two are the same question asked with and
without a resolved row, so the pair belongs together and the class drops back under the limit without a
line being invented or deleted.

`swiftlint lint --strict` is clean and the app builds.

### DIFF
--- a/TablePro/Theme/ThemeDefinition.swift
+++ b/TablePro/Theme/ThemeDefinition.swift
@@ -209,7 +209,6 @@ internal enum ThemeSlot: String, CaseIterable, Sendable {
         case .statusSuccess: return \.status.success
         case .statusWarning: return \.status.warning
         case .statusError: return \.status.error
-
         }
     }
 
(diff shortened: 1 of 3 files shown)


datlechin:
Obsolete. Both violations came from #2781, which #2790 reverted out of `main`, and `swiftlint lint --strict` is clean there now. If the theme engine work returns, the `ThemeDefinition.swift` blank line and `DataGridCoordinator`'s 1101-line body come back with it; the fix is the one in this branch's diff.

PR #2806, opened 2026-09-12
feat(datagrid): add a Map view for results holding a geometry column (#2532)

### LINKED ISSUE #2532: Map view for spatial results
### Problem

A query returning geometry shows WKB hex or WKT text. PostGIS, MySQL spatial, SQL Server geography and DuckDB spatial all produce data that only means something on a map, and there is nowhere to see it.

### Proposed solution

- A Map view beside the grid and the JSON view, shown only when the result holds a geometry column.
- Points, lines and polygons drawn with MapKit, over Apple's own tiles, so nothing is sent to a third-party tile server.
- Click a shape to select its row in the grid, and select a row to highlight its shape.
- Fit to the result's extent, and say plainly when the SRID is not one that can be projected.

### Alternatives considered

Paste WKT into a web viewer, which sends the data to someone else.

### Related database type

PostgreSQL

datlechin:
Fixes #2532.

A result holding a geometry column gains a **Map** segment beside Data, JSON and Chart. Points,
lines and polygons draw over Apple's own tiles, clicking a shape selects its row, and the pane says
what it could not draw and why.

## Why AppKit `MKMapView` and not SwiftUI `Map`

Measured, not preferred. Grepping every MapKit header: **MapKit publishes no overlay selection and
no overlay hit-test API at any layer.** The only selection API in the framework is
`selectAnnotation:` / `selectedAnnotations` / `didSelectAnnotationView:` (`MKMapView.h:201-203`,
`289-294`), and it covers annotations only. A click on a polygon or a line can be resolved only
through `MKOverlayPathRenderer.path` plus `MKOverlayRenderer.point(for:)`, and the renderer is
reachable only via `MKMapView.renderer(for:)`. SwiftUI's `Map` exposes neither its map view nor its
renderers, so a tagged `MapPolygon` can never report a hit. Two further blockers: `MapSelection` is
macOS 15+ (`_MapKit_SwiftUI.swiftinterface:848`), and SwiftUI's `MapPolygon` has no
`interiorPolygons` initializer (`:547`), so it cannot draw a polygon with a hole at all.

Scale says the same thing. Measured on this machine, adding overlays to a live `MKMapView`:

| shape | 5,000 rows | 20,000 rows | 50,000 rows | 200,000 rows |
|---|---|---|---|---|
| one `MKPolygon` per row | 1.32 s | 17.3 s | **125.3 s** | not attempted |
| one `MKMultiPolygon` | 0.003 s | 0.009 s | 0.022 s | **0.089 s** |

A result page holds up to 100,000 rows, so every polygon row goes into one `MKMultiPolygon` and
every line row into one `MKMultiPolyline`, with the selection as a single extra overlay swapped on
click (0.170 ms measured) and points as clustered annotations.

## Privacy, which is the request's premise

Measured: a process showing a fully rendered `MKMapView` holds **zero network sockets**. Tiles are
fetched out of process by `com.apple.geod.xpc`. Nothing about a result reaches a third-party tile
server.

## The reader

One shared SwiftPM target, `Packages/TableProCore/Sources/TableProGeometry`, because the MySQL
plugin already contained a second hand-rolled WKB parser that had to agree with it. It reads WKT,
EWKT, WKB, EWKB hex and GeoJSON, and every spelling in its test table came off a live server during
the investigation: PostGIS 3.6.4, MySQL 8.4.11, MariaDB 12.3.3, DuckDB 1.5.4. They disagree with
each other, which is why the table is the specification:

- PostGIS writes `SRID=n;` only when the SRID is non-zero, writes XYZ as a bare `POINT(1 2 3)`, XYM
  as `POINTM(1 2 3)` rather than the OGC `POINT M (`, and MULTIPOINT bare.
- MySQL writes the parenthesised `MULTIPOINT((1 2),(3 4))` and an invalid `GEOMETRYCOLLECTION()`.
- DuckDB puts a space before the paren: `POINT (1 2)`.
- PostgreSQL keeps raw EWKB hex whenever its `ST_AsEWKT` rewrite fails, with the column still typed
  `geometry`, so hex has to read too.

## The MySQL axis-order trap

MySQL 8.4's `ST_AsText` prints a SRID-4326 point **latitude first**, and
`ST_GeomFromText('POINT(-122.4194 37.7749)', 4326)` fails with *"Latitude -122.419400 is out of
range"*. But the **stored bytes are longitude first**, byte-identical on MySQL 8.4.11 and MariaDB
12.3.3:

```
E6100000 01 01000000 50FC1873D79A5EC0 D0D556EC2FE34240
SRID 4326, little-endian, Point, x = -122.4194, y = 37.7749
```

So TablePro's MySQL geometry text was already correct and needs no reordering. "Make the parser
agree with `ST_AsText`" is the plausible-looking change that would move every MySQL point into the
Southern Ocean, and `GeometryWKBParserTests.storageIsLongitudeFirst` exists to stop it.

## SRID handling

4326, 4269 and 4979 draw directly. Web Mercator (3857, 900913, 3785, 102100, 102113) is inverted by
closed form, verified against PROJ 9.8.1 to 1e-9, with no projection library. A column carrying no
SRID draws when every coordinate fits the longitude/latitude envelope, and says so. Anything else is
refused **by name**, with the remedy:

> SRID 32633 is a projected coordinate system. Maps places longitude and latitude only, so these
> shapes cannot be drawn. Query ST_Transform(geom, 4326) to see them.

Mixed SRIDs in one column draw the majority and state the count of the rest. The map never rewrites
the cell: the grid's text stays authoritative.

Coordinates are validated before any shape is built, because `MKMapPoint(coordinate)` returns the
sentinel `(-1, -1)` outside the valid range rather than wrapping, which is garbage outside
`MKMapRectWorld` and would destroy the extent fold.

## Shipped alongside, because the feature needs them

- **`geo_point` classified as `.integer`.** `classifyByPattern` tests `hasSuffix("INT")` and
  `GEO_POINT` ends in `INT`. Confirmed by compiling and running the real classifier. It drove cell
  alignment, sort comparators and filter operators for every Elasticsearch `geo_point` column, with
  or without a map, so it carries its own CHANGELOG entry.
- **`geomcollection` classified as text.** MySQL 8.0.11 renamed the type; measured on 8.4.11, its
  catalog reports `geomcollection` for a column declared either way, and neither spatial list had it.
- **MySQL geometry lost its SRID**, so nothing downstream could decide whether the coordinates were
  degrees. Two more parser bugs went with it: an empty collection rendered as the invalid
  `GEOMETRYCOLLECTION()`, and a Z or M ordinate desynchronised the cursor.
- **A tab stranded in a mode whose columns vanished.** `ResultsModeAvailability.modes` returns `[]`
  for a query tab with no columns, which hides the switcher and disables all four View menu items
  while still ticking the current one. The items carry no key equivalent, so nothing was left to
  press. A succeeding non-SELECT does this on every run. The mode now reconciles where rows are
  installed.
- `GridSelectionOwner.resolve` was an if-chain naming `.chart` as an exception list of one, so a
  fifth mode fell through and silently claimed the data grid's channel. It is now an exhaustive
  switch that cannot compile without an answer.

## What decides whether Map is offered

The column's **values**, not its declared type, and this is the review's largest change. The type is
wrong in both directions. Oracle's `SDO_GEOMETRY`, Teradata's `ST_GEOMETRY` and SQL Server's
`geography` all classify `.spatial` and hand over text no reader can draw, so a type-only gate
offered Map and drew nothing. MongoDB goes the other way: its GeoJSON arrives in a column the app
types `JSON`, so a type-only gate never offered Map for the engine whose spatial data is most often
GeoJSON.

`SpatialColumn.columns(in:)` now samples up to four values per candidate column. The two
declarations carry different weight, because they say different things:

- **`.spatial`** is the engine's own word that the column holds geometry, so it is believed unless
  the values contradict it. An empty result and an all-NULL column both keep the segment, which is
  what stops the mode appearing and vanishing as pages load.
- **`JSON`** says nothing about geometry, so it earns the segment only by holding a value that reads.

The decision is per column and never per value, the rule the MongoDB UUID codec exists for, and it
is taken once where the rows are installed rather than on every status-bar render: `TabDisplayState`
carries the answer and `MainContentCoordinator.installTableRows` is the only writer.

Working now, with no plugin release: PostgreSQL/PostGIS, MySQL/MariaDB, MongoDB, Snowflake,
Elasticsearch and ClickHouse. SQL Server, Oracle, DuckDB, SpatiaLite, Teradata, Trino, Spanner,
Cassandra and BigQuery hand geometry over in formats the reader does not cover, so their columns
carry no segment. That sentence is now enforced by the code rather than asserted by the docs, which
is what keeps it from going stale the next time a driver changes its text format.

## Two defects found by building it, not by reading it

Both were in this branch and both are fixed here.

**A click on a shape never reached the app.** An `NSClickGestureRecognizer` added to `MKMapView`
has to win an arbitration against MapKit's own pan and zoom recognizers, and it does not: measured,
the action never fired, with and without a delegate answering `shouldRecognizeSimultaneouslyWith`.
`ResultMapSurface` overrides `mouseUp` instead, which is `NSResponder`'s own path and needs no
arbitration, and it calls `super` so panning and zooming are untouched. A press that moves more than
three points is a pan, not a click.

**A hole counted as part of its polygon.** `CGPath.contains` defaults to the non-zero winding rule,
and a polygon's interior rings wind the same way as its exterior, so a click in the middle of a hole
selected a shape that was not under the pointer. Measured:

| point | winding (default) | even-odd |
|---|---|---|
| inside hole | `true` | `false` |
| inside ring | `true` | `true` |
| outside | `false` | `false` |

`MKPolygonRenderer` fills with even-odd for the same reason, which is why the hole looks empty on
screen. The hit test now matches what is drawn.

The reason both shipped in the first place is worth naming: the hit test lived inside the
coordinator, where only a running app could reach it. It is now a pure `ResultMapHitTesting` with a
headless suite, and that suite is what caught the hole.

A third followed from them: the status bar refused to report a selection in Map mode, because
`reportedSelection` keyed off `showsColumnControls`, which is about column chrome. Map both reads
the selection and sets it, so the count is now gated on a dedicated `reportsRowSelection`.

## What the review changed

Fourteen findings, six of them P1, all fixed.

**The iOS target did not compile.** The mobile app builds
`Plugins/MySQLDriverPlugin/GeometryWKBParser.swift`, which now imports `TableProGeometry`, and
`TableProMobile/project.yml` did not list that product. `TableProGeometry` builds for
`generic/platform=iOS Simulator` on its own; the mobile app target still stops on the pre-existing
`@TaskLocal` macro error inside the vendored `oracle-nio` checkout, which predates this branch.

**A map click did not survive the trip to Data mode.** The shared selection channel alone is not
enough: the grid remounts and restores the tab's own stored selection over it, which a map click
never wrote. The `.map` arm now stores it too, with an empty cell rectangle because a shape names a
row and no columns.

**Two of the three paths that replace a result skipped the mode reconciliation.**
`seedBufferFromActiveResult` and `applyResultSetSwitch` both wrote the rows straight into the
registry. All three now go through one `installTableRows`, which is also where the geometry columns
are decided, so the two answers cannot drift apart.

**The selection was one overlay per row.** Measured, that is 1.32s at 5,000 rows and 125s at 50,000,
the same curve the base layer avoids, and it made the renderer lookup a linear scan of the selection
on every overlay MapKit draws. The highlight is now one `MKMultiPolygon` and one `MKMultiPolyline`,
told apart from the base layer by identity rather than by class.

**Hit-testing ran in the wrong order.** The polyline overlay is added after the polygon one, so a
line drawn across a polygon is what the reader is pointing at, and testing polygons first selected
the shape underneath. Point rows now highlight too, marker and cluster both.

**Budgets were checked once per row.** One row can be a MULTIPOINT or a GEOMETRYCOLLECTION of any
size, so the first row alone could put millions of shapes on the map. The budget is spent shape by
shape, and a row it cuts short is reported as capped as well as drawn.

**Nesting had no bound.** Every wire format lets a geometry collection nest without limit and each
level is one more stack frame, so a few hundred bytes of nothing but collection headers would
exhaust the stack rather than fail. All three readers and the projector refuse past
`SpatialLimits.maximumNestingDepth`.

**The pane blamed the coordinate system for a parse failure.** `projectability` carries
`.unsupported(srid: nil)` as its own default, so a column where nothing parsed was indistinguishable
from one whose SRID cannot be projected, and a column of `CIRCULARSTRING` was told its coordinates
were out of range. `readableRows` is asked first.

**Three wire formats were read wrong or not at all.** ClickHouse writes a `Tuple` in parentheses and
an `Array` in square brackets, which `ClickHouseTabSeparatedRowDecoderTests` already pins, so every
`Ring`, `Polygon` and `MultiPolygon` was rejected before parsing. Elasticsearch's `geo_point` is
legally six shapes and only `"lat,lon"` was handled; the object, array and geohash forms are read
now, with the axis order that each one actually uses. And `geo_shape` values were cut off at 10,000
characters by the flattener, which made the JSON unparseable, so a geometry is kept whole.

**The reconciliation UI test could pass while reconciliation was broken.** Its second query opened a
fresh tab with Cmd+T, and a fresh tab starts on Data. It now runs in the tab that was on Map.

## Verification

- `verify.sh build`: PASS.
- `swift test --filter TableProGeometryTests`: 75 passed, 0 failed.
- `verify.sh test` over the affected suites: 102 and 203 executed, all passed.
- `xcodebuild -scheme TableProGeometry -destination 'generic/platform=iOS Simulator'`: BUILD
  SUCCEEDED.
- `verify.sh uitest ResultMapModeUITests`: 6 executed, 6 passed. The Map segment appears only with a
  geometry column, a tab left on Map lands back on Data in the same tab, and **clicking a shape
  selects its row**.
- `verify.sh plugins` (the `AllPlugins` aggregate) fails only on the pre-existing `@TaskLocal` macro
  error inside the vendored `oracle-nio` checkout. No error touches anything here, and `MySQLDriver`
  built alone succeeds with zero errors.
- `swiftlint lint --strict` over every new file, including the package and the plugin, which a bare
  run does not see.
- `docs/scripts/check-writing-style.sh` and `check-docs-against-source.py`: both pass.

## Screenshot

The docs section ships without its `<Frame>`. The feature was driven end to end against a SQLite
database with a `POLYGON`-declared column, and the capture is attached to this PR: six San Francisco
neighbourhoods drawn over Apple Maps in the right places, with `Drawing 6 shapes in SRID 4326.` in
the toolbar. A matched light and dark pair at 1560x960 is what `docs/STYLE.md` requires and what this
machine could not produce, so the page carries prose alone rather than a mismatched image.

## Accessibility

`MKMapView` reports itself as a single image and publishes no children for overlays (measured), and
MapKit has no accessibility API for them. A SwiftUI `accessibilityLabel` on the representable is not
readable from the accessibility tree either (measured: empty). The map is therefore supplementary and the data grid stays
the readable representation of spatial rows; the map view carries a summary label, and point
annotations speak for themselves. This is stated in the docs rather than left for a user to discover.

https://claude.ai/code/session_01GdkP9Kj2oVdzM5acwU69A3


datlechin:
Driven end to end against a SQLite database with a `boundary POLYGON` column holding EWKT: six San Francisco neighbourhoods, drawn where they belong, the toolbar reporting `Drawing 6 shapes in SRID 4326.` and Map selected in the switcher.

Map mode showing six filled polygons over Apple Maps in San Francisco

The docs page ships without its `<Frame>`. `docs/STYLE.md` requires a matched light and dark pair at 1560x960, and this machine could not produce the light half, so the page carries prose rather than a mismatched image.

PR #2812, opened 2026-09-13
Rebuild the appcast pipeline on a tested merge

datlechin:
Preparation for turning on Sparkle's background install. That change removes the user's chance to decline a bad build, so the publishing path and the rollback path both have to be trustworthy first.

## The seed is a live hazard

`sign-and-appcast.sh` handed the whole published feed to `generate_appcast` as a seed. That is safe only while the staging directory holds nothing but the new archive, and generating deltas requires staging a second one.

Verified against the pinned Sparkle 2.9.5 source that CI downloads. `FeedXML.swift:333-339` matches an existing item by version and sets `createNewItem` from it, and two writes then run outside that guard:

- Line 603 assigns the enclosure's attributes unconditionally, so a seeded item matched by a staged archive has its download URL rewritten into the current release's directory. Permanent 404.
- Lines 523-525 remove that item's `<description>` when the staged archive has no sibling notes file.

Both pass the rewind guard at `build.yml:311-317`, which compares only the set of version strings.

`generate_appcast` now sees one release's archive and nothing else. `scripts/ci/merge-appcast.py` splices its two items into the published feed afterwards.

## What replaced the awk merge

The old merge was a line-oriented state machine over XML resting on three untested assumptions, one of which was that no release note ever contains the literal `</item>`. Release-note text flows into the CDATA it scanned.

The new merge moves untouched items byte for byte rather than re-serializing them, because a full ElementTree round trip would rewrite namespace prefixes and 640 KB of CDATA. Seven invariants, each failing the release:

- exactly one item per architecture, for the version being released
- arm64 carries `sparkle:hardwareRequirements`, x86_64 does not, arm64 goes first
- both architectures agree on `sparkle:version`, and it is above every build in the feed
- the feed does not already advertise the version
- every enclosure URL, deltas included, is under this release's download prefix
- every enclosure carries a `sparkle:edSignature`
- nothing outside the two new items changed

## The signature invariant is not theoretical

`Appcast.swift:199` signs an archive only when the app's own Info.plist declares `SUPublicEDKey`. Measured with the pinned 2.9.5 binary against a bundle without it: `generate_appcast` writes an unsigned enclosure, prints "Wrote 1 new update" and exits 0. Every installed updater would then refuse the download with nothing said anywhere, and the rewind guard would not notice.

## Withdrawing a release

Removing the items from the feed is the only way to un-ship a Sparkle release. `scripts/ci/pull-release.py` does it with the same primitives the merge inserts with, so the two cannot disagree about where an item begins and ends. It has a `--dry-run`, refuses a version the feed does not carry, and refuses to empty the feed.

The GitHub release stays up so direct download links keep resolving.

## Release gates

- The release job now fails when `CURRENT_PROJECT_VERSION` is not above the newest **published** release's. That number is what Sparkle compares, and only a sentence in a skill file guarded it. Compared against the newest published release rather than the newest tag, because re-cutting a failed release legitimately leaves a tag at the same build. Verified against real data: a `v0.75.0` tag on today's tree would fire it.
- A ninth release pre-flight item reports the last full-suite verdict on `main`. It reports and does not block, because `main`'s conclusive runs are red far more often than green on merge skew, and a hard gate with no merge queue behind it would stop releases rather than improve them.
- `git tag` becomes `git tag -a` in the runbook. The history mixes both kinds, and a lightweight tag answers `git tag -l --format='%(contents)'` with the commit message.
- The version field in the bug report template becomes required, since the version spread widens once installs are silent.

## The dead sed

`sign-and-appcast.sh:146-149` rewrote download URLs, blaming Sparkle for ignoring `--download-url-prefix`. `beed97837` added that sed and the prefix's trailing slash in the same commit, so the slash already fixed it. Zero URLs in the published feed match the sed's pattern. Removed.

## Verification

- 24 new tests, wired into Repo Hygiene on the free Linux runner. Only a real release ran this code before, so nothing on a pull request ever exercised the script that edits the file every install polls.
- End-to-end rehearsal with the pinned Sparkle 2.9.5, a throwaway Ed25519 key, ad-hoc signed Mach-O bundles for both architectures, against the real 155-item published feed: two signed items in, arm64 first, `<description>` on both, every other byte identical.
- `pull-release.py --dry-run` rehearsed against the live feed.
- `actionlint`, `shellcheck --severity=warning`, and `verify.sh docs` all clean.

No CHANGELOG entry: release tooling, no user-visible behaviour change.

https://claude.ai/code/session_01UVsgACYgHrc9w1SXdDtqV5


datlechin:
CI note: `Package Tests` failed on `TrinoStatementClientTests.testCancelStopsPolling`, and `Run iOS Tests` failed. Neither is from this branch.

This PR touches nothing under `Packages/` (`git diff --name-only main...release/appcast-pipeline` confirms it) and no Swift at all. The Trino case is a cancellation-timing test: it passed 3/3 locally in 0.001s each, and took 0.728s before failing on the runner. iOS tests failed on 3 of the last 5 runs on `main`, including the tip.

Rerunning rather than chasing.

PR #2917, opened 2026-09-16
feat(connections): create and edit a PostgreSQL schema, with owner and privileges

### LINKED ISSUE #2908: Feature request: PostgreSQL schema management — create, rename, change owner and permissions
### Problem

TablePro can list and switch PostgreSQL schemas through the schema selector and **Database → Schema**, but there is no discoverable UI for creating or editing a schema.

Common schema-management tasks therefore require opening a SQL editor and manually running statements such as:

```sql
CREATE SCHEMA app_data AUTHORIZATION app_user;
ALTER SCHEMA app_data RENAME TO app_archive;
ALTER SCHEMA app_archive OWNER TO another_role;
GRANT USAGE, CREATE ON SCHEMA app_archive TO app_user;
```

This is inconsistent with the visual management already available for databases, tables, views, users, and roles. It is also easy to make mistakes with identifier quoting, the selected database, ownership, or privileges.

### Proposed solution

Add PostgreSQL schema-management actions to the schema selector and the relevant sidebar/database-tree context menus:

1. **Create Schema…**
   - Schema name
   - Optional owner/authorization role
   - Preview the generated `CREATE SCHEMA` statement before execution

2. **Edit Schema…**
   - Rename the schema
   - Change its owner
   - View and edit schema privileges such as `USAGE` and `CREATE`
   - Preview the generated `ALTER SCHEMA` / `GRANT` / `REVOKE` statements

3. **Refresh and navigation**
   - Refresh the schema picker, sidebar, database tree, autocomplete metadata, and all windows after a successful change
   - Keep the renamed or newly created schema selected when possible

4. **Safety**
   - Show the fully qualified target and active database before applying changes
   - Disable rename/edit operations for PostgreSQL system schemas where appropriate
   - Quote identifiers through the PostgreSQL driver
   - Respect the connection's Safe Mode and read-only settings
   - Show clear permission errors when the current role does not own the schema or lacks database `CREATE` privileges

The actions should be available consistently in flat, hierarchical-schema, and database-tree layouts. If rename support already exists in one layout, exposing the same capability from the schema selector would make it much easier to discover.

### Alternatives considered

Continue managing schemas through handwritten SQL. This works, but it is slower and less discoverable, and users must manually refresh the schema list afterward.

Issue #2819 covers catalog refresh after schema-changing DDL. This request is specifically for user-facing schema creation and editing controls; the refresh behavior proposed there would complement these actions.

### Related database type

PostgreSQL

### TablePro version

0.74.0 (build 130)

### macOS version & chip

macOS 27.0 / Apple Silicon

datlechin:
Fixes #2908

## What was wrong

TablePro's container lifecycle is complete for databases and half-built for schemas, and the half that exists is reachable from one of its three surfaces.

- **No create path.** `PluginDatabaseDriver.createSchemaStatement(name:)` existed but was an object-copy helper with a single caller (`ObjectCopyPlanner.swift:136`), took no owner, and no UI reached it. There was no `createSchemaFormSpec`/`createSchema` pair mirroring the database ones.
- **No owner, and privileges only principal-first.** Zero owner code existed anywhere. `PluginPrivilegeScope.schema` and `PluginPrivilegeCatalog.schemaPrivileges` already modelled schema grants and PostgreSQL already emitted `GRANT … ON SCHEMA`, but only through Users & Roles: there was no "who has what on this schema" read.
- **Three surfaces, one implementation.** Rename and Drop lived only on the database tree's menu. `DatabaseSwitcherPopover` hardcoded `showsCreateRow = supportsCreateDatabase && switchTarget == .database`, so schema mode could never create, and the `.flat` sidebar shape has no schema row at all.
- **No owner of container DDL.** Only `authorizeRename` called `ExecutionGateProvider`. Drop and Create Database called the driver directly, so Safe Mode's Alert and Touch ID tiers never fired and no audit record was written for the two operations that destroy the most.

## The fix

Container DDL gets one owner, `DatabaseManager+ContainerDDL`, in the shape `DatabaseManager+Principals.executePrincipalChanges` already uses: generate statements, derive the operation kind *from those statements*, authorize through `ExecutionGateProvider`, execute on the schema-change route, record query history, post the `CatalogEvent`. Create-schema, edit-schema, create-database and drop-container all route through it.

The driver side moves from opaque execute to **statement generation**, which is what makes the sheet's preview byte-for-byte what runs and is the genre convention (DataGrip, pgAdmin and TablePlus all show generated SQL before a structure change).

Four additive `PluginDatabaseDriver` requirements, three `DriverPlugin` capability statics, and three new transfer types. The ABI diff is additive: **zero symbols removed**, every addition carries a default. `currentPluginKitVersion` stays at the pending **31**. `git show v0.74.0:…/PluginManager.swift` reports 30 and HEAD reports 31, so this cycle's bump already landed and CLAUDE.md's once-per-cycle rule says to reuse it. All 40 bundled plists already carry 31. No registry re-release needed.

### Native design

- `New %@…` / `Edit %@…`, nouns from `schemaEntityName`, so BigQuery reads "New Dataset…". Ellipses because both open a sheet that needs more before it can complete, which is the HIG's rule.
- A modal sheet, with the statements previewed **inside it**. The HIG allows one sheet at a time, so nesting `SQLReviewSheet` was out; its renderer is now the shared `SQLStatementPreview` so the two cannot drift.
- Rename stays **inline in the outline view**, which the HIG expects of an editable outline cell and which TablePro already does. The Edit sheet's name field also renames, because the issue asks for it and pgAdmin's General tab and Finder's Get Info both do both.
- Privileges are a `Grid`, not a `Table`: the privilege set is the engine's and `TableColumnForEach` is macOS 14.4, past our 14.0 target.
- Every item is gated on a per-facet driver capability, never on `DatabaseType`.

### Engine coverage

Create + owner + privileges on PostgreSQL and PGlite; create + owner on Redshift and CockroachDB, which change an owner but cannot introspect a schema ACL (`aclexplode` is 8.4 and Redshift forked at 8.0.2). Oracle, SQL Server, Snowflake, Trino, DuckDB, BigQuery, Spanner and SurrealDB declare nothing and offer nothing: a schema on Oracle is a user, and T-SQL's `ALTER SCHEMA` is a transfer, not a rename.

## Also fixed, because the primary change is unsafe without them

- **Drop Schema and Drop Database skipped the execution gate entirely**, so no Touch ID and no audit record for the most destructive actions in the sidebar. Their only Safe Mode enforcement was a hidden menu item, which covers the read-only level alone. The gate now runs, with `confirmationPreCleared` so the sidebar's own dialog is not doubled.
- **New Database ran ungated too**, and the switcher offered it on a read-only connection.
- **Redshift, CockroachDB and PGlite advertised `supportsDropSchema` with nothing behind it.** `dropSchema` sat on `PostgreSQLPluginDriver`, a sibling of those classes rather than their superclass, so the menu item threw a raw English `NSError`. `LibPQDriverCore.swift:223` already documents this exact rule for `renameTable`. Moving `dropSchema` and `createSchemaStatement` to the protocol extension also restores the `CREATE SCHEMA` steps that Duplicate Database silently dropped on those engines.
- **Drop Schema promised to delete dependent objects on engines with no `CASCADE`.** `dropsDependentObjects` was `isSchema`; SQL Server and BigQuery refuse a non-empty schema outright, so the confirmation stated something the engine could not do. It now reads `supportsCascadeDrop`.
- **Dropping a schema left its Recent Tables entries behind**, so each one opened a tab whose query failed with "relation does not exist", surviving a reconnect and a restart. `CatalogEditAdoption.adoptContainerDrop` returned before the recents cleanup for anything but a database.

## Review

Two Codex passes read the diff cold and found fifteen findings between them. Every one was verified against the source and fixed; the ones worth naming, because they are the traps in this area:

- **`REVOKE` removes only what the executing role granted.** Two roles granting `USAGE` to the same grantee produce two ACL entries, so unchecking one box removed one entry, left the other in force, and reported success. `PluginSchemaGrant` now carries the grantor, a cell is editable only where the connected role is the sole grantor, and the rest show checked and dimmed.
- **A role can be literally named `public`.** Deciding the pseudo-group by string comparison rendered a grant to that role as `TO PUBLIC` and handed it to every user on the server. Grantee kind is now a `@frozen PluginSchemaGrantee` derived from the ACL's grantee id, which is zero only for the group.
- **Latched dirty state.** A checkbox toggled on and back off stayed "touched", so the apply-time merge overwrote a concurrent grant with the stale value and ran a `REVOKE` the preview never showed. Dirtiness is now derived by comparing against the loaded baseline.
- **A `BEGIN` on a shared session driver joins the user's open transaction.** PostgreSQL has no nested transaction, so the `COMMIT` would have taken a query tab's uncommitted writes with it and a rollback would have discarded them. PGlite is non-poolable and always lands there. The plan is now wrapped only on a pooled lease.
- **A partially committed plan could not be retried.** On a non-transactional engine a `CREATE` or `RENAME` can land before a later statement fails; a retry reissued it. The service reports the committed count and the editor rebases onto what committed.
- **`LATERAL` is PostgreSQL 9.3.** The ACL read now uses the set-returning-function-in-subquery form `PostgreSQLPrincipalQueries.schemaGrants` already uses, which older servers accept.

## Verified

| Step | Result |
|---|---|
| `build` | PASS, `.analysis/feat-schema-management/logs/build-TablePro-134250.log` |
| `test` (16 suites) | PASS, 185 of 185, `logs/test-SchemaEditEligibilityTests-134347.log` |
| `PostgreSQLDriver` scheme | BUILD SUCCEEDED |
| `abi` vs merge base | Diff reported, **0 symbols removed**, all additions defaulted: additive |
| `lint --strict` | PASS, 0 violations |
| `docs` | PASS |
| `plugins` (AllPlugins) | FAIL, pre-existing: the only error is `unknown attribute 'usableFromInlinenonisolated'` in a `@TaskLocal` macro expansion inside the vendored `oracle-nio` checkout. Unrelated to this change; the PostgreSQL target builds clean. |

New tests: `PostgreSQLSchemaStatementPlannerTests` (statement text, quoting of `my"schema` and `a; DROP SCHEMA public`, the grantor gate, PUBLIC vs a role named `public`, the pgAdmin #5926 untouched-grantee guard), `SchemaFormRulesTests`, `SchemaEditEligibilityTests`, `RecentTablesStoreSchemaClearTests`, six new `DatabaseTreeMenuSpecTests` cases, and the cascade-promise case in `DatabaseDropRequestTests`.

## Not verified

**`TableProUITests/SchemaManagementMenuUITests` did not execute locally.** The runner fails to initialize with `Timed out while enabling automation mode.` before any test code runs. This is environmental, not this change: a pre-existing suite (`AuxiliaryWindowCloseUITests`) fails identically on this machine, across three attempts and after clearing `testmanagerd`. The suite asserts the negative half of the capability contract on SQLite, where no schema exists; it will run in CI.

The positive UI path needs a live PostgreSQL and does not run deterministically, so it is covered by the unit suites above rather than by automation.

**No screenshots.** The sheet needs a live PostgreSQL connection to render with roles and privileges, and this machine's UI automation is not initializing, so a driven capture was not possible. The two new sheets and the menu items are unphotographed and the `docs/` pages describing them carry no `<Frame>` yet; both need a capture pass on a machine with a working runner.


datlechin:
## UI revamp of the schema sheet

The first cut of the sheet was wrong in a way a diff cannot show, so here is what changed and why.

### The defect behind the dead checkboxes

`SchemaGranteeRow.canEdit` read the **live** `granted` set:

```swift
granted.contains(privilege) ? revocable.contains(privilege) : true
```

Ticking an empty box moved the privilege into `granted`. A grant that does not exist on the server yet has no grantor, so `revocable` could never contain it, so the cell went `.disabled` the instant it was ticked and could never be cleared again. `setPrivilege` guards on the same predicate, which made it a one-way latch.

Editability is a fact about the server's ACL, so it is now read once at load into `SchemaGranteeRow.locked` and never derived from the form. A role added in the sheet locks nothing. Two tests pin it: one toggles a privilege the connected role granted off and on and asserts it stays editable, one adds a role and asserts the same.

### Layout

The old sheet was a fixed 520x620 with three sections that each stretched, so the fields clustered at the top and left two dead zones. Now:

- Three labelled sections (General, Privileges, Statements) in a `ScrollView`, each as tall as its own content.
- The sheet asks for `minHeight: 320, maxHeight: 680` and sizes to what the sections add up to.
- A pinned footer carrying the statement count and the buttons, so they never scroll away.
- The privileges table is sized to its rows, capped at six before it scrolls internally.
- Privileges open on `No roles have access yet. Add one to grant it.` rather than an empty box, since a new schema has no grants.
- Create mode says `Name the schema to see the statements.` instead of the edit-mode `No changes to apply.`

### The privileges control

`SchemaPrivilegeTable` is now a real `Table` with `TableColumnForEach`, so the column headings, the row metrics and the checkboxes are AppKit's. The hand-built `Grid` is gone, and with it the flat non-native checkboxes.

That API is macOS 14.4, so **this raises the app's minimum from macOS 14.0 to 14.4** (`project.yml`, plus `CLAUDE.md`, `README.md` and four docs pages). Flagging it clearly because it is a product decision, not a detail of this sheet.

### Verified on the rebased tree

Build PASS (`logs/build-TablePro-141759.log`), 127 of 127 tests across 7 suites, lint 0 violations, docs PASS.

### Not verified

**No screenshots.** I stood up a throwaway PostgreSQL 17 cluster with roles, grants and a commented schema, launched the Debug build against a sandboxed store, and did get one clean capture confirming the new sheet renders correctly: native blue checkboxes, real `Role / Usage / Create` headings, the three sections, and no dead zones.

A reliable light/dark pair was not possible. Another process on this machine repeatedly launched its own TablePro copy and fired `tablepro://` deeplinks into my instance, which opened foreign connection windows that stole focus and intercepted the clicks driving the sheet. The docs pages therefore still carry no `<Frame>` for these sheets and that remains pending.

One thing that needs a second look when the shots are taken: in that single capture a brand-new schema showed two grantee rows with every box ticked. `granteeRows` is only written by `loadCurrent` (edit mode only) and `addGrantee` (which creates an empty row), so stray synthetic input from the interference is the likely cause, but I could not reproduce it deterministically and am not claiming it is clean.

Issue #2915, opened 2026-09-16
Column completion never appears, and no popup after an empty prefix in FROM (MySQL)


jiangjc3:
### Environment
- TablePro 0.74.0 (build 130)
- macOS 15.7.9 (24G830), Apple Silicon (arm64)
- MySQL connections: a local one (localhost, database `product_center_core`) and a remote one (database `cdfsh_shared`)
- SQL editor, Vim mode on (the bugs below reproduce with it on; see notes)

### Summary
Keywords and table names complete, but column completion after a qualifier never appears — there is no popup at all. Separately, `select * FROM ` with an empty prefix also shows no popup.

### Steps to reproduce
1. Connect to MySQL and open a query tab on the database that contains the table (the database picker at the top of the editor shows that database; the table is visible in the sidebar and queries against it run fine).
2. Type `select * FROM ` (nothing after `FROM`) and wait.
3. Type `select * FROM categor` — `category` is suggested here, so table completion works.
4. Continue the statement to `select * FROM category t where ` and then type `t.`.
5. Also try `select * FROM category where category.` (table name + dot, no alias).

### Expected
- After `FROM ` with an empty prefix, the table list should pop up (the Autocomplete docs show `SELECT * FROM |  -- All tables`).
- After `t.` / `category.`, the columns of `category` should be suggested.

### Actual
- After `FROM ` with an empty prefix: no popup at all.
- After `t.`: no popup at all and no column suggestions.
- After `category.`: same — no popup at all.

Both the alias form and the plain table-name form fail, so this doesn't look like alias parsing. Keywords and table names complete normally throughout.

### Timeline within the session
Column completion **did work once**: with two connection windows open, `select * from spu t where t.` suggested columns. After I toggled Insert/Normal mode a few times it stopped, and it has not worked since — including after a full app restart. After the restart, keywords and table names complete again, but column completion still does not.

### Workarounds tried
- Using a table name picked from the completion popup instead of typing it: no change.
- Confirmed the table exists in the current database and the editor's database picker points at that database: both fine.
- Opening the table's Structure/Columns tab in the sidebar: columns are listed there, so the metadata is readable from the app.

### Additional context (source reading, 0.74.0 tag)
- `SQLCompletionService.isSuppressedEmptyPrefix` explicitly does not suppress an empty prefix for `.from`/`.join`, so the missing popup after `FROM ` looks like a bug rather than intended behavior.
[...]

PR #2923, opened 2026-09-16
fix(editor): match a quoted identifier on the name inside the quotes

datlechin:
Found while investigating #2915. Independent of that fix.

## The bug

Typing an opening backtick killed autocomplete for the rest of the identifier. On MySQL, `SELECT * FROM cat` suggests `category`; `SELECT * FROM ` + a backtick + `cat` suggests nothing, and the popup closes on the backtick itself. The grid filter field shares the path, so it behaves the same way.

## Root cause

`SQLTokenBoundary.segmentStart` counts a backtick and a double quote as token characters, deliberately: the segment is what an accepted completion replaces, so it has to cover the opening quote or the quote would be stranded in front of the inserted name. `SQLTokenBoundaryTests.segmentStartQuotedIdentifier` pins that.

But the same segment was also used as the text to match on, and no candidate's filter text carries a quote. `SQLCompletionProvider.filterByPrefix` tries an anchored match, then a contains match, then a fuzzy match that needs every character of the pattern to appear in the target in order. A backtick appears in none of them, so every candidate was dropped, keyword items included. `CompletionEngine.getCompletions` returns nil for an empty item list, so the request reported no completion context at all.

`extractPrefix` already trims quotes off `dotPrefix` two lines below, which is why `t.` works and `` `t `` does not.

## The fix

The segment and the text it matches on are two different questions, so `SQLTokenBoundary` now answers both: `segmentStart` still covers the quote, and `matchText(of:)` drops it. The analyzer resolves `prefix` through it (and `dotPrefix`, which had the rule inlined), and `SQLCompletionService.rank` does the same for the incremental path, which reads its own prefix off the live token rather than from the analyzer. Both had the quote; both needed the same answer.

The replacement range is untouched, so accepting a suggestion still overwrites the opening quote. A lone backtick after FROM is now an empty prefix, so it lists every table rather than nothing.

## What this does not cover

An unterminated double quote still completes nothing, and that is a different defect: `SQLContextAnalyzer.isInsideString` reads `"` as opening a string literal. That is correct on MySQL and MariaDB, where it is a string delimiter, and wrong on PostgreSQL, SQLite, Oracle and SQL Server, where it quotes identifiers. Fixing it needs the dialect, which the analyzer is not given. `unterminatedDoubleQuoteReadsAsAString` pins the current behaviour so the gap is visible.

No re-quoting on accept. After typing a backtick and accepting `category`, the backtick goes and you get a bare `category`. No completion re-quotes today, with or without this change; `SQLCompletionItem` carries no quoting hint.

## Verified

- `TableProTests`, the 15 autocomplete suites: 549 cases, 549 passed.
- The new cases were run against the unfixed code first: 13 failures across `SQLContextAnalyzerTests`, `SQLCompletionProviderTests`, `CompletionEngineTests` and `QueryCompletionRankingTests`. All pass with the fix.
- `swiftlint --strict` on every changed Swift file: clean.

No screenshots. The change is "a popup opens where it did not", which is what the four suites assert.


datlechin:
Codex review found a regression in the first commit and it is fixed in `025506f94`.

On MySQL and MariaDB `"` opens a string literal, not an identifier. The incremental re-rank path cannot tell the two apart, and trimming the quote turned a lone `"` into an empty prefix, which matches every candidate. That held the popup open inside a string literal where it used to close. `SQLCompletionService.rank` now declines a token that is nothing but quotes and returns an empty list, so the popup closes and the next character goes back through the analyzer, which reads the quote in its own context. `quotesOnlyPrefixRanksNothing` covers it for both quote characters.

The fresh-request path was never affected: `SQLContextAnalyzer.isInsideString` already refuses a cursor inside a string.

The review's other point, that a double-quoted identifier prefix still completes nothing on PostgreSQL and the other standard-SQL engines, is the gap this PR documents rather than closes. Reaching it means teaching the analyzer which quote character the connection's dialect uses for identifiers, which changes string detection, and that is its own change with its own tests. `unterminatedDoubleQuoteReadsAsAString` pins the current behaviour so it stays visible.

380 cases across 7 suites pass.

Issue #2095, opened 2026-08-12
Query authoring for MongoDB and PostgreSQL needs real autocomplete, dialect data, and diagnostics


MinhQuang28:
## Problem

Writing queries in TablePro is painful for MongoDB and PostgreSQL. The editor renders and executes fine, but the authoring layer (autocomplete, dialect data, diagnostics, formatting) is built for generic SQL only, so both of these get a degraded experience.

---

### MongoDB: autocomplete is effectively dead

`SQLCompletionAdapter` is the only completion delegate wired into the editor (`SQLEditorView.swift:67`), and it never branches on `PluginManager.editorLanguage(for:)`. MQL goes through the SQL pipeline unchanged.

The visible result: **typing a dot in MQL produces nothing.** `SQLCompletionProvider.swift:128-149`:

```swift
if let dotPrefix = context.dotPrefix {
    guard let schemaProvider else { return [] }
    if let derived = context.tableReferences.first(where: {
        $0.isDerived && $0.identifier.caseInsensitiveCompare(dotPrefix) == .orderedSame
    }), let columns = derived.derivedColumns, !columns.isEmpty {
        return columns.map { SQLCompletionItem.column($0, dataType: nil, tableName: derived.identifier) }
    }
    if let tableName = await schemaProvider.resolveAlias(dotPrefix, in: context.tableReferences) {
        ...
    }
    if items.isEmpty {
        if await schemaProvider.isKnownSchema(dotPrefix) { ... }
        else if await schemaProvider.isKnownDatabase(dotPrefix) { ... }
    }
    return items          // <- empty for every MQL dot
}
```

For `db.` the dot prefix is `db`. For `db.users.` it is `users`. Neither is a SQL alias, schema, or database, so both fall through to an empty array.

What is left is 23 static snippets in `MongoDBPlugin.swift:110` (`db.`, `.find`, `.aggregate`, `.updateOne`, ...), plus `sqlDialect = nil`. So there is no completion for:

- collection names
- field names inside a filter or update document
- `$` operators (`$match`, `$gte`, `$in`, `$lookup`, `$unwind`, `$set`, `$inc`)
- aggregation pipeline stages, which need a different operator set than a query filter

Trigger characters are `[".", " "]` only (`SQLCompletionAdapter.swift:73-75`), so `$` and `{` never open the popup at all.

`MongoShellParser` also has gaps that surface only after pressing Run:

[...]

PR #2925, opened 2026-09-16
fix(editor): open the type list on a PostgreSQL cast with nothing typed

datlechin:
Found while investigating #2915. Two halves of one mismatch: the docs and the code disagreed about which cursor positions open the popup with nothing typed, and in one of them the code was wrong.

## The code half

`SQLCompletionService.isSuppressedEmptyPrefix` holds the list of clauses that auto-open on an empty prefix. Everything outside it falls to `default: return true` and waits for `Ctrl+Space`. That list was written for 0.21.0, which suppressed noisy empty-prefix suggestions in clauses like WHERE.

0.65.0 (#2095) added `::` cast completion and registered `:` as a trigger character, but never added `.castTarget` to the list. So typing `SELECT id::` fired the trigger, produced the type list, and then threw it away. `SELECT id::js` worked, because a non-empty prefix skips the check entirely.

The type list after `::` is exactly what the list is for: short, closed, and the only thing that can go there. `.castTarget` is now on it.

The other positions the docs claimed stay suppressed, because they are the noise 0.21.0 removed: a bare WHERE, a function's open paren, and a comparison's value side all sit in front of the full column list.

## The docs half

`docs/features/autocomplete.mdx` showed four positions opening on an empty cursor that do not:

- `SELECT * FROM users WHERE |` under Column names
- `SELECT COUNT(|` and `WHERE date_column > |` under Functions and operators
- `WHERE status = |` under Casts and enum values

Each comment now says `Ctrl+Space for …`, and the opening paragraph names the positions that do auto-open rather than describing them as "clauses with a short answer", which was too vague to check. `SELECT id::|` needed no edit: the code change makes it true.

## Still open

`WHERE status = ` on an enum column is the one case where the docs' original promise looks right and the code does not deliver it. `SQLContext.comparisonColumn` is computed for exactly that position and `isSuppressedEmptyPrefix` never reads it. Allow-listing it is not a one-word change: the suppression runs on the clause, not on what came back, so `.where_` would open the whole column list on every comparison rather than the handful of enum labels. Left alone, documented as `Ctrl+Space`.

## Verified

- `TableProTests`, 7 autocomplete suites: 378 cases, 378 passed.
- `castTargetOpensOnAnEmptyPrefix` was run against the unchanged switch first and failed; the other five in the new suite pass either way, which is what makes them a pin on the behaviour this does not change.
- `verify.sh docs`: house style and source claims both agree.
- `swiftlint --strict` on both changed Swift files: clean.


datlechin:
Codex review found two real problems and both are fixed in `d14a0bc71`.

**SQL Server also spells scope resolution `::`.** `SQLContextAnalyzer.endsWithCastOperator` reads any `::` as a cast without consulting the dialect, so unsuppressing `.castTarget` for everyone answered `hierarchyid::` and `geometry::STGeomFromText` with the MSSQL data-type list, selectable with Return. The gate is now the dialect's own declaration: `SQLCompletionService` unsuppresses the empty prefix only when the connection's `SQLDialectDescriptor` declares `::` with category `.cast`, which only PostgreSQL does (`PostgreSQLDialect.swift:295`). `doubleColonWithoutACastOperatorStaysSuppressed` covers the MSSQL shape. The mis-classification itself predates this PR and is unchanged: a typed prefix such as `hierarchyid::Pa` has always reached `.castTarget`.

**`SELECT COUNT(|` does auto-open, and my docs edit said it did not.** `(` is a trigger character and the analyzer classifies that position as `.select`, which an empty prefix does not suppress (`SQLContextAnalyzerTests.testFunctionArgClause` pins it). That line is back to `-- Columns and *`, and the opening paragraph no longer claims function calls are suppressed; it now lists the clauses that open and names WHERE, HAVING, GROUP BY and ORDER BY as the ones that do not.

379 cases across 7 suites pass, and `verify.sh docs` agrees on both checks.

PR #2928, opened 2026-09-16
fix(hig): let every pull-down draw its own disclosure chevron

datlechin:
The query editor's **Run** split button drew two chevrons, and so did the connection form's **Tags** row. Auditing the rest of the app for the same cause found five more menus in the same shape, three of which were drawing a glyph nobody could see.

## Root cause

A SwiftUI `Menu` on macOS is an `NSPopUpButton` carrying its own `NSPopUpIndicatorView`, and a `ControlGroup`'s menu segment carries the same one. That indicator is always drawn unless `.menuIndicator(.hidden)` takes it away.

SwiftUI reduces a macOS menu label to **(first Image = icon) + (first Text = title)**, reorders it icon-first regardless of source order, drops every other subview, and only then appends the platform indicator. So a chevron handed to a menu as its label becomes the control's *icon* and lands beside the real indicator rather than replacing it.

None of this is documented. It was measured with standalone SwiftUI probes compiled with `swiftc` and screenshotted with `screencapture` on macOS 27.0:

| Case | Renders |
| --- | --- |
| `ControlGroup { Button; Menu(_, systemImage: "chevron.down") }` | `▶ Run │ ⌄ ⌄` |
| `ControlGroup { Button; Menu { … } label: { EmptyView() } }` | `▶ Run │ ⌄` |
| `Menu { … } label: { Image("chevron.down") }`, `.menuStyle(.button)` | `⌄ ⌄` |
| `Menu { … } label: { Text("Add tags") }`, `.menuStyle(.button)` | `Add tags ⌄` |
| `.borderlessButton` label `HStack { Image, Text, Image("star.fill") }` | the star is dropped entirely |
| `.borderlessButton` label `HStack { Text, Image("star.fill") }` | `★ gpt ⌄`, with the star as the **leading** icon |

That last pair is why three of these could not be caught by reading the code: a chevron written after an icon and a title renders nothing at all, and the source looks identical either way. `SetPickerView.swift:32` already carries a comment from an author who hit the same measurement and fixed exactly one site.

## What changed

Two that showed two chevrons:

- `QueryEditorBar.swift`: the Run split button's menu half takes an empty label, so the segment's own chevron is the whole control.
- `ConnectionTagEditor.swift`: **Add tags** becomes the pull-down's own title instead of inert text sitting beside it, which also means clicking the words opens the menu. Once a tag is picked the chips say what is selected and the label goes empty.

Three that drew a glyph SwiftUI was discarding:

- `QueryContainerPicker.swift`, and both pickers in `AIChatPanelView.swift`. Measured pixel-identical before and after; this is dead code coming out, not a visual change.

Two that hid the real indicator to draw their own:

- `InspectorFieldRow.swift` and `ValuePickerFieldView.swift`. These rendered correctly (11x7pt, dimming properly on disable), so they were not defects. They are converted because the app should not hand-draw chrome the platform owns; the chevrons are now the system's 7x4pt.

`Menu(primaryAction:)` is the documented macOS split button and renders identically to the fixed `ControlGroup`, but `.disabled()` dims **both** halves. `canOpenRunMenu = canRun || hasQueryText || hasResults`, so a disconnected tab with results on screen needs Run dark and Clear Results live. The `ControlGroup` stays, as its existing comment already said it must.

`.menuIndicator(.visible)` is stated rather than left to the default on the four menus whose only affordance is now that indicator. The floor is macOS 13 and this machine renders 27, so the default is unverifiable where it matters; naming it costs nothing and removes the dependency.

## Not changed

- `.menuIndicator(.hidden)` on icon-only "more" menus (`ellipsis`, `plus`, `gearshape`, `clock`, `ellipsis.circle`). That is the documented idiom, the same as `NSMenuToolbarItem.showsIndicator = NO`.
- `TypePickerFieldView.swift` and `CopyObjectsConfigureView.swift`: plain `Button`s presenting a `.popover`. No platform indicator exists there, so the glyph is the only affordance they have.
- `TableProMobile/ConnectedView.swift:260`: iOS draws no automatic indicator, so removing the chevron there would leave nothing.
- `SafeModeToolbarItem`, `SidebarViewOptionsButton`: native AppKit, already correct.

## Tests

`MenuDisclosureIndicatorTests` scans `TablePro/Views` for a chevron used as a menu label, matching both `Menu(_:systemImage:)` and the `label:` closure form, and scoped to `Menu` so the two `Button`-plus-popover sites stay allowed. Run against the merge base it reports all seven sites; against this branch, none, with 43 menus inspected.

`ConnectionTagEditorUITests` covers the interaction half. It asserts the control's width rather than its label, because a SwiftUI `Menu` is an `NSPopUpButton` whose label no suite here has ever resolved, and the identifier reaches a wrapper carrying no name. Measured, the control is 79pt wide titled **Add tags** and 24pt with an empty label, so width is what says the placeholder is inside the control rather than beside it.

`QueryRunUITests.openRunMenu` reached the menu half as a fraction of the Run half's width. Measured on the merge base, that click **no longer opens the menu at all**: both cases fail with `No matches found for Descendants matching type Menu`. It resolves `query-run-menu` by identifier instead, which is what the sibling inspector suites already do, and `testRunMenuOffersRunAllStatements` passes again.

## Verification

Run through `.claude/skills/fix-issue/scripts/verify.sh` on macOS 27.0 with Xcode-beta, after merging `origin/main`:

| Step | Result |
| --- | --- |
| `generate` | PASS |
| `build` | PASS |
| `test MenuDisclosureIndicatorTests IconOnlyControlLabelTests` | PASS, 2 of 2 |
| `uitest ConnectionTagEditorUITests` | PASS, 2 of 2 |
| `uitest InspectorFieldAffordanceUITests` | PASS, 5 of 5 |
| `uitest ColumnDefaultPickerUITests` | PASS, 2 of 2 |
| `uitest QueryRunUITests` | 1 of 2, up from 0 of 2 on the merge base |
| `swiftlint --strict` over the changed files | 0 violations |

### The one test still red, and why it is not this change

`QueryRunUITests.testRunAllStatementsRunsEveryStatementInTheTab` fails at `chooser.label.contains("2")`: the result-set chooser's label comes back empty after running several statements.

That is a defect on the merge base, not here. `ResultStatementLinkUITests` reaches the same chooser through **Query > Execute All Statements** in the menu bar, with no involvement from the Run control, and fails identically with `Three statements must produce three results, got `. Measured with this branch's six view files checked out from `origin/main`, so the app source was byte-identical to the merge base: 0 of 4 passed, same assertion, same empty label. `ResultSetPinUITests` passes, so the chooser works for a single result; it is the multi-result title that is empty.

Reported separately rather than folded in here, because the cause is in the result-set chooser and this change does not touch it.


datlechin:
## Before / After

Captured from a Debug build driven with `osascript`, launched against a throwaway `TABLEPRO_UI_TEST_SANDBOX`. Same build settings, same crop region, same appearance, the only difference being `QueryEditorBar.swift` at the merge base versus this branch.

| | Run split button's menu half |
| --- | --- |
| Before | `▶ Run │ ⌄ ⌄` |
| After | `▶ Run │ ⌄` |

The two PNGs are on the machine that produced this branch and still need attaching by hand:

- `/private/tmp/claude-501/-Users-ngoquocdat-Workspaces-TablePro/1b99ba8e-44b0-472d-96f2-846571b911a9/scratchpad/shots/before-run-dark-big.png`
- `/private/tmp/claude-501/-Users-ngoquocdat-Workspaces-TablePro/1b99ba8e-44b0-472d-96f2-846571b911a9/scratchpad/shots/after-run-big.png`

The connection form's Tags row has no matching pair. The capture run cannot reach the Appearance tab through `osascript`, whose accessibility tree does not expose the form's sidebar rows, and I stopped rather than spend longer on it. `ConnectionTagEditorUITests` measures that row instead, which is the stronger evidence: the control is 79pt wide carrying the title **Add tags** and 24pt with an empty label, and clicking the words opens the menu. Before this change the words were a sibling `Text` outside the control, and only the chevron was clickable.

datlechin:
## Correction: these menus were losing their name

Pushed as `fix(hig): keep a chevron-less menu's name in its label`.

Giving the four menus an `EmptyView()` label and naming them with `.accessibilityLabel` left them **nameless to VoiceOver**. `.accessibilityLabel` on a `Menu` is not additive: it replaces whatever the label was providing, with nothing.

Measured with System Events reading `AXTitle` off a standalone SwiftUI app on macOS 27.0:

| Construction | AXTitle |
| --- | --- |
| `Menu { … } label: { Text("Add tags") }` | `Add tags` |
| the same, plus `.accessibilityLabel(Text("Add tags"))` | *(empty)* |
| `Menu { … } label: { EmptyView() }.accessibilityLabel("Value Options")` | *(empty)* |
| `label: { Label { Text(name) } icon: { EmptyView() } }` + `.labelStyle(.iconOnly)` | `name` |
| `label: { Image("ellipsis").accessibilityLabel(name) }` | `name` |

The last two are the shapes that work: the name has to be inside the label closure. An icon-less `Label` rendered `.iconOnly` draws nothing but the control's own disclosure chevron and still publishes a name, verified in all four shipped contexts plus inside the `ControlGroup`.

`InspectorFieldRow` is the one this mattered most for. Its original `Label("Value Options", systemImage: "chevron.down")` was correctly named; the first version of this branch replaced it with `EmptyView()` and silenced it. It is named again.

### After the correction

`uitest ConnectionTagEditorUITests InspectorFieldAffordanceUITests ColumnDefaultPickerUITests QueryRunUITests`: 20 of 22, the two failures being the one case described in the description.

### The remaining red case now has a named cause

`QueryRunUITests.testRunAllStatementsRunsEveryStatementInTheTab` asserts on `result-set-menu`'s label, and `ResultSetMenu.swift:26` carries `.accessibilityLabel(model.title)` on its `Menu`. By the measurement above, that is exactly what makes the name empty, so the chooser has never published one and the assertion could not have passed. `ResultStatementLinkUITests` fails the same way through the menu bar.

A scan of `TablePro/Views` finds 12 menus with the name in that position, `ResultSetMenu` among them. Fixing them is a separate change, since the cause is accessibility naming rather than the disclosure chevron this PR is about, and it will turn both suites green.

datlechin:
The one red case in the description is closed by #2929, which is stacked on this branch. It was neither a multi-result bug nor a lost name: the chooser publishes its name in `AXTitle` and both suites asserted on `label`, which is `AXDescription`. `QueryRunUITests` is 2 of 2 there, and `ResultStatementLinkUITests` 4 of 4.

Issue #483, opened 2026-03-28
Oracle connection fails with uncleanShutdown on non-TLS server


wiscaksono:
## What happened

Tried connecting to an Oracle database (port 1521, plain TCP, no TLS) and got this:

```
Failed to connect to <redacted>:1521/<redacted>: OracleSQLError(code: uncleanShutdown) - Some information has been reducted to prevent accidental leakage of sensitive data. For additional debugging details, use `String(reflecting: error)`.
```

The same server works fine from SQL Developer and DBeaver with identical credentials. So the database itself is reachable and accepting connections -- just not from TablePro.

## What I think is going on

Based on oracle-nio issue lovetodream/oracle-nio#21, the underlying error is probably `NIOSSL.NIOSSLError.uncleanShutdown`. My Oracle server runs on standard port 1521 with plain TCP (no TLS/TCPS). It looks like the Oracle driver is trying to negotiate TLS, the server doesn't speak TLS, and it just drops the connection.

There's also a firewall between my machine and the server, but that shouldn't matter since other clients get through without issues.

## What I expected

A successful connection, or at least a clearer error message (e.g. "TLS handshake failed, try disabling SSL" or something along those lines).

## Environment

- macOS (Apple Silicon)
- TablePro: latest version from GitHub Releases
- Oracle server: port 1521, plain TCP, no TLS requirement

## Suggestion

It would help to have either:
- An option in the connection form to disable TLS for Oracle connections
- Auto-detection of whether the server supports TLS before attempting the handshake

Toni-99:
Can confirm, that Oracle Thin Drivers are not supported.

zetkey:
had same issue with:
- MBP M3 Pro
- TablePro: latest version from GitHub Releases
- Oracle server: port 1521, plain TCP, no TLS requirement
- Oracle version 23ai (Enterprise Edition)
- Connect using service name

if allowed to compare, my config runs well on dbeaver

datlechin:
@wiscaksono @zetkey try to update both TablePro app and Oracle in Settings then try connect to the Oracle connection. Hope it would fixes the issue.

wiscaksono:
> [@wiscaksono](https://github.com/wiscaksono) [@zetkey](https://github.com/zetkey) try to update both TablePro app and Oracle in Settings then try connect to the Oracle connection. Hope it would fixes the issue.

[...]

PR #2935, opened 2026-09-17
fix(plugin-oracle): bound the login handshake and read the accept connect flags correctly

### LINKED ISSUE #2919: In the case of vpn, it is not possible to connect to the remote database
### What happened?

ImageAfter using uniconnect to connect to the office vpn, the database cannot be connected in tablepro, but it is possible to use DBeaver

Image

Image

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Database type

None

### TablePro version

0.74.0

### macOS version & chip

macos15.7.9/Apple M2 Pro

### Screenshots / Logs

_No response_

datlechin:
Fixes #2919.

Oracle 11.2.0.4 reached over a corporate VPN. TablePro reports **Login Handshake Timed Out**; DBeaver with Oracle JDBC thin 23.2 connects in 1000 ms over the same VPN, same host, same credentials.

Two defects, both measured. The VPN is shared context, not the differentiator.

## 1. The driver could not hear a server declining native network encryption

`Accept.decode` read the accept packet's connect flags at `readerIndex + 18/+19`. The decoder hands it the whole packet with the reader moved past the 8-byte header, so after the version and options the reader sits at 12 and that read landed on packet bytes **30/31**, the reconnect-address length, instead of **22/23** where ACFL0 and ACFL1 live.

Confirmed against two independent implementations:

- go-ora `v2/network/accept_packet.go`: `ACFL0 = packetData[22]`, `ACFL1 = packetData[23]`, and its own writer puts them there. Its gate at `v2/connection.go:505` is `ACFL0&1 != 0 && ACFL0&4 == 0 && ACFL1&8 == 0`.
- python-oracledb thin reads the same byte, as body offset 14 past an 8-byte header.

Measured against a hand-rolled TNS listener speaking protocol version 314, driving the real driver: an accept carrying `0x04` (`DISABLE_NA`) at byte 22 still got the 156-byte `0xDEADBEEF` negotiation packet; the same byte moved to 30 made the driver skip it. So the negotiation was decided by an unrelated length field.

The fork also dropped go-ora's first condition, `ACFL0 & 0x01`, with a comment that some servers do not echo the support bit. That observation was an artifact of the offset: with the wrong bytes every server reads `0x00`. Restoring the condition turns two tests green that are red at the pinned revision today (`authHappyPath` and `fastAuthHappyPath` both trapped with `Invalid state: advancedNegotiationSent`).

The accept flags are now read at their own absolute offsets, an accept shorter than the 32 bytes every server sends is rejected instead of walking the reader off the end, and the gate matches go-ora.

## 2. Nothing bounded the login, so a stalled handshake ran until the server gave up

`OracleCoreConnection` wrapped connect in `withOracleTimeout` with no `onTimeout`. The task group cancels the child and then awaits it, and the child is parked in `EventLoopFuture.get()`, which ignores cancellation. oracle-nio had no handshake timer of its own; `connectTimeout` bounds the TCP connect only.

Measured: a 2 s deadline over a 6 s uncancellable call surfaced at 6.40 s, and an accept-then-silence hung 20 s and unblocked only when the listener closed. The dialog therefore appeared whenever the server gave up, not at 30 s, and never if it did not.

This is the `Cancelling a connect does not stop the driver` invariant in CLAUDE.md, unenforced.

- The driver now schedules a login deadline on the event loop and fails the startup promise with the handshake phase attached, so the dialog names the step the server stopped answering at.
- `OracleConnection.start` no longer chains `channel.closeFuture` behind a startup error, which held the caller until the server hung up. Measured: an advanced-negotiation error now throws in 0.009 s where it took 20.009 s.
- The app keeps a backstop 5 s behind the driver, and a connect that lands after it closes its own handle rather than leaking a socket per abandoned attempt.

## Also in scope, same phase, all measured

- A marker packet during the handshake was a `preconditionFailure`. A release build against a fake listener exited 133. The plugin is loaded in-process, so that was a pre-auth remote kill of the app, the same class as #1746.
- An accept under 32 bytes trapped in `moveReaderIndex(forwardBy: 20)`, live in `-O`.
- Repeated resend packets were re-sent without a counter: five identical negotiation packets in two milliseconds against a listener that answered nothing else.
- A negotiated encryption or checksum algorithm with no Diffie-Hellman material logged `Native network encryption negotiated` and then sent the rest of the login in clear text. A real server cannot produce that, since the key exchange is the only key source, but a silent downgrade to plaintext is the wrong failure, so it now fails the connect.

## Network Encryption on the connection form

New dropdown mapping 1:1 to `SQLNET.ENCRYPTION_CLIENT`: Rejected, Accepted (default, what Oracle's own clients send), Requested, Required. Nothing set `Configuration.nativeNetworkEncryption` before this, so the diagnosis had no lever on the one server that matters; now a reporter can isolate the negotiation.

New error codes are wired through `OracleConnectErrorClassifier` so they reach a named dialog instead of falling to the generic `default:` arm.

## Not done, on evidence

- Reading the legacy SDU for protocol versions below 315. Real, and go-ora reads and clamps it, but it is independent of this bug and feeding an unfloored value into the frontend post-processor is worse than ignoring it: measured, `maxSize = 10` spins at 99.7% CPU and 5.66 GB RSS and never returns, `maxSize = 0` trips a precondition, `maxSize = 8` force-unwraps nil. Reported separately.
- Changing the offered algorithm lists. The data-integrity list `[0,1,3,4,5,6]` is byte-identical to go-ora's, so "we offer SHA-2 checksum IDs an 11.2 server cannot do" is not a divergence from the reference client.
- The Diffie-Hellman, AES-CBC and checksum wire format. Byte-identical to go-ora, and still unverified against a real encrypting server. This change removes the unconditional entry into that path; it does not claim the path is right.

## Shipping

Oracle is registry-only. The fix rides the next app release and the next `plugin-oracle-v*` build; it is not being back-published to 0.74.0.

Driver change: TableProApp/oracle-nio `fix/2919-bound-login-handshake`, commit `42f8518`. Re-pinned in all six tracked places (`licenses.yml` carries the revision twice).

## From the Codex review

Seven findings, all addressed. Four existed only because of the new field, which turned out to carry more weight than it looked:

- `Required` was a guarantee in name only: it changed the offered algorithm list and nothing else. The driver now refuses a server that will not negotiate, and refuses a negotiation that comes back having selected no cipher.
- The login-timeout close was swallowed. The cleanup had already moved to `readyToLogOff`, so the close waited on a logoff reply the failing server never sends. A failure before authentication now hard-closes through a new `closeImmediately` cleanup action, since there is no session to log off from.
- `advancedNegotiationRequired` hit the generic encryption predicate first and surfaced the wrong remediation. It is routed to its own error now.
- The resend budget was per connection rather than per packet, so eleven isolated resends across a long session would have closed a healthy one. It resets on any non-resend message.
- The field also sets what Oracle configures separately as `CRYPTO_CHECKSUM_CLIENT`, because the driver negotiates both in one exchange. Said so in the option, the settings table and the docs rather than implying otherwise.
- The iOS driver read the field but the iOS form had no control for it, so an iPad-only user could not reach the workaround. Added the picker, the view-model property, the load and save path and three tests.
- The app-side backstop cannot unwind a connect that never returns at all, because the task group awaits its child. The driver's own deadline is the real bound; the backstop now only guards the late-handle leak, and says so.

## Verification

- oracle-nio: the four affected suites, 29 tests, all passing.
- `TableProOracleCoreTests`: 80 tests, no failures, including new coverage for the option, its wire strings, the level mapping and the phase naming.
- `verify.sh build OracleDriver`: PASS against the new pin.
- `verify.sh lint` over every changed Swift file: 0 violations. `verify.sh generate`: PASS, and the re-pin survives regeneration.
- No PluginKit change, so no ABI bump.

One gap worth stating: `Packages/TableProOracle` does not build on Xcode 27 beta at either pin, old or new, because oracle-nio's `@TaskLocal` macro emits `usableFromInlinenonisolated` under Swift 6.4. That is pre-existing and unrelated to this change; the runs above used a locally patched build checkout, and CI builds on its own toolchain.

Two notes for review that look alarming and are not. `Capabilities.encode` / `init(from:)` also serialize the accept flags, but only into a client-internal REF CURSOR blob that never reaches the server. And the pre-fix behaviour on modern servers was not "the negotiation always ran": bytes 30/31 are a length field, so it was already being skipped whenever that length had `0x08` set.


datlechin:
## Second commit: the five collateral findings

The investigation for this issue produced a register of five verified defects in the same subsystem. All five are now fixed here rather than left for later. Driver revision re-pinned `42f8518` -> `a786030`.

**1. An 11.x server's negotiated SDU was ignored.** Below protocol 315 the session data unit lives only in the 16-bit pair at packet bytes 12 and 14, and the driver kept its own proposal of 8192, so it sent packets larger than the server agreed to whenever a DBA had tuned it down. `go-ora` reads both and clamps the session unit to the transport unit; so do we, with a floor at Oracle's own minimum of 512 because a smaller value is not a slow connection but a broken one: measured against the frontend post-processor, `maxSize = 10` spins at 99.7% CPU and 5.66 GB resident without returning, 8 force-unwraps nil and 0 trips a precondition.

The new test caught a bug in the first version of this fix. NIO's `getInteger(at:)` only sees bytes that are still *readable*, and I had read the SDU after advancing the reader past it, so it silently answered nil. Every fixed-position field in that decoder is now read together, before the reader moves.

**2. `isChannelFatal` listed 3 of the 12 codes the driver closes the channel for.** `uncleanShutdown`, which is every mid-statement drop, fell through, so the connection stayed marked alive over a channel the driver had already torn down and the next query waited on a socket nobody was reading. It now mirrors the driver's set, plus server ORA-28 and ORA-600 through a new `serverErrorNumber` parameter.

**3. A silent redial was discarding session state.** `reconnectedConnection()` replayed only the session schema. Everything else died with the socket: the connection's Startup Commands, any `ALTER SESSION` the user ran, the rows in a global temporary table, `DBMS_LOCK` allocations, package state and any open transaction. Worse, oracle-nio sets the session time zone at login from this Mac, so a redial did not merely lose a user's `ALTER SESSION SET TIME_ZONE = '+07:00'`, it replaced it.

The load-bearing part was that `ping()` went through `executeQuery`, so the health check healed the dropped socket itself. The monitor reported healthy, the app's reconnect never ran, and `applyTimeoutAndStartupCommands`, which is the one path that puts startup commands back, was never reached.

Fixed in MySQL's shape rather than inventing a local answer: a new `OracleSessionFootprint` mirroring `MySQLSessionFootprint` reads what the session is holding from the statements that go through it, and a rebuild guard refuses a silent redial with a reason in the user's words. `ping()` no longer rebuilds at all. Every ambiguity resolves toward "the session is holding something", and `OracleSessionFootprintTests` asserts the gate and the message agree for every flag, which is what stops them drifting apart.

**4. A shipped build could not produce a packet capture.** `OracleTraceHandler` was `#if DEBUG` only, which is why diagnosing this issue took a hand-rolled mock server, and why #483, #1746 and #1786 were the same.

This is not an env-var flip: a TNS login capture is an offline password oracle, and a token login is a live bearer token in the clear. The tracer is now always in the pipeline with `ORANIO_TRACE_PACKETS` deciding, and a latch set at the one site the authentication messages are emitted stops it dumping bodies from that point on. Past the boundary it logs the direction, packet type, flags, declared length and body byte count, which is what diagnoses a stalled handshake, and no payload byte ever. The latch is set where the messages are emitted rather than by enumerating connection states, because a state added later would not know to exclude itself. A debug build can still ask for the credentials with `=2`; a shipped binary cannot.

**5. A listener redirect was never followed.** `followingRedirect` had no production caller, so a shared server dispatcher or a RAC SCAN listener surfaced as a raw struct dump. The driver now follows the chain, up to five hops.

Three things the obvious version gets wrong, all avoided: the follow-up connect sends the descriptor the listener handed back rather than one rebuilt from the configuration, because python-oracledb does and rebuilding it makes shared server a coin flip; a `PROTOCOL=tcps` redirect on a plain-TCP configuration is refused rather than followed, since following it would connect in clear text to an endpoint that asked for TLS; and the login deadline bounds the whole chain rather than each hop, so five hops cannot outlive a caller that allowed one login timeout. The existing same-address refusal is gone: it was never a loop bound, since two servers pointing at each other never repeat consecutively, and it refused a legal same-address redirect carrying connect data.

## Verification

- oracle-nio: 45 tests in 7 suites, no failures.
- `TableProOracleCoreTests`: 99 tests, no failures, up from 80.
- `verify.sh build OracleDriver`: PASS. `verify.sh lint`: 0 violations. `verify.sh docs`: PASS.

The exhaustive switch over `OracleCoreError` caught the one mistake in this batch: a new case with no diagnostic arm, which is exactly what that pattern is there for.

## What is still not proven

Nothing here was run against a real Oracle server. The redirect mechanics, the SDU clamp and the footprint are all covered offline, but the redirect chain in particular is only as safe as sending the server's own connect string makes it, and a real shared-server dispatcher has not accepted one of these follow-ups.

One question this opened and did not close: oracle-nio's `autoCommit` defaults false and nothing in the app's write path calls `commitTransaction`. If Oracle writes really are left uncommitted, the silent redial was discarding saved rows rather than only settings, which is a larger bug than this one and needs measuring against a server before anyone acts on it.

PR #2937, opened 2026-09-17
feat(sidebar): group partitions under their parent table

### LINKED ISSUE #2523: Group partitions under their parent table in the sidebar
### Problem

A partitioned table puts every partition in the sidebar as its own row. A table partitioned by month over a few years buries everything else in the list, and there is no way to tell a partition from a real table.

### Proposed solution

- Nest partitions under their parent as children, collapsed by default.
- A distinct icon, and the partition bound shown beside the name.
- Count the partitions on the parent row rather than listing them all when collapsed.
- A setting to hide partitions entirely for anyone who never wants to see them.
- The same treatment on every engine that reports a parent relationship: PostgreSQL, MySQL, Oracle, ClickHouse.

### Alternatives considered

The sidebar filter, typed anew every time.

### Related database type

PostgreSQL

datlechin:
Fixes #2523.

## What was missing

PR #1989 shipped the first of the issue's five bullets, for PostgreSQL only: a partitioned table is listed once and its partitions nest under it. The other four needed data the driver contract could not carry. `PluginTableInfo` has no bound and no count, `fetchPartitions` returns only a name and a type, and a partition child was modelled as a `TableInfo`, whose identity is `schema.name_type`. That identity is correct only where a partition is a relation. On MySQL and Oracle a partition is metadata on one table, and its name is unique only within that table, so every partitioned table's `p0` would collide with every other table's `p0` and with a real table called `p0`. That is why those engines were never wired at all.

## The contract

`PluginPartitionInfo` is the new transfer type: name, its own schema, a display-ready bound, ordinal position, row estimate, and `relationType`, which is what tells a PostgreSQL partition from a MySQL one. The kind travels rather than a bare is-it-a-relation flag, because PostgreSQL 11 allows a foreign table as a partition and a foreign table is read-only: calling one an ordinary table offers Truncate on another server's data. `PluginDriverAdapter` now has one shared plugin-type vocabulary, so a table and a partition cannot classify the same string differently. `fetchPartitionDetails(table:schema:)` is a new defaulted requirement whose default maps the existing `fetchPartitions`, so a plugin built before this exists keeps answering. `fetchPartitions` itself is untouched. `PluginTableInfo` gains `partitionCount` through a third initializer overload, with both existing signatures kept verbatim.

Both changes are additive, and both reuse kit 31, so no version field moves. v0.74.0 shipped kit 30; 6ab9d887b raised it to 31 earlier this cycle and all 40 plugin `Info.plist` files already carry 31, and no release has shipped that number. CLAUDE.md's rule is that the first ABI change after a release bumps the version and every later change in the same cycle reuses the pending one, because only the value at release time ever ships. Bumping again here would close the registry's three-kit retention window for no gain, which is the failure that left six releases unable to install any registry plugin. Measured before writing it: the three-initializer ladder compiles under `-swift-version 6`, `PluginTableInfo(name:type:comment:)` still resolves to the five-parameter initializer, and JSON saved before the field existed decodes with `partitionCount` nil.

## Per engine

**PostgreSQL** projects `pg_get_expr(relpartbound, oid)` and the child's own `nspname`, and folds the count into the listing query it already runs. Measured on PostgreSQL 17.11 against a schema of 1200 partitions and 300 plain tables: 12ms without the count subquery, 9 to 13ms with it. The bound arrives as `FOR VALUES FROM ('2024-01-01') TO ('2024-02-01')`; every row under one parent repeats that prefix, so the row shows `FROM ('2024-01-01') TO ('2024-02-01')` and leaves a bare `DEFAULT` alone.

**MySQL and MariaDB** join a grouped `information_schema.PARTITIONS` into the table listing, which supplies both the count and the `PARTITIONED TABLE` typing in one round trip. Measured on MariaDB 12.3.3: a table that is not partitioned still has one all-null row in that view, so the filter is `PARTITION_NAME IS NOT NULL`; a subpartitioned table repeats its partition name once per subpartition, so the count is `COUNT(DISTINCT PARTITION_NAME)`. `PARTITION_DESCRIPTION` holds the bare upper bound for `RANGE` and the bare values for `LIST`, and nothing at all for `HASH` and `KEY`, so the row puts the syntax back. `RANGE COLUMNS` and `LIST COLUMNS` are matched by prefix, because an equality test on the method left both without a bound. Databend answers the same wire protocol through the same driver without this catalog, so it keeps the plain listing, matching the eight sibling methods that already guard on it.

**Oracle** joins `ALL_PART_TABLES` and reads `ALL_TAB_PARTITIONS`. It ships no bound: `HIGH_VALUE` is a `LONG` column, the datatype this driver already avoids because OracleNIO cannot decode it, so the row carries the position instead. Being partitioned is projected separately from the count, because an interval-partitioned table is partitioned and has no usable count: `ALL_PART_TABLES.PARTITION_COUNT` there means the range the server will extend into. Reading the missing count as "not partitioned" would have taken those tables' partitions out of the tree entirely.

The shaping of that list lives in `OraclePartitionMapping`, which takes plain values and is compiled into the test target. Everything that imports `TableProOracleCore` pulls in OracleNIO, whose `@TaskLocal` macro does not expand under Xcode 27, so on a developer Mac the Oracle plugin, the Oracle package and its tests cannot be built at all. The workflow already documents this. Keeping the list's shape out of the driver is what lets it be tested here; the driver extension is left thin enough to read in one pass.

**ClickHouse is out of scope.** Its partitions are data parts that churn on every merge and already have a first-class home in the Structure tab's Parts view, with Optimize, Drop Partition and Detach Partition on it. No client surfaces them in an object tree.

## In the sidebar

Every partition is a `DatabaseTreeNode.Kind.partition` whose `tableRef` is non-nil exactly when the partition is a relation. PostgreSQL partitions go through it and keep opening, dropping, truncating, renaming and favouriting on the paths they already used. MySQL and Oracle partition rows offer Copy Name, Copy Bound and Copy Table Name and nothing that writes.

The bound sits in the caption slot beside the name, the one `TableRow` already uses for an object comment, truncating from the middle so both ends of a range survive. The count sits in the same shape `redisRow(.namespace)` already uses for a Redis key count. Not SwiftUI's `.badge(_:)`: it is documented for List rows, tab bars, toolbar items and menus, and a per-cell `NSHostingView` inside an `NSOutlineView` is none of those, so it compiles and renders nothing. AppKit has no outline-row badge at all, and the HIG says nothing about child counts on a disclosure row.

**Show partitions** is a new General setting, on by default, in the sidebar's View Options and in Settings > Sidebar. It travels the coordinator's `update(from:)` prop diff rather than `observeObjectListAppearance`, which only reconfigures rows and would have left the partitions on screen. Turning it off does not clear the persisted expansion set.

## Three defects this found, all reproduced on a live PostgreSQL 17.11

**A partition in another schema pointed every statement at a different relation.** `fetchPartitions` never projected the child's namespace and the driver stamped the parent's schema on every row, which `qualifyingSchema` then fed to open, drop, truncate, copy and MCP. Reproduced with `archive.orders_2023 PARTITION OF public.orders` beside an unrelated `public.orders_2023`: Delete on the partition row emitted `DROP TABLE "public"."orders_2023"` and dropped the unrelated table while the partition survived.

**A partition was unreachable when its parent was not readable.** The exclusion keyed on the child alone, so it dropped the row whether or not the parent ever reached the listing. Reproduced with a role holding `SELECT` on `public.m_2024` and nothing on `public.m`: the schema listed zero tables while `SELECT * FROM public.m_2024` worked. The exclusion now also requires the parent to appear in `information_schema.tables`, reusing that view rather than restating its privilege predicate so the two cannot drift. Measured: the ordinary case is byte-identical and the 1500-table schema returns the same 301 rows in the same time.

**A foreign-table partition was listed twice.** PostgreSQL 11 allows a foreign table as a partition, and the foreign-table union arm carried no partition filter. Reproduced through `postgres_fdw`: `ft.child_2024` appeared both under Foreign Tables and nested under `ft.parent`.

## Tests

New: `PostgreSQLPartitionBoundTests`, `MySQLPartitionBoundTests`, `MySQLPartitionQueryTests`, `SidebarPartitionRowTests`, `GeneralSettingsPartitionsTests`, and partition cases in `DatabaseTreeNodeTests` and `PluginDriverAdapterPartitionDefaultTests` covering the legacy-plugin bridge. Extended: `PostgreSQLPartitionFilterTests` for the parent-visibility predicate, the foreign-table arm, the projected namespace and the count column; `DatabaseTreeMenuSpecTests` and `SidebarMenuBuilderTests` for the new view option.

Every bound string asserted in those suites is what the server actually returned, not a transcription: PostgreSQL 17.11 for `FROM (…) TO (…)`, `IN (…)`, `WITH (modulus 4, remainder 0)` and `DEFAULT`, and MariaDB 12.3.3 for `2024`, `MAXVALUE`, `1,2,3`, `10,10` and `'de','fr','nl'`.

No `TableProUITests` case for the toggle: driving it needs a live partitioned server, and the sidebar's drawn rows take no synthetic click, so the flow does not run deterministically on a runner.

## PluginKit ABI

`scripts/check-pluginkit-abi.sh` against the merge base reports one new defaulted requirement, one new struct, one new field with a new init overload, and the existing five-parameter init gaining `@_disfavoredOverload`. That last line is the only apparent removal, and it is not one: measured with `nm` over two library-evolution builds of the same initializer with and without the attribute, the emitted symbols are identical, so no shipped plugin loses a symbol it references.

## What was verified, and what was not

Built, `AllPlugins` compiled, the suites above run, and `swiftlint --strict` clean over every changed file. Both docs checks pass.

**Oracle needs a plugin release, not just the app release.** Oracle is registry-only, so an app release alone delivers none of its partition support: the newest published `plugin-oracle` predates these methods and reports Oracle tables as ordinary tables, and the registry readiness gate only checks that some compatible binary exists. `plugin-oracle` has to be built from this commit and published at kit 31 before or with the app release.

Not compiled locally: `OraclePlugin.swift`, `OraclePluginDriver+Partitions.swift` and `OracleSchemaQueries.swift`, along with the new `OracleSchemaQueriesTests` cases. Xcode 27 cannot expand oracle-nio's `@TaskLocal` macro, so `AllPlugins`, the `OracleDriver` scheme and `swift build` on the package all fail on the dependency before reaching any of this. CI runs Xcode 26.4.1, where `Compile every plugin` and `Run TableProOracle package tests` both cover it. CI is the first compile of those three files.

## Out of scope, and said so rather than left quiet

A partition is still not reachable through the sidebar filter. That regressed when #1989 stopped listing partitions flat, it predates this branch, and filtering only the partitions that happen to be expanded would be a silent failure mode rather than a fix. Table columns are not searchable either. Making partitions searchable needs a per-schema name fetch while the filter is active, which is its own change.


datlechin:
## After

Sidebar with a partitioned table expanded, its partition count beside its name and each partition labelled with its bound

A live PostgreSQL 17.11. `events` carries its partition count, and its partitions show their bounds with `DEFAULT` sorted last. `m` and `orders` show their counts without being expanded.

There is no before shot for the bound or the count, because neither existed: 0.74.0 nests PostgreSQL partitions under their parent with bare names and nothing on the parent row.

sidebar-partitions

PR #2962, opened 2026-09-17
fix(plugins): read MySQL catalogs the server can answer, and classify what it lists

datlechin:
PR #2950 fixed the MySQL table list and reported four defects it had found but not fixed. This is those, plus what the investigation into them turned up, in one PR.

## What was wrong

**1. The export dialog read its own catalog.** `ExportDriverMetadataReader.fetchTablesGroupedByDatabase` sent one hand-written `information_schema.TABLES` query for every database, whatever the engine. `.byDatabase` is the grouping every plugin gets unless it overrides it, so that MySQL query also went to Kafka, Cassandra and Teradata, where it throws and takes the whole dialog with it. Measured through three real proxies in front of MySQL 8.4.11: DBLE 3.23 answers every `information_schema` read with zero rows, MyCat 1.6.7.5 refuses it with 3000 or 1064, and ShardingSphere-Proxy 5.5.3 answers one backend's physical catalog, whose schema names are not the logical ones.

**2. MySQL table kinds were three.** Everything that was not a view became `TABLE` or `PARTITIONED TABLE`, so an OceanBase external table offered row editing and Truncate, a MariaDB sequence offered Truncate (server: 1031), and a MariaDB system-versioned table offered Truncate (server: 4137). MariaDB also lists a session's temporary tables, so a temporary table shadowing a base table drew the same name twice.

**3. `information_schema` blindness went past the table list.** Behind DBLE the column, index and foreign-key reads answer zero rows with no error. A structure comparison read that as "this table has no indexes and no foreign keys" and generated `DROP INDEX` and `DROP FOREIGN KEY` for every one of them; the ER diagram came up empty.

**4. The `BINARY` operator.** MySQL 8.0.27 deprecated it, and every table-list read raised three of Warning 1287 from 8.4 on.

**5. A `SELECT` with no result set was answered with a guess.** The driver scraped a name out of the statement and ran `DESCRIBE` on it, so `SELECT COUNT(*) INTO @c FROM users` came back carrying the columns of `users`, and behind ShardingSphere a declined catalog read came back as `1146 Table 'db1.information_schema' doesn't exist`.

Plus, found on the way: partitioned tables were missing from Backup Dump and the import picker (live since #2937 shipped in 0.75.0), ClickHouse answered every database's table list from `currentDatabase()`, and iOS listed system views as tables with Truncate and Drop on them.

## What changed

- **Export**: the server-wide read is an optimization that no longer has to succeed. A database it did not describe is read through that engine's own `fetchTables(schema:)`, one that cannot be read becomes an empty group instead of failing the dialog, object kinds come from the driver, and a dismissed dialog no longer raises an alert. ClickHouse's table list names the database it was asked about.
- **Kinds**: a `SEQUENCE` kind, external and system tables passed through, `SYSTEM VERSIONED` carried as a trait on `TableInfo` so only Truncate is withheld, session temporary rows dropped, and the same vocabulary on iOS. Backup Dump and the import picker take partitioned tables, and a PostgreSQL partitioned parent expands to its partitions so a narrowed dump carries the rows.
- **Proxy-safe metadata**: one catalog-visibility verdict per database, learned from reads the driver already makes. Where the catalog is blind, columns, indexes and foreign keys come from `SHOW FULL COLUMNS`, `SHOW INDEX` and `SHOW CREATE TABLE`, all-or-nothing so a partial answer can never read as "none". The foreign-key read is two unjoined catalog statements merged in Swift, which also fixes composite key ordering on MariaDB. A healthy server sends exactly the statements it sent before.
- **Compare**: a per-table index or foreign-key read that throws now fails that table, and a table whose read failed is reported by that alone. It used to lose its snapshot, turn up as "only in target", and offer `DROP TABLE`.
- **`CAST(... AS BINARY)`** instead of the deprecated operator, and the `DESCRIBE` guess is gone.

## Verification

- Build and the touched suites pass (368 cases), lint clean except violations that predate the branch in files this change touches, docs checks pass.
- Measured on MySQL 5.5, 5.6, 5.7, 8.0.11, 8.0.16, 8.4, 9.7, MariaDB 5.5, 10.2, 10.6, 11.4, 12.3 and TiDB 7.5/8.5, through DBLE 3.23, ShardingSphere-Proxy 5.5.3 and MyCat 1.6.7.5, with the real driver compiled against the real client library.
- `CAST(... AS BINARY)`: identical rows, no warnings, on all eleven MySQL and MariaDB servers and TiDB. Every `COLLATE` spelling fails with 1253 on at least one of them.
- Foreign keys parsed from `SHOW CREATE TABLE` match the catalog field for field on 8.4.11 and 11.4.13, including a composite and a cross-database key. MySQL 8.0.11 and 8.0.12 print neither `RESTRICT` nor `NO ACTION`, so that one field cannot be recovered there; 8.0.13 and later print it, and the limitation is in the docs.
- OceanBase was not run: the Docker VM had 2.2 GB free of the 6 GB it needs. Its type spellings come from its own source.

A four-lens adversarial review of the diff found 12 defects, all fixed here. Besides the `DROP TABLE` above:

- One refused catalog read used to pin a database to the `SHOW` path for the whole session, taking its comments, partition counts and generation expressions with it. Measured: on a 2,402-table schema with a 1 ms statement cap, that read is refused 10 times out of 10, on a healthy server. A refusal now teaches nothing; only an answer does.
- `fetchForeignKeys` looked its answer up by the caller's spelling of the table and the catalog keys it by the server's, so on `lower_case_table_names=1` a query on `Orders` came back with no foreign keys at all.
- ClickHouse's database literal escaped the quote but not the backslash. Measured on 24.8.14.39: a database named `q\' OR 1=1 -- ` listed every table on the server.
- A MariaDB sequence reopened from Recent Tables came back as a view, so the menu offered Drop View.
- The export tree classified `information_schema` objects as system tables, so an SQL export wrote `CREATE TABLE` for a view.
- iOS typed an external table as a system table, which is row-editable there.
- A failed partition read let a narrowed PostgreSQL dump write an empty table and report success; that database is now reported as failed.


datlechin:
No before/after screenshots in this PR. The visible changes are a MariaDB sequence's row (number icon, **Sequence** kind, **Drop Sequence** in its menu), partitioned tables appearing in the Backup Dump and import pickers, and the export tree's grouping. Capturing them needs a live MariaDB connection driven with synthetic input, and while the fixture was up another session's UI test run raised a "XCTest is trying to Enable UI Automation" password prompt on this machine, which blocks that. Everything here is covered by unit tests instead, including one per raw type the listing can emit.

PR #3007, opened 2026-09-20
test(plugin-oracle): expect the qualified dictionary view the trigger list now reads

datlechin:
## Summary

`main` is red. #2999 qualified every Oracle dictionary view at SQL level, so `OracleObjectQueries.triggerList` now reads `FROM SYS.ALL_TRIGGERS`, but `triggerListSelectsActionType` still expects the unqualified `FROM ALL_TRIGGERS` and fails on every run.

The source is right and the assertion is stale, so the assertion moves to the qualified name. The rest of the case is unchanged: it still pins `ACTION_TYPE` being selected and `TRIGGER_BODY` staying last, which is the reason the test exists (`TRIGGER_BODY` is a LONG, so a column after it cannot be read).

## Verification

`OracleObjectQueryTests`: 17 executed, 17 passed.

Filtering that suite by its test function name matches nothing and still reports success, which is how this stayed hidden locally; the run above filters by the suite.


datlechin:
Superseded by #3003, which landed the same assertion change first. Closing as a duplicate; main is green on OracleObjectQueryTests.

PR #3037, opened 2026-09-21
fix(plugin-redis): list databases on a server that refuses CONFIG

### LINKED ISSUE #3036: Redis: CONFIG GET databases error aborts the sidebar instead of falling back to 16
## What happened?

Connecting to an AWS ElastiCache (Redis OSS, cluster mode disabled) node fails to list databases. The window shows:

```
CONFIG: ERR unknown command 'CONFIG', with args beginning with: 'GET' 'databases'
```

ElastiCache [removes `CONFIG`](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/ClientConfig.RestrictedCommands.html) along with `bgsave`, `debug`, `migrate`, `replicaof`, `save`, `shutdown` and `sync`. It is removed, not ACL-denied, so the server answers with an `unknown command` error rather than `NOPERM`.

`docs/databases/redis.mdx` documents the intended behaviour:

> The sidebar then lists one entry per database, `db0` upward, counted from the server's own `CONFIG GET databases` (16 if it does not answer).

The fallback does not cover this case. In `Plugins/RedisDriverPlugin/RedisPluginDriver.swift:266-273`:

```swift
private func databaseCount(on conn: any RedisCommandChannel) async throws -> Int {
    guard conn.supportsDatabaseSelection else { return 1 }
    let reply = try await conn.run(["CONFIG", "GET", "databases"])
    guard let array = reply.arrayValue, array.count >= 2, let count = array[1].intValue, count > 0 else {
        return 16
    }
    return count
}
```

The `return 16` guard only catches a *successful* reply of an unexpected shape. An error reply never reaches it: `RedisCommandChannel.run` calls `throwIfError(name)` (`Plugins/RedisDriverPlugin/RedisCommandChannel.swift:76-79`), and `RedisReply.throwIfError` (`Plugins/RedisDriverPlugin/RedisReply.swift:87-91`) throws `"\(label): \(message)"` — which is where the `CONFIG: ` prefix in the message comes from. Line 268 therefore throws straight out through `fetchTables` (line 251) and `fetchDatabases` (lines 368-373).

Cluster-mode connections are unaffected: `supportsDatabaseSelection` is false, so line 267 returns 1 before the probe runs. Only cluster-mode-disabled nodes hit this.

## Steps to reproduce

No ElastiCache account needed — any Redis with the command removed reproduces it exactly:

```bash
redis-server --port 6399 --save '' --appendonly no --rename-command CONFIG ''
```

1. New Connection > Redis > `127.0.0.1`, port `6399`, SSL disabled
2. Save & Connect
3. The database list fails with the message above

The server side, for reference:

```
$ redis-cli -p 6399 CONFIG GET databases
ERR unknown command 'CONFIG', with args beginning with: 'GET' 'databases'
```

Connect and auth themselves are fine — `RedisConnectProbe` sends no `CONFIG`. Only database enumeration fails. `CONFIG` is also the only command in the driver's vocabulary that ElastiCache blocks; `INFO`, `SCAN`, `SELECT`, `DBSIZE`, `TYPE`, `TTL` and the rest are all permitted, so the browser works once the count is known.

## Expected behavior

A refused `CONFIG GET databases` falls back to 16, as the docs describe. Something like:

```swift
let reply = (try? await conn.run(["CONFIG", "GET", "databases"])) ?? .null
```

`.null` has no `arrayValue`, so the existing guard supplies the 16.

Worked around locally with a loopback proxy that answers `CONFIG GET databases` with `*2\r\n$9\r\ndatabases\r\n$2\r\n16\r\n` (byte-identical to a real Redis reply) and forwards everything else. With that in front of the same endpoint the sidebar lists db0-db15 with correct key counts from `INFO keyspace`, which confirms the count probe is the only blocker.

## Database type

Redis (AWS ElastiCache, Redis OSS, cluster mode disabled)

## TablePro version

0.75.0

## macOS version & chip

macOS 26.6.2 / Apple Silicon (M1 Pro)

junghyuneun:
Fixes #3036

An AWS ElastiCache node lists no databases at all. The window shows the server's own refusal where the keyspace should be:

```
CONFIG: ERR unknown command 'CONFIG', with args beginning with: 'GET' 'databases'
```

## Root cause

`docs/databases/redis.mdx` already states the intended behaviour:

> The sidebar then lists one entry per database, `db0` upward, counted from the server's own `CONFIG GET databases` (16 if it does not answer).

The fallback did not cover a server that answers with an error. In `RedisPluginDriver.databaseCount(on:)`:

```swift
let reply = try await conn.run(["CONFIG", "GET", "databases"])
guard let array = reply.arrayValue, array.count >= 2, let count = array[1].intValue, count > 0 else {
    return 16
}
```

That `guard` only catches a *successful* reply of an unexpected shape. An error reply never reaches it, because `run` is the choke point that turns one into a throw: `throwIfError(name)` in `RedisCommandChannel.run`, which is also where the `CONFIG: ` prefix on the message comes from. The throw left `databaseCount` and took `fetchTables` and `fetchDatabases` with it.

ElastiCache [restricts `config`](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/ClientConfig.RestrictedCommands.html) alongside `bgsave`, `debug`, `migrate`, `replicaof`, `save`, `shutdown` and `sync`. It is removed rather than denied, so the answer is `unknown command` and not `NOPERM`. Both now fall back; an ACL that denies `config|get` was failing in exactly the same way.

`CONFIG` is also the only command in this driver's vocabulary that ElastiCache blocks. `INFO`, `SCAN`, `SELECT`, `DBSIZE`, `TYPE`, `TTL` and the rest are all permitted, so the browser works once the count is known.

## What changed

The rule moved into `RedisDatabaseCount`, a pure file with no connection behind it, for the reason `RedisReply.swift` gives in its own header: the plugin imports CRedis and the test target cannot, so logic that needs covering lives apart from the connection. `RedisPluginDriver.swift` keeps the two call sites and loses the private helper, and the new file joins the list of plugin sources the test bundle compiles, beside `RedisDatabaseIndex.swift` and `RedisQueuedCommandPolicy.swift`.

A cluster node is still answered from the topology rather than probed, which is what the old first line did.

No docs change: the documented behaviour is what this restores.

## Tests

`RedisDatabaseCountTests`, 12 cases over a stub channel modelled on the one in `RedisQueuedReplyTests.swift`:

- a removed `CONFIG` (the ElastiCache shape) lists 16
- a denied `CONFIG` (`NOPERM`) lists 16
- a server that answers is taken at its word, and is asked exactly `CONFIG GET databases`
- seven unusable answers, including a count of `0` and `-1`, fall back
- a cluster node reports one keyspace and is never probed

The error strings are measured, not invented. Redis 8.10.1 started with `--rename-command CONFIG ''` reproduces the reported message byte for byte:

```
$ redis-cli -p 6399 CONFIG GET databases
ERR unknown command 'CONFIG', with args beginning with: 'GET' 'databases'
```

## Verification, and its limits

I do not have Xcode on this machine, only the Command Line Tools, so **I have not built the project or run the suite**. CI is the real gate here and I will fix whatever it reports.

What I could check, with the CLT Swift 6.4 compiler against faithful stand-ins for `RedisReply`, `RedisCommandChannel` and `run` copied from this repo:

- `swiftc -typecheck -swift-version 6 -strict-concurrency=complete` over the new file and the test file: clean
- the same logic compiled and **run**, which is where the numbers above come from:

```
ElastiCache (CONFIG removed) -> 16
ACL denied (NOPERM)          -> 16
server answers 4             -> 4 [["CONFIG", "GET", "databases"]]
cluster node                 -> 1 probed: []
zero count                   -> 16
```

Separately, the fix was confirmed against the real thing before it was written: a loopback proxy answering `CONFIG GET databases` with `*2\r\n$9\r\ndatabases\r\n$2\r\n16\r\n`, byte-identical to a real Redis reply, in front of an untouched ElastiCache endpoint makes the sidebar list db0-db15 with correct counts from `INFO keyspace`. The count probe is the only thing standing between the driver and that keyspace.

SwiftLint and SwiftFormat were not run for the same reason; the new files follow the spacing, access control and doc-comment style of the files beside them.

## Checklist

- [x] Tests added
- [x] `CHANGELOG.md` updated under `[Unreleased]`, credit left off
- [x] Docs unchanged: the change makes the code match what `docs/databases/redis.mdx` already documents
- [x] No user-facing strings added
- [ ] SwiftLint/SwiftFormat not run locally (no Xcode); relying on CI


datlechin:
Hi @junghyuneun, thanks for your contribution! Unfortunately, this issue has already been fixed in PR #3038.

